PR #9, opened 2019-06-28
update codebase to latest 0.2.0 level

tobespc:
refresh the codewind codebase with the latest code prior to releasing 0.2.0


deboer-tim:
Toby told me to leave a nice message, so here it is: Toby, you're awesome. I can't believe you made all this code change in a week.

Next time let's fix the build so we can move here completely and you can't take credit for everything the team does!

Issue #48, opened 2019-07-29
Increase code coverage of functional tests


rajivnathan:
Functional testing will cover the Turbine module functions in index.ts
https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/server/src/index.ts

The functional tests refer to the tests under https://github.com/eclipse/codewind/tree/master/src/pfe/file-watcher/server/test/functional-test

GeekArthur:
@rajivnathan You can assign this issue to me

ssh24:
**Functions to be tested**
- [x] createProject
- [x] deleteProject
- [x] setLocale
- [ ] setLoggingLevel
- [ ] readWorkspaceSettings
- [ ] testDeploymentRegistry
- [ ] registerListener
- [ ] getProjectTypes
- [ ] getProjectCapabilities
- [ ] projectAction
- [ ] projectSpecification
- [ ] logs
- [ ] shutdown;
- [ ] deleteProject
- [ ] updateStatus
- [ ] deploymentRegistryStatus
- [ ] checkNewLogFile
- [ ] updateProjectForNewChange

PR #60, opened 2019-07-30
Turbine test automation scripts

GeekArthur:
This PR is for the issue: https://github.com/eclipse/codewind/issues/48

- Improve maintainability and scalability by using `getopts` function, we can easily add flags/parameters and usage to the scripts
- Turbine tests can be ran inside PFE container
- Console log that running the test scripts and PFE container log will be saved in the log file (`<TEST_SUITE>-test.log`) for debug purpose
- test.sh is used for kicking off setup.sh and exec.sh for cronjob and fresh users' testing
- exec.sh is used for testing/running functional tests
- setup.sh is used for install and uninstall PFE container
- cronjob.sh is used for kicking off cronjob/nightly run


GeekArthur:
@rajivnathan @maysunfaisal @ssh24, please feel free to leave comments and questions here to review this PR, thanks

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -0,0 +1,114 @@
...
+    BUCKET_NAME=turbine-$TEST_SUITE
+    TEST_DIR=$CW_DIR/src/pfe/file-watcher/server
+    TEST_DIR_CONTAINER=/file-watcher/server
+    TEST_OUTPUT_DIR=~/test_results/$TEST_SUITE/$DATE_NOW/$TIME_NOW
+    TEST_OUTPUT=$TEST_OUTPUT_DIR/test_output.xml
+    TEST_LOG=$TEST_OUTPUT_DIR/$TEST_SUITE-test.log
+    CODEWIND_CONTAINER_ID=$(docker ps | grep codewind-pfe-amd64 | cut -d " " -f 1)
+
+    mkdir -p $TEST_OUTPUT_DIR \
+    && docker exec -it $CODEWIND_CONTAINER_ID bash -c "cd /file-watcher/server; npm install" \
```
ssh24: We shouldn't need to do npm install twice here

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -0,0 +1,114 @@
...
+    if [[ $? -ne 0 ]]; then
+        echo -e "${RED}Setup is failed. ${RESET}\n"
+        exit 1
+    fi
+    
+    # Clone projects to workspace
+    declare -A PROJECT_URLS
+    PROJECT_URLS["codewindtestpython"]="https://github.com/microclimate-dev2ops/SVTPythonTemplate"
+    PROJECT_URLS["codewindtestgo"]="https://github.com/microclimate-dev2ops/microclimateGoTemplate"
+    PROJECT_URLS["codewindtestlagom"]="https://github.com/microclimate-dev2ops/lagomJavaTemplate"
```
ssh24: Can you comment these urls for now from lagom to swift since we are only testing python and go?

src/pfe/file-watcher/server/test/scripts/exec.sh
GeekArthur: One is for dependencies, the other one is for devDependencies **only**, you change your $NODE_ENV to development and that's why you only need one npm install, but the production doesn't

src/pfe/file-watcher/server/test/scripts/exec.sh
GeekArthur: Why we only test python and go? Thought we already finished the creating functional tests?

Screen Shot 2019-07-30 at 5 16 25 PM

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -0,0 +1,119 @@
...
+        i)
+            IGNORE_POST_CLEANUP=$OPTARG
+            # Check if ignore post cleanup argument is corrent
+            if [[ ($IGNORE_POST_CLEANUP != "y") && ($IGNORE_POST_CLEANUP != "n") ]]; then
+                echo -e "${RED}Ignore post cleanup argument is not correct. ${RESET}\n"
+                usage
+                exit 1
+            fi
+            ;;
+        c)
```
maysunfaisal: What does this option do currently?

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -0,0 +1,114 @@
...
+
+function run {
+    docker exec -it $CODEWIND_CONTAINER_ID bash -c "cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml npm run $TEST_SUITE:test:xml" | tee $TEST_LOG
+    docker cp $CODEWIND_CONTAINER_ID:/test_output.xml $TEST_OUTPUT
+    echo -e "${BLUE}Test logs available at: $TEST_LOG ${RESET}"
+
+    if [[ -z $DEVELOPMENT_RUN ]]; then
+        echo -e "${BLUE}Upload test results to test dashboard. ${RESET}"
+        $WEBSERVER_FILE $TEST_OUTPUT_DIR > /dev/null
+        curl --header "Content-Type:text/xml" --data-binary @$TEST_OUTPUT --insecure "https://$DASHBOARD_IP/postxmlresult/$BUCKET_NAME/test" > /dev/null
```
maysunfaisal: You should check if `DASHBOARD_IP` is present if `DEVELOPMENT_RUN` is empty to handle error case

src/pfe/file-watcher/server/test/scripts/setup.sh
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env bash
+
+# Colors for success and error messages
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+BLUE='\033[0;36m'
+RESET='\033[0m'
+
+CW_DIR=~/codewind
```
maysunfaisal: Can we also configure where to clone Codewind repo? If for example, a user has Codewind repo already in `~/codewind` then it cannot do `git clone $CODEWIND_REPO -b $TEST_BRANCH \
    && $CW_DIR/run.sh`

maysunfaisal: or check if there is codewind already present, skip and use the existing codewind run.sh

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: The code configuration is set in a way where we can maintain the list of supported project types. We should first have working test case for specific project type and then start to broaden it

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: The production code install is already done by default as part of the container creation.

GeekArthur: No, production code install dependencies for Turbine when build the PFE image, there are no dependencies that production code can install by default for the Mocha tests, and this PR is for Mocha tests. By default production code set NODE_ENV=production, then for Mocha tests, `npm install` will only install dependencies, we need to run `npm install --only=dev` to install devDependencies.

You can pull the latest master and set NODE_ENV=production, then try `npm install` for Mocha tests, you will figure out the behaviour

src/pfe/file-watcher/server/test/scripts/exec.sh
GeekArthur: I would say it's better you can finish all project types for the function, otherwise it's very confused what is done and what is not done from the list you post above.

src/pfe/file-watcher/server/test/test.sh
GeekArthur: With this option, script does uninstall, install, test, uninstall
Without this option, script only does test
This is the one Elson suggests in today's status meeting

src/pfe/file-watcher/server/test/scripts/exec.sh
GeekArthur: Actually the curl command will throw error by itself if no `DASHBOARD_IP`, but I can add code to handle the check explicitly

src/pfe/file-watcher/server/test/scripts/setup.sh
GeekArthur: This case won't happen because we call uninstall function before install function when user uses our test.sh script to do the testing if user pass `setup` option, if user wants to use the existing Codewind user can run test.sh without option `setup`.

GeekArthur: Also by design, test.sh is the only script that user need to use for testing.

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: The list above is being tested on go and python. Regardless, all project types should work the same way. If that is the case you will still need to change the other project names to have the correct prefix, i.e `codewindtestliberty`, `codewindtestspring` etc

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: Mocha tests don't use any production dependencies. All dependencies needed for mocha tests are dev dependencies. What I am saying is we don't need to do the first `npm install` as it is redundant because the production dependencies are already installed when the container is built.

src/pfe/file-watcher/server/test/scripts/setup.sh
maysunfaisal: understood

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -0,0 +1,119 @@
...
+            fi
+            ;;
+        *)
+            usage
+            exit 0
+            ;;
+    esac
+done
+
+# Chekc mandatory arguments have been set up
```
maysunfaisal: Check spelling

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -0,0 +1,119 @@
...
+if [[ (-z $TEST_TYPE) || (-z $TEST_BRANCH) || (-z $TEST_SUITE) ]]; then
+    echo -e "${RED}Mandatory arguments are not set up. ${RESET}\n"
+    usage
+    exit 1
+fi
+
+if [[ (-z $TEST_CONFIGURATION) ]]; then
+    # Run test cases
+    ./scripts/exec.sh -t $TEST_TYPE -s $TEST_SUITE
```
maysunfaisal: It'll be nice if you could put all these exec, setup and other cmds in this script into functions so that look like setup.sh

src/pfe/file-watcher/server/test/scripts/setup.sh
```diff
@@ -0,0 +1,100 @@
...
+            fi
+            ;;
+        *)
+            usage
+            exit 0
+            ;;
+    esac
+done
+
+# Chekc if the mandatory arguments have been set up
```
maysunfaisal: Check spelling

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -0,0 +1,119 @@
...
+    -h # Display the man page
+EOF
+}
+
+while getopts "t:b:s:i:c:h" OPTION; do
+    case "$OPTION" in
+        t) 
+            TEST_TYPE=$OPTARG
+            # Check if test type argument is corrent
+            if [[ ($TEST_TYPE != "local") && ($TEST_TYPE != "kube") ]]; then
```
rajivnathan: Can we make local a default so that the test.sh script can be run without any arguments?

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -0,0 +1,119 @@
...
+                echo -e "${RED}Test type argument is not correct. ${RESET}\n"
+                usage
+                exit 1
+            fi
+            ;;
+        b) 
+            TEST_BRANCH=$OPTARG
```
rajivnathan: Can we make the default branch master so that the test.sh script can be run without any arguments?

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -0,0 +1,119 @@
...
+            ;;
+        s) 
+            TEST_SUITE=$OPTARG
```
rajivnathan: What are the options for this argument? We should have a default for this as well and probably list the available options if possible.

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -0,0 +1,119 @@
...
+            ;;
+        i)
+            IGNORE_POST_CLEANUP=$OPTARG
```
rajivnathan: This should have a default as well. Cleanup can be true by default?

src/pfe/file-watcher/server/test/scripts/setup.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
+
+# Colors for success and error messages
+GREEN='\033[0;32m'
+RED='\033[0;31m'
+BLUE='\033[0;36m'
+RESET='\033[0m'
+
+CW_DIR=~/codewind
```
rajivnathan: I think the tests should be run against the codewind repository where the test.sh script is run. We can avoid cloning the repository again in another location and the added benefit is that If a user clones the codewind repository and makes changes to the tests they can run tests with their current changes without having to push the changes and specify a different branch.

src/pfe/file-watcher/server/test/scripts/setup.sh
```diff
@@ -0,0 +1,100 @@
...
+    -f # Setup function, currently support 'install' and 'uninstall' - Mandatory
+    -h # Display the man page
+EOF
+}
+
+# Install function does the following things
+# 1 Download Codewind repository
+# 2 Build Codewind images
+# 3 Run Codewind containers
+function install {
```
rajivnathan: I'm not sure we need an install/uninstall based on the previous comment about reusing the existing repo instead of cloning a separate one.

src/pfe/file-watcher/server/test/scripts/exec.sh
GeekArthur: If all your new tests only need devDependencies, I am fine remove the first `npm install`

src/pfe/file-watcher/server/test/scripts/exec.sh
GeekArthur: Will update the prefix, you can double check those project names with prefix and make sure they are the names you use in your test cases. Or if we can, in the test cases we may use env var to get the project names dynamically instead of hardcode project names

src/pfe/file-watcher/server/test/test.sh
GeekArthur: I can do the change

src/pfe/file-watcher/server/test/scripts/setup.sh
GeekArthur: Sure thing

src/pfe/file-watcher/server/test/test.sh
GeekArthur: I change my mind, I think IGNORE_POST_CLEANUP should be 'yes' by default, only cronjob needs to do the post cleanup and we can set it up for our cronjob

src/pfe/file-watcher/server/test/cronjob.sh
```diff
@@ -0,0 +1,84 @@
...
+CODEWIND_REPO=git@github.com:eclipse/codewind.git
+
+# Make all options mandatory to ensure cronjob admin knows what exectly each cronjob does, it's for debug purpose
+function usage {
+    me=$(basename $0)
+    cat <<EOF
+Usage: $me: [-<option letter> <option value> | -h]
+Options:
+    -t # Test type, currently support 'local' and 'kube' - Mandatory
+    -s # Test suite - Mandatory
```
ssh24: Did we discuss to make this optional and fallback to functional for default?

src/pfe/file-watcher/server/test/scripts/setup.sh
```diff
@@ -0,0 +1,85 @@
...
+    -f # Setup function, currently support 'install' and 'uninstall' - Mandatory
+    -h # Display the man page
+EOF
+}
+
+# Install function does the following things
+# 1 Build Codewind images
+# 2 Run Codewind containers
+function install {
+    ../../../../../run.sh
```
ssh24: Can we pass the `CW_DIR` variable to this file from exec.sh as well and use it here as well?

src/pfe/file-watcher/server/test/scripts/setup.sh
```diff
@@ -0,0 +1,85 @@
...
+        echo -e "${RED}Setup is failed. ${RESET}\n"
+        exit 1
+    fi
+}
+
+# Uninstall function does the following things
+# 1 Remove Codewind and Codewind app containers
+# 2 Remove Codewind and Codewind app images
+function uninstall {
+    ../../../../../stop.sh
```
ssh24: Similary here

src/pfe/file-watcher/server/test/cronjob.sh
GeekArthur: See the command above `# Make all options mandatory to ensure cronjob admin knows what exectly each cronjob does, it's for debug purpose`

GeekArthur: I am thinking of user only uses `tesh.sh`, we make all options optional and set up default options value for users' convenience. But `cronjob.sh` is only for our cronjob/nightly run, it's better to force us to fill out all options for debug purpose

src/pfe/file-watcher/server/test/scripts/setup.sh
GeekArthur: We can't because there is no relationship between `setup.sh` and `exec.sh`, `test.sh` calls `setup.sh` and `exec.sh`. Since we don't know the path of user's Codewind repo, we can only get the Codewind repo path using the relative path. If we want the absolute path we can do something like this
```
cd ../../../../../
CW_DIR=$(pwd)
cd -
```
But again, it's still get it from relative path

src/pfe/file-watcher/server/test/scripts/setup.sh
GeekArthur: Same above

GeekArthur:
@ssh24 @rajivnathan @maysunfaisal Make some changes for fix and moving repo clone to `cronjob.sh`, please review

src/pfe/file-watcher/server/test/cronjob.sh
ssh24: Ok. Makes sense.

src/pfe/file-watcher/server/test/cronjob.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+
+rm -rf $CW_DIR \
+&& git clone $CODEWIND_REPO -b $TEST_BRANCH \
+&& cd $CW_TEST_DIR \
+&& ./test.sh -t $TEST_TYPE -s $TEST_SUITE -p $POST_CLEANUP -c $TEST_CONFIGURATION \
+&& rm -rf $CW_DIR
+
+if [[ $? -ne 0 ]]; then
+    echo -e "${RED}Cronjob is failed. ${RESET}\n"
```
ssh24: Nitpick: `Cronjob has failed.`

rajivnathan: Updated with other minor message changes.

GeekArthur:
@rajivnathan The first two issues you mention above are due to your latest commit (Message updates), in your latest commit (Message updates) you change the `declare` command option from `-A` to `-a` here: https://github.com/eclipse/codewind/pull/60/commits/7e0616fc1d0f92cf5f67e7e7111062ebb0e36ea5#diff-d94662cd602d8ca5cd85083d8743fbdbL51. This change is not correct because `PROJECT_URLS` should be a dictionary which is associative array not an indexed array, so need to use `-A` instead of `-a`. I already verified all projects are cloned properly once revert your change.

FYI this is the man page of `declare` command:
```
jingfus-mbp:codewind jingfuwang$ declare --help
declare: declare [-aAfFgilnrtux] [-p] [name[=value] ...]
    Set variable values and attributes.
    
    Declare variables and give them attributes.  If no NAMEs are given,
    display the attributes and values of all variables.
    
    Options:
      -f	restrict action or display to function names and definitions
      -F	restrict display to function names only (plus line number and
    		source file when debugging)
      -g	create global variables when used in a shell function; otherwise
    		ignored
      -p	display the attributes and value of each NAME
    
    Options which set attributes:
      -a	to make NAMEs indexed arrays (if supported)
      -A	to make NAMEs associative arrays (if supported)
      -i	to make NAMEs have the `integer' attribute
      -l	to convert the value of each NAME to lower case on assignment
      -n	make NAME a reference to the variable named by its value
      -r	to make NAMEs readonly
      -t	to make NAMEs have the `trace' attribute
      -u	to convert the value of each NAME to upper case on assignment
      -x	to make NAMEs export
    
    Using `+' instead of `-' turns off the given attribute.
    
    Variables with the integer attribute have arithmetic evaluation (see
    the `let' command) performed when the variable is assigned a value.
    
    When used in a function, `declare' makes NAMEs local, as with the `local'
    command.  The `-g' option suppresses this behavior.
    
    Exit Status:
    Returns success unless an invalid option is supplied or a variable
    assignment error occurs.
```

@ssh24 I saw the latest commit in your PR is trying to fix the script: https://github.com/eclipse/codewind/pull/95/commits/d5bcd592fe9fcd1942bf7c3a96e6a077ce278349, actually you don't need to change the structure to fix, all you need to do is change `-a` to `-A` for `declare` command. Explanation is above.

ssh24:
Oh, yes that might have broken it. I kind of like the structure in https://github.com/eclipse/codewind/commit/d5bcd592fe9fcd1942bf7c3a96e6a077ce278349 that's why I did it like that as a json object. But if you guys want we can keep the old version as well. Thoughts @GeekArthur @rajivnathan ?

GeekArthur:
@ssh24 I would suggest use the old version because
1 Use dictionary data structure is better than array for our case, using dictionary we can easily track the project clone URL because index is project name not number, also we can save lines of code by using dictionary
2 The new version relay on `jq`, it would be better we just use pure bash script instead of installing `jq` explicitly

GeekArthur:
@rajivnathan Also, thanks for helping me update those messages, they look very nice now

ssh24:
I just tried with `-A` and it gives me error:
```
./scripts/exec.sh: line 51: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
Cloning projects to /Users/sakib@ibm.com/workspace/dev/projects/codewind/codewind-workspace 

Cloning https://github.com/microclimate-dev2ops/swiftTemplate. 

curl: (52) Empty reply from server

Running functional tests.
```

With `-a` it gives me the following error and only clones the last project in the array
```
Cloning projects to /Users/sakib@ibm.com/workspace/dev/projects/codewind/codewind-workspace 

Cloning https://github.com/microclimate-dev2ops/swiftTemplate. 

curl: (52) Empty reply from server

Running functional tests. 
```

I think your script also relies on `declare` module? It does not show me the listed options when I do `--help`
```
sakibs-mbp:test sakib@ibm.com$ declare --help
bash: declare: --: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```

GeekArthur:
@ssh24 Which version of your bash? Can you try `bash -version`? I think your bash version is out of date, `declare` is buildin for bash

ssh24:
This is my bash version
```
Sakibs-MacBook-Pro:codewind sakib@ibm.com$ bash --version
bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
Copyright (C) 2007 Free Software Foundation, Inc.
hostname -s
whoami
```

GeekArthur:
@ssh24 Yes, the issue is your bash version. 10 years ago bash introduces associative arrays, but you version is 12 years ago. Try to update your bash to any version higher than 4.0 (4.x, 5.x) then it will work.

ssh24:
I upgraded bash but still see this error:
```
Cloning https://github.com/microclimate-dev2ops/nodeExpressTemplate.git. 

curl: (52) Empty reply from server
Cloning https://github.com/microclimate-dev2ops/javaMicroProfileTemplate. 

curl: (52) Empty reply from server
Cloning https://github.com/microclimate-dev2ops/SVTPythonTemplate. 

curl: (52) Empty reply from server
Cloning https://github.com/microclimate-dev2ops/springJavaTemplate. 

curl: (52) Empty reply from server
Cloning https://github.com/microclimate-dev2ops/lagomJavaTemplate. 

curl: (52) Empty reply from server
Cloning https://github.com/microclimate-dev2ops/swiftTemplate. 

curl: (52) Empty reply from server
Cloning https://github.com/microclimate-dev2ops/microclimateGoTemplate. 

curl: (52) Empty reply from server
```

Something wrong with curl?

GeekArthur:
Yes, the issue is form curl and the error message means either the server is not up or the curl command is wrong. You can check two things:
1 Your PFE is up and running
2 On your branch, don't use your new version of project cloning and make sure change back to the old version of project cloning properly
If you still see the same error after you check the two things above you can echo the curl command and manually try the curl on your environment to see if it's a curl command issue or environment issue

I am just testing and it works
Screen Shot 2019-08-08 at 10 59 07 AM

ssh24:
I have PFE up and running and I am using the existing version of the clone logic on master. And it still does not work for me. But with the new logic added in https://github.com/eclipse/codewind/commit/d5bcd592fe9fcd1942bf7c3a96e6a077ce278349 works.

This is my curl version
```
sakibs-mbp:test sakib@ibm.com$ curl --version
curl 7.54.0 (x86_64-apple-darwin18.0) libcurl/7.54.0 LibreSSL/2.6.5 zlib/1.2.11 nghttp2/1.24.1
Protocols: dict file ftp ftps gopher http https imap imaps ldap ldaps pop3 pop3s rtsp smb smbs smtp smtps telnet tftp 
Features: AsynchDNS IPv6 Largefile GSS-API Kerberos SPNEGO NTLM NTLM_WB SSL libz HTTP2 UnixSockets HTTPS-proxy 
```

ssh24:
Ok, so I figured out why my curl command was not working. And it is because I have been working with dev mode set to true. With dev mode set to true, when we copy the test files over the portal server restarts (it should not restart) and by the time we call curl, the portal server is not up yet.

I will add `/test` folder under file-watcher to be ignored in development mode to avoid this because the test itself takes care about the server.

PR #62, opened 2019-07-30
import tarfile

laven26:
Signed-off-by: laven-s <laven.s@ibm.com>

### DIFF
--- a/src/initialize/src/classes/ProjectInitializer.ts
+++ b/src/initialize/src/classes/ProjectInitializer.ts
@@ -18,6 +18,7 @@ import { detectType } from '../controllers/typeDetectionController';
 import { writeCwSettings } from '../controllers/cwSettingsController';
 import { importProjectFromGitRepo } from '../controllers/gitController';
 import { InitializeResult } from 'initializeTypes';
+import { importProjectFromTar } from '../controllers/tarController';
 
 export default class ProjectInitializer {
   projectName: string;
@@ -44,6 +45,16 @@ export default class ProjectInitializer {
       this.projectMountDirectory,
       gitBranch
     );
+    return this.initializeExistingProject();
+  }
+
+  async initializeProjectFromTar(
+    tarFile: string
+    ): Promise<InitializeResult> {
+      await importProjectFromTar(tarFile,
+        this.projectName,
+        this.projectMountDirectory
+        );
     return this.initializeExistingProject();
   }
 
--- a/src/initialize/src/controllers/tarController.ts
+++ b/src/initialize/src/controllers/tarController.ts
@@ -15,7 +15,7 @@ import request from 'request';
 
 export function importProjectFromTar(tarFile: string, name: string, dest: string): Promise<void> {
 
-    const fileName = `/app/${name}.tar.gz`;
+    const fileName = `/tmp/${name}.tar.gz`;
     const file = fs.createWriteStream(fileName);
 
     return new Promise((resolve, reject) => {
--- a/src/initialize/src/controllers/tarController.ts
+++ b/src/initialize/src/controllers/tarController.ts
@@ -36,7 +36,15 @@ export function importProjectFromTar(tarFile: string, name: string, dest: string
             resolve();
           });
     })
-    .on('error', reject);
+    .on('error', (err) => {
+        try {
+            fs.removeSync(fileName);
+        }
+        catch (err) {
+            console.log(err);
+        }
+        reject(err);
+    });
     });
 
 }
--- a/src/initialize/src/controllers/tarController.ts
+++ b/src/initialize/src/controllers/tarController.ts
@@ -15,7 +15,7 @@ import request from 'request';
 
 export function importProjectFromTar(tarFile: string, name: string, dest: string): Promise<void> {
 
-    const fileName = name + '.tar.gz';
+    const fileName = `/app/${name}.tar.gz`;
     const file = fs.createWriteStream(fileName);
 
     return new Promise((resolve, reject) => {
--- a/src/initialize/src/controllers/tarController.ts
+++ b/src/initialize/src/controllers/tarController.ts
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2019 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+import fs from 'fs-extra';
+import { exec } from 'child_process';
+import request from 'request';
+
+export function importProjectFromTar(tarFile: string, name: string, dest: string): Promise<void> {
+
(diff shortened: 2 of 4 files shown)


src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,41 @@
...
+import fs from 'fs-extra';
+const { exec } = require('child_process');
+import request from 'request';
+
+export async function importProjectFromTar(tarFile: string, name: string, dest: string): Promise<void> {
+    
+    const fileName = name + '.tar.gz';
+    let file = fs.createWriteStream(fileName);
+
+    await new Promise((resolve, reject) => {
```
makandre: Why don't we return the Promise here?  It will be awaited by the caller

src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        resolve();
+        
+    })
+    .on('error', (error) => {
+        reject(error);
+        })
+    })
+    .catch(error => {
+        throw new Error('Extracting tar file failed');
```
makandre: I think we should reject the `error` here?

src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+import fs from 'fs-extra';
+const { exec } = require('child_process');
+import request from 'request';
+
+export async function importProjectFromTar(tarFile: string, name: string, dest: string): Promise<void> {
```
makandre: If we get rid of the `await` below this doesn't need to be a async function

makandre:
@laven26 How are the changes coming along?

src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    return new Promise((resolve, reject) => {
+    request(tarFile)
+    .pipe(file)
+    .on('finish', () => {
+          exec(`tar -xvf ${fileName} -C ${dest}`, (err: any, stdout: any, stderr: any) => {
+              fs.removeSync(fileName);
+          });
+
+        resolve();
```
makandre: Just noticed something... this actually needs to go into exec callback above.  And removeSync should have try/catch

```
try {
    fs.removeSync(fileName); 
}
catch (err) {
   // probably just log this here.
}
resolve();
```

src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    })
+    .on('error', (error) => {
```
makandre: A more concise syntax here can be `.on('error', reject);

src/initialize/src/classes/ProjectInitializer.ts
```diff
@@ -47,6 +48,16 @@ export default class ProjectInitializer {
     return this.initializeExistingProject();
   }
 
+  async initializeProjectFromTar(
+    tarFile: string
+    ) {
```
makandre: missed return type here

makandre:
LGTM

sghung:
@rajivnathan can you help review this? Thank you

sghung:
@tobespc or @markcor11, can you help review this? Thank you

src/initialize/src/main.ts
```diff
@@ -42,7 +42,9 @@ export async function initializeProject(projectName: string, projectDirectory: s
   const projectInitializer = new ProjectInitializer(projectName, projectDirectory);
 
   try {
-    if (isGitCloneRequired(projectName, gitRepository)) {
+    if (gitRepository.endsWith("tar.gz")) {
```
hhellyer: I'm not that comfortable with overloading GIT_REPO to also mean "any url to a tar.gz". Can we just use a different env var and check if that was set or if GIT_REPO was set?

src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,42 @@
...
+                /* Logging error when removing temp dir fails */
+                console.log(err);
+            }
+            if (err) {
+                return reject(err);
+            }
+            resolve();
+          });
+    })
+    .on('error', reject);
```
hhellyer: Does this remove the empty or partial file if the download fails? (e.g. on a bad url or github being down)

src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,42 @@
...
+ *******************************************************************************/
+
+import fs from 'fs-extra';
+import { exec } from 'child_process';
+import request from 'request';
+
+export function importProjectFromTar(tarFile: string, name: string, dest: string): Promise<void> {
+
+    const fileName = name + '.tar.gz';
+    const file = fs.createWriteStream(fileName);
```
hhellyer: This should probably be an absolute filename, otherwise it will just be created in the current working directory (and I'm not sure exactly where that will be in all cases).

src/pfe/portal/modules/projectInitializer.js
```diff
@@ -189,10 +189,17 @@ const createEnvForContainerCreation = (user, projectName, gitInfo) => {
...
+      env.push(
+        `PROJ_NAME=${projectName}`,
+        `TAR_FILE=${gitInfo.repo}`,
+      );
+    } else {
+      env.push(
+        `PROJ_NAME=${projectName}`,
+        `GIT_REPO=${gitInfo.repo}`,
+      );
+    }
```
makandre: The `gitInfo.branch` check can go into this `else` block right?  Since it would be applicable only for non-tarball scenario

src/pfe/portal/modules/projectInitializer.js
```diff
@@ -189,10 +189,17 @@ const createEnvForContainerCreation = (user, projectName, gitInfo) => {
   env.push(`CW_EXTENSIONS=${JSON.stringify(cwExtensions)}`);
 
   if (projectName && gitInfo.repo) {
-    env.push(
-      `PROJ_NAME=${projectName}`,
-      `GIT_REPO=${gitInfo.repo}`,
-    );
+    if (gitInfo.repo.endsWith("tar.gz")) {
```
makandre: A minor nit: while you're making changes, could you make this check for `.tar.gz`?  (add dot in the front)

src/initialize/src/controllers/tarController.ts
```diff
@@ -0,0 +1,50 @@
...
+                /* Logging error when removing temp dir fails */
+                console.log(err);
+            }
+            if (err) {
+                return reject(err);
+            }
+            resolve();
+          });
+    })
+    .on('error', () => {
```
makandre: I believe there's an `err` passed here you can then pass it to `reject`

src/initialize/src/controllers/tarController.ts
hhellyer: Please use `/tmp` not `/app`. `/app` exists in the initialiser pod but the initialiser code is run directly as an npm when we are in che and `/app` doesn't exist there.
(Also /tmp is a better location for temporary files.)

sghung:
Thanks for the reviews Andrew and Howard

Issue #92, opened 2019-08-01
Investigate using dnf instead of yum in Codewind PFE


johnmcollier:
We should look at using `dnf` for installing packages in the Codewind PFE container, rather than `yum`. We've hit some issues before where yum failed to install packages and it didn't error out when that happened. This is because yum doesn't error out if packages fail to install, whereas dnf does.

PR #93, opened 2019-08-01
Fix dead buildah link and error if curl 404s

johnmcollier:
Signed-off-by: John Collier <John.J.Collier@ibm.com>

Fixes a dead link in the Dockerfile that was used to retrieve the buildah RPM. Also updates the curl command so that it fails if the link 404s, since it appears it doesn't do that automatically

Also opened https://github.com/eclipse/codewind/issues/92 to look at using dnf in PFE for the future


johnmcollier:
Exec'd into PFE and verified that the buildah executable was present and functioning.

PR #95, opened 2019-08-01
Increase code coverage of functional tests

ssh24:
## Description

Increase code coverage for the functional test suite.

Related to https://github.com/eclipse/codewind/issues/48

Tests are only added for Go, Python and Lagom type projects

Functions tested:
- [x] createProject (@ssh24)
- [x] deleteProject (@ssh24)
- [x] setLocale (@ssh24)
- [x] setLoggingLevel (@ssh24)
- [x] readWorkspaceSettings (@ssh24)
- [x] testDeploymentRegistry (@ssh24)
- [x] deploymentRegistryStatus (@ssh24)
- [x] registerListener (@ssh24)
- [x] getProjectTypes (@ssh24)
- [x] getProjectCapabilities (@ssh24)
- [x] projectAction (@ssh24)
- [x] projectSpecification (@ssh24)
- [x] logs (@ssh24)
- [x] shutdown (@ssh24)
- [x] updateStatus (@ssh24)
- [x] checkNewLogFile (@ssh24)
- [x] updateProjectForNewChange (@ssh24)

### Che result
Screen Shot 2019-08-08 at 3 11 29 PM

### Local result
Screen Shot 2019-08-08 at 3 17 07 PM


ssh24:
@jingfuwang @maysunfaisal @rajivnathan The PR is ready for review now. Could you please review? Thanks.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
```diff
@@ -33,64 +43,57 @@ describe("Functional Test Suite", () => {
...
-  describe("generic suite", () => {
-    describe("setLocale", () => {
-      localeTestSuite();
+      utils.writeToFile(workspace_settings_file, workspace_settings_file_content_json, (err) => {
+        if (err) done(err);
+        fs.chown(workspace_settings_file, 1001, 2004, (err) => {
+          if (err) done(err);
+          utils.getRegistry((err, res, body) => {
+            if (err) done(err);
+            expect(res);
```
GeekArthur: If you want to check the target(variable/object) existence, Mocha has `expect(res).to.exist;` for checking both `null` and `undefined`. It's a general comment and can apply to all lines for checking the target(variable/object) existence

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
```diff
@@ -8,23 +8,33 @@
...
 const dockerProjects = app_configs.supportedDockerProjects;
 const projectConfigs = app_configs.projectConfigs;
 
-describe("Functional Test Suite", () => {
+const socket = new SocketIO();
+const genericSuite = new GenericTestSuite(socket);
+const projectSuite = new ProjectTestSuite(socket);
+
+describe("PFE - functional test", () => {
   beforeEach("check for pfe to be up", (done) => {
```
GeekArthur: Is it necessary to check for PFE to be up for each test case? Currently we have ~200 test cases, seem like we check ~200 times for checking PFE to be up. It will increase the overall testing time as the test cases grow

src/pfe/file-watcher/server/test/functional-test/configs/pfe.config.ts
```diff
@@ -10,13 +10,16 @@
  *******************************************************************************/
 
 export const pfe = {
-    "PROTOCOL": process.env.PFE_PROTOCOL || "http",
+    "PROTOCOL": process.env.PFE_PROTOCOL || process.env.IN_K8 ? "https" : "http",
     "HOST": process.env.PFE_HOST || "localhost",
-    "PORT": process.env.PFE_PORT || "9090",
+    "PORT": process.env.PFE_PORT || process.env.IN_K8 ? "9191" : "9090",
 };
```
GeekArthur: You can add `BASE_URL` here to concatenate PROTOCOL, HOST and PORT, then all test cases can use `BASE_URL` and don't need to concatenate by itself.

src/pfe/file-watcher/server/test/functional-test/configs/pfe.config.ts
```diff
@@ -10,13 +10,16 @@
...
 
 const apis = ["/api/v1/"];
```
GeekArthur: I think we can name `apiv1` in case we have upper version API in the future. In Microclimate, some test cases use different version API, if we specify the version here, it's easier for us to apply API version to the following APIs for future.

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
```diff
@@ -0,0 +1,45 @@
+/*******************************************************************************
```
GeekArthur: Are there any specific difference between `project.config.ts` and `app.config.ts`? If they are all config files for project, we can merge them into one config file only for project.

src/pfe/file-watcher/server/test/functional-test/suites/generic/index.ts
```diff
@@ -0,0 +1,61 @@
...
+        this.projectTest = new ProjectTest("project test");
+        this.workspaceTest = new WorkspaceTest("workspace test");
+    }
+
+    /**
+     * @function
+     * @description The run test function that triggers all generic test classes.
+     *
+     */
+    runTest(runOnly?: boolean): void {
```
GeekArthur: I think this parameter `runOnly` is used for only running individual test suite? If so does that mean if we want to run individual test suite we have to manually change the test case to make `runOnly` true then copy the test case to PFE container?

Generally for running individual test suite/unit, Mocha has option
`-g, --grep <pattern>            only run tests matching <pattern>` but you need to have nested `describe` to describe the test suite then Mocha can grep the specific test suite/unit

GeekArthur:
@ssh24 Hey Sakib, thanks for creating the new functional test cases to increase the code coverage! Overall is good to me, I leave some comments and still review it, hopefully I can finish reviewing by the end of today.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
ssh24: `expect(res);` and `expect(res).to.exist;` are the exact same.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
ssh24: We don't check 200 times. This is a before hook for the whole test. Only checks once.

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
ssh24: Yeah. `project.config.ts` is project specific configs. `app.config.ts` is just the test app configuration.

src/pfe/file-watcher/server/test/functional-test/suites/generic/index.ts
ssh24: Nope. You don't have to manually copy anything to the container. Only set `runOnly` to true and run `./test.sh local` and it will only run that test suite.

src/pfe/file-watcher/server/test/functional-test/configs/pfe.config.ts
ssh24: Sure, I can change it to call `apiV1`

ssh24: Do we need to call it `apiv1`? It is an array of various versions...so if we do have v2 api in the future, we can just add it to the array and it should be fine.

src/pfe/file-watcher/server/test/functional-test/configs/pfe.config.ts
ssh24: Sure, I think we can do that. Create `BASE_URL` and use that for the tests.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
GeekArthur: This is what I got:

Screen Shot 2019-08-09 at 5 17 35 PM

ssh24: Actually sorry. Yes forgot it was a `beforeEach` hook. Now that I remember, we need to check that PFE is up before each test to verify that it didn't go down. Because our tests can't run without PFE being up.

src/pfe/file-watcher/server/test/functional-test/suites/generic/index.ts
GeekArthur: But all the test cases we run are inside container, does the `runOnly` you set is against you local test case? If you directly run `./test.sh`, does it still trigger the old test cases inside container?

ssh24: `runOnly` set is applied to the test cases inside the container because the `./test.sh` script copies over the `/test` directory to the container before running the test.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
GeekArthur: Then we go back the original question, is it necessary to check for PFE to be up for ~200 times? I think check before each test suite would be sufficient, it's same idea for our Microclimate test automation, we don't check that for each test case.

ssh24: I think it is needed. Because test cases in a suite depend on PFE to be up and if suddenly PFE goes down our test will be stuck on timeouts. With the simple GET request we can easily check if PFE is up before every test. It is not an overhead IMO.

src/pfe/file-watcher/server/test/functional-test/configs/pfe.config.ts
GeekArthur: Yes, it's an array, but for adding v2 we have to do this:
```
const apis = ["/api/v1/", "/api/v2/"];

export const pfeAPIs = {
    "projects": apis[0] + "projects",
    "registry": apis[1] + "registry"
};
```

Instead, I mean we can make it more readable for adding v2 for future like this:
```
const apiv1 = "/api/v1/";
const apiv2 = "/api/v2/";

export const pfeAPIs = {
    "projects": apiv1 + "projects",
    "registry": apiv2 + "registry"
};
```

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
GeekArthur: Actually our test cases are running inside PFE container, if PFE goes down our tests will terminates immediately and there is no timeout issue for PFE goes down. Given that, I don't think we need to check if the PFE container is up or not, test cases themselves can help us.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
GeekArthur: I am not sure if `expect(res);` is proper way to use, but from the Chai man page: https://www.chaijs.com/api/bdd/, `expect(res).to.exist;` is the official way to check existence and it's more readable

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
GeekArthur: I still can't see the obvious difference between the two files, but if you want make it two files I fine with it, but leave detailed comments or instructions to ensure people can add project config codes to the correct file in the future.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
ssh24: Yes, logically they do the exact same. If you want I can change it but they do the exact same and it's in a lot of places so will be tedious.

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
ssh24: That's actually true. Good catch. We had this implementation before when we were running before outside container so it made sense. But now if PFE goes down, you are right the test will also go down. So it's not needed anymore

src/pfe/file-watcher/server/test/functional-test/configs/pfe.config.ts
ssh24: Sure, we can do that.

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
ssh24: Yes, that is a good idea. I added the comments on top of the file.

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/logging.ts
```diff
@@ -0,0 +1,52 @@
...
+export default class LoggingTest {
+    testName: string;
+
+    constructor(testName: string) {
+        this.testName = testName;
+    }
+
+    run(runOnly?: boolean): void {
+        (runOnly ? describe.only : describe)(this.testName, () => {
+            const logLevels = ["info", "warn", "error", "trace", "debug", "invalid"];
```
GeekArthur: It's a general comment, we can define the variable type to make the variable more strict and readable.
eg.
```
const logLevels: Array<string> = ["info", "warn", "error", "trace", "debug", "invalid"];
const invalidLevel: string = _.cloneDeep(logLevels).pop();
```
I find some variables we define some we don't, in order not to block this PR to merge, we can do this improvement when we extend the test cases for other project types

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/logging.ts
```diff
@@ -0,0 +1,52 @@
...
+            const invalidLevel = _.cloneDeep(logLevels).pop();
+            for (const level of logLevels) {
+                it(`set logging level to ${level}`, async () => {
+                    try {
+                        await genericLib.setLoggingLevel(level);
```
GeekArthur: I notice one issue when I run the test cases, we don't have the PFE log for setting the log level to `warn` and `error`. If the test cases failed and we don't have PFE log, then it's not good for us to debug.

Screen Shot 2019-08-12 at 2 23 50 PM

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/project.ts
```diff
@@ -0,0 +1,64 @@
...
+    run(runOnly?: boolean): void {
+        (runOnly ? describe.only : describe)(this.testName, () => {
+            this.runGetProjectTypes();
+        });
+    }
+
+    private runGetProjectTypes(): void {
+        describe("getProjectTypes function", () => {
+            it("get project type without a location", async () => {
+                const info: any = await genericLib.getProjectTypes();
```
GeekArthur: We don't have the PFE log for the test case `get project type without a location`, but the test cases passed then we should already verified the output from PFE and the PFE should have the output in the PFE log. If the test cases failed and we don't have PFE log, then it's not good for us to debug. Wonder is this due to the issue of combining PFE log and Mocha tests log?

Screen Shot 2019-08-12 at 2 35 21 PM

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/workspace.ts
```diff
@@ -0,0 +1,299 @@
...
+            }
+        });
+    }
+
+    private afterEachHook(settingsPath: string, backupSettingsPath: string): void {
+        afterEach("delete backup settings file, revert back original file and load the current workspace settings", async () => {
+            if (await fs.existsSync(backupSettingsPath)) {
+                await fs.renameSync(backupSettingsPath, settingsPath);
+                await fs.chownSync(settingsPath, 1001, 2004);
+                await genericLib.readWorkspaceSettings();
```
GeekArthur: We can add expect or verification code here for `genericLib.readWorkspaceSettings();` to make sure the setting file is back to the original version correctly.

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/workspace.ts
```diff
@@ -0,0 +1,299 @@
...
+                        await fs.writeFileSync(settingsPath, JSON.stringify(badSettingsContent));
+                        await fs.chownSync(settingsPath, 1001, 2004);
+                    }
+                });
+
+                it("read workspace settings with invalid deployment registry", async () => {
+                    const info: any = await genericLib.readWorkspaceSettings();
+                    const failureMsg = "Codewind detected an error with the Deployment Registry x^72@!$&. Please ensure it is a valid Deployment Registry.";
+                    const targetEvent = eventConfigs.events.deploymentRegistryStatus;
+                    console.log("Info: %j", info);
```
GeekArthur: Is this `console.log()` a debug code for test cases? We can remove this, otherwise it's pretty confused that it's from PFE log or from Mocha tests

Screen Shot 2019-08-12 at 2 43 52 PM

src/pfe/file-watcher/server/test/functional-test/suites/generic/index.ts
GeekArthur: Verified, it works, nice!

src/pfe/file-watcher/server/test/functional-test/functional.test.ts
GeekArthur: I see, I think we can do this improvement when we extend the test cases for other project types. It's fine for this PR.

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/project.ts
```diff
@@ -0,0 +1,64 @@
...
+                const projectType = "spring";
+                const info: any = await genericLib.getProjectTypes(path.join(app_configs.codewindWorkspaceDir, app_configs.projectPrefix + projectType));
+                expect(info);
+                expect(info.statusCode);
+                expect(info.statusCode).to.equal(200);
+                expect(info.types);
+                expect(info.types).to.deep.equal([projectType]);
+            });
+
+            it("get project type with an undefined project location", async () => {
```
GeekArthur: I run the tests, one test case failed, issue is we expect [] but actually it returns ['docker'] for test case `get project type with an undefined project location`

Screen Shot 2019-08-12 at 3 04 12 PM

ssh24: I fixed this. It should work now

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/logging.ts
ssh24: That is probably product related. We need to add the PFE log to the server code when level `warn` or `error` is selected.

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/project.ts
ssh24: I think it doesn't output the log because by default the log level is set to `INFO` and with Rajiv's change on logs, it will not print trace or debug level messages such as this case.

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/workspace.ts
ssh24: Yeah that was for test debugging. Will remove it.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
```diff
@@ -0,0 +1,206 @@
...
+                                    expect(event.eventData).to.haveOwnProperty(eventKey);
+                                    if (eventKey === "projectID") {
+                                        expect(event.eventData[eventKey]).to.equal(projectID);
+                                    }
+                                    if (eventKey === "appStatus" || eventKey === "buildStatus") {
+                                        expect(event.eventData[eventKey]).to.equal(state);
+                                    }
+                                    if (eventKey === "appErrorStatus" || eventKey === "detailedBuildStatus") {
+                                        expect(event.eventData[eventKey]).to.equal(errMsg);
+                                    }
```
GeekArthur: We need to check the `detailedAppStatus` as well when we update the application status.

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/logging.ts
ssh24: Yes, I fixed it.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
```diff
@@ -0,0 +1,125 @@
...
+        const types = ["MODIFY", "CREATE"];
+
+        for (const type of types) {
+            describe(`update file with type = ${type.toLowerCase()}`, () => {
+                updateProject(type);
+            });
+        }
+
+        function updateProject(type: string): void {
+            let filesList = ["file1", "file2", "file3"];
```
GeekArthur: For testing `CREATE` event, why would we need to test it three times for `file1`, `file2` and `file3` instead of just test one time?

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
```diff
@@ -0,0 +1,125 @@
...
+            if (type.toLowerCase() === "modify") filesList = project_configs.filesToUpdate[projectLang];
+            for (const file of filesList) {
```
GeekArthur: Seems like the variable `file` is only used for logging? When we call function `updateProjectForNewChange` for testing, we pass `eventArray` and seems like the only change in  `eventArray` is the `type`, the `path` is always hardcoded to `projData.location + "Dockerfile"`? Do we actually  use the variable `file` from the `filesToUpdate` list in our config file?

```
export const filesToUpdate: any = {
    "go": ["Dockerfile", "main.go"],
    "python": ["Dockerfile", "app.py"],
    "lagom": ["Dockerfile",
        "hello-api/src/main/java/com/example/rp/test/lagomendpoints/api/HelloService.java",
        "hello-impl/src/main/java/com/example/rp/test/lagomendpoints/impl/HelloModule.java",
        "hello-impl/src/main/java/com/example/rp/test/lagomendpoints/impl/HelloServiceImpl.java"]
};
```

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/project.ts
maysunfaisal: If thats the case, should the tests run with default log level of "debug"

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
```diff
@@ -0,0 +1,372 @@
...
+            }
+            if (setting === "mavenProfiles" || setting === "mavenProperties") {
+                if (!project_configs.mavenProfileCapabilities[projData.projectType]) {
+                    value = [];
+                    combo["eventKeys"].push("error");
+                }
+                // for spring and liberty project types support maven profiles we need to set the value here
+            }
+
+            it(`config project specification settings ${setting} to ${value}`, async () => {
```
GeekArthur: Screen Shot 2019-08-12 at 5 17 39 PM

Can you see the project setting test case for `internalPort` has been triggered when you run the tests? I can't find it when I run the tests

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/project.ts
GeekArthur: Yeah, if only `DEBUG` log level can print that log message, we need to to either set default log level to `DEBUG` or set to `DEBUG` then set it back to `INFO`

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/logging.ts
GeekArthur: Screen Shot 2019-08-12 at 5 42 59 PM

Does our product code treat this differently for different log level?

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
ssh24: Yes, that was a mistake. Good catch. Fixed it now.

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/project.ts
ssh24: Yes, I can set it to `DEBUG` by default

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
ssh24: Should be fixed now.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
ssh24: Sure, we can only do one file.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/logs.ts
```diff
@@ -0,0 +1,173 @@
...
+                const info: any = await checkNewLogFile(projectID, type);
+                expect(info);
+                expect(info.statusCode);
+                expect(info.statusCode).to.equal(400);
+                expect(info.error);
+                expect(info.error).to.haveOwnProperty("msg");
+                expect(info.error.msg).to.equal("Bad request");
+            });
+        }
+
```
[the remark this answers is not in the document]
ssh24: The `404` is actually for the inf file existence. It's a difficult edge case to test and not really related to the `checkNewLogFile` function.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
```diff
@@ -0,0 +1,206 @@
...
+                        const info: any = await updateStatus(itData);
+                        expect(info);
+                        expect(info.statusCode);
+                        expect(info.statusCode).to.equal(400);
+                        expect(info.error);
+                        expect(info.error).to.haveOwnProperty("msg");
+                        expect(info.error.msg).to.equal(`Invalid ${statusType} state: ${itData[key]}`);
+                    });
+                }
+
```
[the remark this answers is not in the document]
ssh24: We already check it on the for loop above with invalid projectID.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
ssh24: The `detailedAppStatus` is not passed on the source code of the `updateStatus` function in line https://github.com/eclipse/codewind/blob/1681bf42aebda2594837632aa8db16fc91572d7a/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts#L159

So we cannot test `detailedAppStatus` from `updateStatus`

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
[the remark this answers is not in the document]
GeekArthur: Does the loop above only check the `existence of project ID`? I mean the invalid `project ID`.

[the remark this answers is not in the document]
ssh24: Nope, it calls the `updateStatus` function without projectID. See here https://github.com/ssh24/codewind/blob/6dc536d0e045015acb92ee2185f03d06bf1612a7/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts#L98

src/pfe/file-watcher/server/test/functional-test/suites/generic/tests/logging.ts
ssh24: Yeah, I don't know why it does not print the same for trace with `warn` or `error`

GeekArthur: It's fine, I don't think it's test case issue, given it's just set log level function, we can keep an eye on this later on

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
GeekArthur: We call `updateProjectStatus` with `detailedAppStatus` when we ping app, but if we only test `updateStatus` that should be fine

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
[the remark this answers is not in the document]
GeekArthur: Yes, that's the function I check. But `without projectID` != `invalid projectID`. Invalid projectID is pass in projectID but the projectID doesn't exist. Seems like our product doesn't have that code path, I think the test should be good

GeekArthur:
@ssh24 Changes are look good to me, once you fix all my comments above you can let Rajiv double check and merge it.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
ssh24: Yes, that is correct.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/update-status.ts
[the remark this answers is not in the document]
ssh24: Yup

ssh24:
@GeekArthur All your comments are fixed. You can also approve the PR and then @rajivnathan can merge it.

GeekArthur:
LGTM, approved.

ssh24:
I have noticed something broke the che tests. I am investigating. Please do not merge.

ssh24:
The tests are now fixed. @maysunfaisal Can you please review the final commit and then we should be good to merge this.

**Local**
Screen Shot 2019-08-16 at 3 58 03 PM

**Che**
Screen Shot 2019-08-16 at 4 07 53 PM

PR #97, opened 2019-08-01
Get repositories instead of templates from extensions

makandre:

Based on recent discussion with @tobespc, extensions will now return additional template repositories to Codewind (and let Codewind read the templates from it), rather than extensions doing the parsing and returning template objects to Codewind

This PR is to handle that scenario and is required by https://github.com/kabanero-io/appsodyExtension/issues/18


makandre:
@hhellyer could you review this please?

makandre:
Re line 75, yes, that will added to `repos`, since it is passed in as the `initialValue` to the `reduce` function.

I'll add a test or 2 to make sure this is the case.

makandre:
Superseded by https://github.com/eclipse/codewind/pull/114

PR #100, opened 2019-08-02
Use github API to download project templates.

hhellyer:
This PR changes codewind to:
- Use github API to download zipped templates.
- Remove git from docker images.
- Remove simple-git from package.json files.

This removes the git executable from our codewind-pfe image and reduces the image size by around 50Mb.

### DIFF
--- a/src/initialize/src/controllers/gitController.ts
+++ b/src/initialize/src/controllers/gitController.ts
@@ -9,24 +9,109 @@
  *     IBM Corporation - initial API and implementation
  *******************************************************************************/
 
-import git from 'simple-git/promise';
 import fs from 'fs-extra';
 import path from 'path';
+import request from 'request';
+import { URL } from 'url';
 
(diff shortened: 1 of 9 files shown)


src/initialize/src/controllers/gitController.ts
```diff
@@ -9,24 +9,109 @@
...
 
 import { replacePlaceholderNameWithProjectName } from './replaceController';
+import { exec } from 'child_process';
 
 export async function importProjectFromGitRepo(repo: string, name: string, dest: string, branch?: string | undefined): Promise<void> {
   if (fs.readdirSync(dest).length > 0) {
     throw new Error('Git clone failed - directory to clone into is not empty');
   }
 
+  const tmpName = `/tmp/${branch}_${Date.now()}`;
```
liamchampton: Is this path also supported with the windows temp dir?

hhellyer: No, we only expect this to run inside a Linux container.

Issue #104, opened 2019-08-02
Codewind / PFE fails to load on Kube


johnmcollier:
I'm getting the following error in PFE after applying the fix for https://github.com/eclipse/codewind/issues/71 (https://github.com/eclipse/codewind-che-plugin/pull/47)

```
```[02/08/19 15:05:49 /portal/routes/environment.route.js] [ERROR] TypeError: Cannot read property 'spec' of undefined
    at getTektonDashboard (/portal/routes/environment.route.js:34:55)
    at process._tickCallback (internal/process/next_tick.js:68:7)
[02/08/19 15:05:53 /portal/routes/environment.route.js] [ERROR] TypeError: Cannot read property 'spec' of undefined
    at getTektonDashboard (/portal/routes/environment.route.js:34:55)
    at process._tickCallback (internal/process/next_tick.js:68:7)
[02/08/19 15:05:53 /portal/routes/environment.route.js] [ERROR] TypeError: Cannot read property 'spec' of undefined
    at getTektonDashboard (/portal/routes/environment.route.js:34:55)
    at process._tickCallback (internal/process/next_tick.js:68:7)
[02/08/19 15:05:53 /portal/routes/environment.route.js] [ERROR] TypeError: Cannot read property 'spec' of undefined
    at getTektonDashboard (/portal/routes/environment.route.js:34:55)
    at process._tickCallback (internal/process/next_tick.js:68:7)```
```

tetchel:
since no one is working today i will have a look at this

tetchel:
this happens if tekton is not installed. tekton installation should not be mandatory. i will put in a fix

tetchel:
fixed in #110

Issue #106, opened 2019-08-02
Update Codewind on Kube install docs for the Tekton integration


johnmcollier:
The Tekton integration in PFE that was recently delivered introduced new setup steps if installing Codewind on Che:
```
oc apply -f tekton/codewind-che-plugin/setup/install_che/codewind-tektonrole.yaml
oc apply -f tekton/codewind-che-plugin/setup/install_che/codewind-tektonbinding.yaml
```

Our manual install docs on https://www.eclipse.org/codewind/installoncloud.html need to be updated to reflect that.

sishida:
@johnmcollier and @tetchel 
Hi, team, I'm looking at these new commands as well as the installoncloud doc and was wondering if I could please get more info. For example, where within the installoncloud doc page do we want to include the new commands? I'm guessing we should create a new section? Thanks!

johnmcollier:
@tobespc @markcor11 Can you please help Sarah with the new install instructions for the Tekton integration? Thanks!

sishida:
@tobespc and @markcor11 
This is the only Tekton info that we currently have in Codewind documentation: https://github.com/eclipse/codewind-docs/blob/master/docs/_documentations/codewindandtektonpipelines.md
I am happy to add more details if we want to. Please just let me know. Thank you!

sishida:
@johnmcollier, @tobespc: Now that https://github.com/eclipse/codewind-docs/pull/67 has been merged, is this issue good to close? If so, could you please close it? (I'm not a committer yet.) Thank you!

johnmcollier:
@sishida Yup!

PR #110, opened 2019-08-04
Fix Environment API when Tekton is not installed

tetchel:
- Fix TypeError if tekton is not installed
- Have different statuses for success (url) vs not installed vs local (ie, n/a) vs error, so client can display better feedback
- Only look up dashboard url once instead of on every api call
- Update tests and API doc
- Rename running_on_icp field to running_on_k8s

- Fix ingress logic which errored on Openshift

Underlying tekton-getting logic is not changed. I still can't get the success case working; likely because I am not installing tekton as expected (#106 ? I did run those steps but this code still cannot find the tekton ingress).

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -51,9 +51,7 @@ paths:
                   tekton_dashboard_url:
                     type: string
                     example: '9.20.195.90.nip.io'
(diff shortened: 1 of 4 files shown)


tetchel:
Fixes https://github.com/eclipse/codewind/issues/104

Tagging @markcor11 @johnmcollier so i can assign you to review

PR #112, opened 2019-08-05
Add support for building from Msys2 on Windows

jgwest:
When running inside Msys2 on Windows, uname returns msys rather than Windows, so we check `uname -o` for msys.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -146,6 +146,11 @@ else
   export PLATFORM="-$ARCH"
 fi
 
+# If using Mingw64/Msys2 (a faux-Linux build environment on Windows), replace mingw with windows
+if [[ $HOST_OS == *"MINGW"* ]]; then
+  export HOST_OS=windows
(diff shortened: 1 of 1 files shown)


start.sh
```diff
@@ -146,6 +146,12 @@ else
   export PLATFORM="-$ARCH"
 fi
 
+# If using Msys2 on Windows, replace msys with windows
+UNAME_O=$(uname -o);
```
johnmcollier: Just as an FYI, on macOS at least, `-o` isn't an option for uname and it spits out an error
```
johns-mbp-3:codewind johncollier$ uname -o
uname: illegal option -- o
usage: uname [-amnprsv]
```

This won't prevent the script from working on macOS, but we may not want it to display that error.

jgwest: Fixed: I'm now just using the previous `uname` val.

PR #114, opened 2019-08-06
Get repositories instead of templates from extensions

makandre:

Based on recent discussion with @tobespc, extensions will now return additional template repositories to Codewind (and let Codewind read the templates from it), rather than extensions doing the parsing and returning template objects to Codewind

This PR is to handle that scenario and is required by kabanero-io/appsodyExtension#18

### DIFF
--- a/src/pfe/portal/modules/Templates.js
+++ b/src/pfe/portal/modules/Templates.js
@@ -54,7 +54,32 @@ module.exports = class Templates {
     }
 
     let newProjectTemplates = [];
-    await Promise.all(this.repositoryList.map(async function getTemplates(repository) {
+
+    // reduce function, want to take repository list and index
+    // them by url, and use that to avoid processing duplicate entries
+    const reducer = (repos, repo) => {
+      if (repo.url && !repos[repo.url])
+        repos[repo.url] = repo;
+      return repos;
+    };
+
+    // apply reduce function to create a copy of the repository list index by url
+    const repos = this.repositoryList.reduce(reducer, {});
+
+    // query providers for more repositories
+    for (let provider of Object.values(this.providers)) {
+      try {
+        const extraRepos = await provider.getRepositories();
+        if (Array.isArray(extraRepos))
+          // apply reduce function here again, urls that we already have are ignored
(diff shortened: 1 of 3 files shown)


makandre:
Hi @rwalle61, sorry I somehow messed up my old branch & PR (https://github.com/eclipse/codewind/pull/97) so I'm resubmitting my changes under this PR.

This PR contains the unit tests that you requested.

test/src/unit/template.test.js
```diff
@@ -10,6 +10,7 @@
 *******************************************************************************/
 const chai = require('chai');
 
+global.codewind = { RUNNING_IN_K8S: false };
```
rwalle61: why's this needed? and shouldn't we unset it after the test (or don't use a global) to decouple tests?

src/pfe/portal/modules/Templates.js
```diff
@@ -54,7 +54,32 @@ module.exports = class Templates {
     }
 
     let newProjectTemplates = [];
-    await Promise.all(this.repositoryList.map(async function getTemplates(repository) {
+
```
rwalle61: nit: I'd prefer this code in a subfunction, as this function is getting large

test/src/unit/template.test.js
makandre: Without this, there's an undefined error when loading the `Templates` class, which is due to https://github.com/eclipse/codewind/blob/master/src/pfe/portal/modules/utils/sharedFunctions.js#L19 trying the read the global

I'm not sure how the existing tests were able to run before, they too fail without this.

Setting `RUNNING_IN_K8S` to false explicitly is not needed tho, and I changed that.

src/pfe/portal/modules/Templates.js
makandre: moved

makandre:
@rwalle61 Ok, I removed the global.  The tests pass if they are all run during the build because of https://github.com/eclipse/codewind/blob/master/test/src/unit/projectInitializer.test.js#L20

(I was running `template.test.js` on its own hence I had to add that)

PR #144, opened 2019-08-09
Issue #143: Skip publishing docker images from non-release branch

### LINKED ISSUE #143: Skip publishing docker images from non-release branch build
This allows publishing docker images only from 'master' or release branches so that it allows for creating a collaboration branch with no impact on dockerhub images.

josephca:
fixes #143


josephca:
> The `.` in the regex is not escaped so this will match eg 19a12 or 123 or 123456, it's a small thing but it is not what was intended

Good catch. Fixed.

PR #178, opened 2019-08-12
Drop build.sh call from Jenkinsfile

### LINKED ISSUE #180: Remove direct call of build.sh from Jenkins
**Description of the enhancement:**
We need Codewind Jenkins to trigger PR builds opened by non-committers. Initially, comment trigger was considered as an option - this would allow PR build triggered if committers add a defined comment e.g. 'trigger_build' to PRs. Unfortunately, no luck after several tries. 

As a workaround, Jenkins is configured to trigger PR builds from non-committers by using 'Jenkinsfile' from the 'master' branch instead. This prevents malicious actions being added to Jenkinsfile within the PR. Currently, Jenkinsfile contains 'build.sh' calls. So, we need to remove the call to avoid undesirable actions performed against Codewind build systems.

josephca:
fixes #180


tetchel:
is there something that blocks a malicious "contributor" from modifying the Jenkinsfile in particular?

johnmcollier:
Is there anything in place to help prevent the contents of the Jenkinsfile and build.sh from getting out of sync?

josephca:
> is there something that blocks a malicious "contributor" from modifying the Jenkinsfile in particular (as opposed to the build script files)?
> 
> edit: yes, jenkinsfile from master is used if the PR is not from a committer

You've got the answer. The trust level can be configured from the Jenkins. So, no modification of Jenkinsfile from the PR is not allowed by non-committers.

josephca:
> Is there anything in place to help prevent the contents of the Jenkinsfile and build.sh from getting out of sync?

Good point. I would suggest putting a comment on the build.sh so that the change should be reflected to Jenkinsfile as well.

PR #182, opened 2019-08-12
fix recursive mkdir error

yangcao77:

The recursive mkdir is only supported in node 10.12.x and above
use fs-extra `ensureDirSync` instead


yangcao77:
@rajivnathan Can this be merged? Thanks

PR #184, opened 2019-08-13
Use dynamic port allocation for pfe in run.sh

hhellyer:
[This PR is ready to deliver but shouldn't be merged until the IDE plugins are consuming the installer version that finds the ports for our containers for them. This was delivered to the installer under: https://github.com/eclipse/codewind-installer/pull/67 but needs to be in the installer packaged with the plugins otherwise this change will break anyone testing the plugins against a development codewind build.] 

This PR starts codewind on a dynamic port during development to match the behaviour of the installer and updates the tests to dynamically find the IP address and port number for Codewind.

This is so we match the installer behaviour changed under https://github.com/eclipse/codewind-installer/pull/67 and https://github.com/eclipse/codewind-installer/issues/69

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -15,7 +15,7 @@ services:
     depends_on:
       - codewind-performance
     ports:
-      - "127.0.0.1:9090:9090"
+      - "127.0.0.1:34000-35000:9090"
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
     networks:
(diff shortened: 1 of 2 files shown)


hhellyer:
Please don't merge this until after the plugins have started using the installer to obtain the port number from a running codewind instance dynamically.

hhellyer:
@eharris369 - Is the Eclipse plugin using the installer to find the codewind url now or is it still hardcoded? (If it is I can merge this.)

hhellyer:
The Eclipse plugin has been updated, this can be merged.

PR #195, opened 2019-08-14
Remove IDC test folder

ssh24:
### Description

This PR is to remove the IDC test folder with old references. We don't use the old IDC test anymore and they have old references that needs to be pruned.


ssh24:
@maysunfaisal Can you please review this?

Issue #132, opened 2019-08-08
Appsody projects stuck in Building: Calculating difference state


eharris369:
If I create a new appsody project, disable auto build, make a change and then enable auto build, the project gets stuck in `Building: Calculating difference` state.

If changing the auto build setting is not supported for appsody projects then a capability should be added for auto build so that the IDEs know to disable it.  If no capability is added then the projects should not get stuck in this state.

tetchel:
@makandre

makandre:
What stack?    Right, auto-build is not supported.  I know we had discussed about hiding the auto-build option for appsody but I'm unsure if we actually raised that as an issue.  @sghung might know.

rajivnathan:
I just saw this happen with a Microprofile project so doesn't appear to be specific to appsody.

Steps to reproduce:
1. Clone https://github.com/microclimate-dev2ops/javaMicroProfileTemplate to the codewind workspace (Do not create using Codewind)
2. Delete the Dockerfile-build file
3. Add the project to Codewind using the "Add existing project" action
4. It should fail to build, state will be [Stopped] with a build failure status
5. Without making any other changes just try a manual build again.

Result:
You end up stuck in [Stopped] [Building - Calculating difference] state.

yangcao77:
The fix for microprofile project is PR https://github.com/eclipse/codewind/pull/199

For appsody project,  if the update failed. it should return an exit code as well. So our code can caught the error and set the build status correctly.

makandre:
@yangcao77 could you open an appsody issue with the details at https://github.com/eclipse/codewind-appsody-extension/issues?

makandre:
Actually I've identified the problem on appsody side and the cause is slightly different

yangcao77:
Is any code change required in Turbine to get appsody project issue fixed? @makandre I'm glad to help out if there's any.

makandre:
Thanks @yangcao77 I've a fix https://github.com/eclipse/codewind-appsody-extension/pull/6; it does not require Turbine changes.

sghung:
Re-opened since https://github.com/eclipse/codewind/pull/199 is not in yet

rajivnathan:
@sghung can this be closed now?

sghung:
Yes, this can be closed now

PR #199, opened 2019-08-14
Liberty project update failure cannot be caught

yangcao77:

This PR is for #132 
For microprofile projects, should return exitcode!=0 if the update failed.

### DIFF
--- a/src/pfe/file-watcher/scripts/liberty-container.sh
+++ b/src/pfe/file-watcher/scripts/liberty-container.sh
@@ -265,7 +265,9 @@ elif [ "$COMMAND" == "update" ]; then
 			$util updateBuildState $PROJECT_ID $BUILD_STATE_FAILED "buildscripts.unexpectedError"
 		elif [[ $exitCode -eq 7 ]]; then
         	$util updateBuildState $PROJECT_ID $BUILD_STATE_FAILED "buildscripts.invalidDeploymentRegistry"
-    	fi
+    	elif [[ $exitCode -eq 8 ]]; then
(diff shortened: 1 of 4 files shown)


yangcao77:
@rajivnathan Can you help review this PR?

src/pfe/file-watcher/scripts/liberty-container.sh
```diff
@@ -261,10 +261,14 @@ elif [ "$COMMAND" == "update" ]; then
 	# get the status code of the build, if the build failed, update the status
 	if [[ $exitCode -ne 0 ]]; then
 		echo "idc build failed for $ROOT with exit code $exitCode $(date)"
-		if [[ $exitCode -eq 7 ]]; then
+		if [[ $exitCode -eq 1 ]]; then
```
maysunfaisal: Dont think we should handle exit code 1 here, we can handle it in projectUtil.executeBuildScript function. The local if block is missing this error handling but it is present for K8, maybe that explains why it was seen only on local. We can add that if block for ec 1 over there; so projectUtil can handle any issues if script execution goes wrong and exits w/1

yangcao77: As long as the return code is non-zero. we need to exit a non-zero exit code. Otherwise, the `processManager.spawnDetached` won't catch the failure. 
The update case should be the same as create case: https://github.com/eclipse/codewind/blob/390e587c030d3c063093109d429a4990eb9da1ff/src/pfe/file-watcher/scripts/liberty-container.sh#L212

PR #200, opened 2019-08-15
Install built-in extensions on startup

makandre:
This is a simple initial crack at installing extensions that will be pre-packaged into Codewind (first candidate: https://github.com/eclipse/codewind-appsody-extension)

The high-level idea is:

1) In docker build, we download release zips of extensions to `/extensions` folder in the image. @sghung will submit a separate PR for this

2) When Codewind starts, the zips in `/extensions` are unzipped to `/codewind-workspace/.extensions`
   - for this initial implementation, purposely not trying to do any fancy version checking
   - if a copy of the extension already exists in `/codewind-workspace/.extensions`, it is simply backed-up by renaming it with a `__old` suffix

3) Then extensions in `/codewind-workspace/.extensions` are loaded as usual (directories with `__old` suffix are skipped)

Some of the changes here are necessary to allow the extension's directory name to be different than the extension's name in `codewind.yaml`

@sghung @hhellyer @rwalle61 @rajivnathan could someone review this PR?

### DIFF
--- a/src/pfe/portal/modules/ExtensionList.js
+++ b/src/pfe/portal/modules/ExtensionList.js
@@ -14,8 +14,15 @@ const path = require('path');
 const Extension = require('./Extension.js');
 const Logger = require('./utils/Logger');
 const ExtensionListError = require('./utils/errors/ExtensionListError');
+const utils = require('./utils/sharedFunctions');
+const { promisify } = require('util');
+const exec = promisify(require('child_process').exec);
 
 const log = new Logger(__filename);
+
+const extensionsDir = '/extensions';
+const extensionsPattern = /^(\S+)(-[\d.]+)\.zip$/; // e.g. extension-name-0.0.1.zip
+const suffixOld = '__old';
 
 /**
  * The ExtensionList class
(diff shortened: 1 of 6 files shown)


src/pfe/portal/modules/User.js
```diff
@@ -80,6 +80,15 @@ module.exports = class User {
 
       // Create the list of codewind extensions
       this.extensionList = new ExtensionList();
+
+      // Attempt to install built-in extension packages
+      try {
+        await this.extensionList.installBuiltInExtensions(this.directories.extensions);
+      } catch (error) {
+        log.error('Fail to install built-in Codewind extensions');
```
rwalle61: nit: can we have the 2 errors in one entry, to make the log slightly clearer? i.e. 
```suggestion
        log.error(`Failed to install built-in Codewind extensions. Error ${util.inspect(error)}`);
```

src/pfe/portal/modules/ExtensionList.js
```diff
@@ -14,16 +14,68 @@ const path = require('path');
...
+  
+    // get the zips from the /extensions directory
+    const entries = await fs.readdir(extensionsDir, { withFileTypes: true });
+    
+    for (let entry of entries) {
+      
+        let match;
+  
+        // look for files with names matching the expected pattern
+        if (entry.isFile() && (match = extensionsPattern.exec(entry.name))) {
```
rwalle61: Comment: I think debugging would be easier if we separated this assignment and if statement, but https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec suggests that this is a common pattern with regex

makandre:
@rwalle61 I think once we have the other change to actually inject an extension into the build, it's going to affect the tests and we'll need to make adjustments there.  I will work with @sghung to have some coverage.

src/pfe/portal/modules/User.js
makandre: And remove `log.error(error)` below?

src/pfe/portal/modules/ExtensionList.js
makandre: I'll keep it as-is then, in the interest of keeping the code looking cleaner.

PR #208, opened 2019-08-16
Support Project Https Protocol Pings

maysunfaisal:

Issue https://github.com/eclipse/codewind/issues/269

This PR enables Turbine layer to ping a project successfully if it's https enabled. This will allow the Codewind IDE plugins to open the application over https on the browser.

This feature is enabled by setting the boolean property `isHttps` in `.cw-settings`

Tested on both Local & Kube. Also updated the Portal cw-settings tests.

IDE now reports an https project as running in the overview page:
Screen Shot 2019-08-16 at 5 50 38 PM

Here is my python application over https in the browser:
Screen Shot 2019-08-16 at 5 50 44 PM

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/projectUtil.ts
+++ b/src/pfe/file-watcher/server/src/projects/projectUtil.ts
@@ -76,6 +76,7 @@ export interface ProjectEvent {
     mavenProfiles?: string[];
     mavenProperties?: string[];
     contextRoot?: string;
+    isHttps?: boolean;
 }
 
 export interface ProjectLog {
@@ -324,6 +325,9 @@ async function executeBuildScript(operation: Operation, script: string, args: Ar
     }
     if (operation.projectInfo.contextRoot) {
         projectInfo.contextRoot = operation.projectInfo.contextRoot;
+    }
+    if (typeof operation.projectInfo.isHttps == "boolean") {
+        projectInfo.isHttps = operation.projectInfo.isHttps;
     }
 
     const projectMetadata = projectsController.getProjectMetadataById(projectID);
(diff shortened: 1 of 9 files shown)


maysunfaisal:
@rajivnathan @stalleyj can you please help review the PR? Thx!

stalleyj:
@maysunfaisal The various settings are only supposed to be in either .cwSettings or the inf to prevent them getting out of sink, link contextRoot or healthcheck. I thought the IDE's would get those values from the settings file @tetchel how does the plugin get the value of contextroot ?

maysunfaisal:
@stalleyj The IDE wanted the essential project settings key to be included in .inf files.. They are updated whenever a setting value changes and during project creation. You can find this discussion from couple of months back in Portal issue 1053 and has been in the product for close to over a month now
@tetchel

maysunfaisal:
@stalleyj thx!

tetchel:
Maysun is correct, the IDEs only pick up project data such as contextRoot from the inf (`api/v1/projects`).

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -966,15 +1011,24 @@ export async function isApplicationUp(projectID: string, handler: any): Promise<
             updateDetailedAppStatus(projectID, containerInfo.ip, port, path);
         }
 
+        const isHttps: boolean = await getProjectWWWProtocolMap(projectID);
```
rajivnathan: Do we need the `getProjectWWWProtocolMap()` function? Shouldn't `projectInfo` contain the `isHttps` property? If we don't need that function we don't need the `projectWWWProtocolMap` either.

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -901,6 +907,44 @@ export async function isContainerActive(projectID: string, handler: any): Promis
     }
 }
 
+// /**
```
rajivnathan: This commented code should be removed.

PR #215, opened 2019-08-19
Add in the Kube logic for Turbine cron tests

maysunfaisal:
Added in the K8 logic for Turbine cron tests

This PR 
- creates workspace
- run Turbine functional tests
- removes workspace


maysunfaisal:
@rajivnathan @ssh24 @jingfuwang can you pls help review the PR?

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -25,24 +25,41 @@ EOF
 }
 
 function clone () {
-    curl -H "Content-type: application/json" -d '{"projectName": "'$1'", "parentPath": "'$2'", "url": "'$3'"}' "http://localhost:9090/api/v1/projects"
+    if [ $TEST_TYPE == "local" ]; then
+        curl -H "Content-type: application/json" -d '{"projectName": "'$1'", "parentPath": "'$2'", "url": "'$3'"}' "http://localhost:9090/api/v1/projects"
+    elif [ $TEST_TYPE == "kube" ]; then
+        kubectl exec -it $CODEWIND_POD_ID -- curl -H "Content-type: application/json" -d '{ "projectName": "'$1'" , "parentPath": "'$2'", "url": "'$3'" }' "https://localhost:9191/api/v1/projects" --insecure
```
GeekArthur: Do we also need to use the curl to clone projects to workspace for kube case? Think we design to use Devfile, then projects will be downloaded to workspace automatically.

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -25,24 +25,41 @@ EOF
...
+    fi
 }
 
 function setup {
     DATE_NOW=$(date +"%d-%m-%Y")
     TIME_NOW=$(date +"%H.%M.%S")
-    BUCKET_NAME=turbine-$TEST_SUITE
-    TEST_DIR=$CW_DIR/src/pfe/file-watcher/server
+    BUCKET_NAME=turbine-$TEST_TYPE-$TEST_SUITE
+    TEST_DIR=$CW_DIR/src/pfe/file-watcher/server/test
```
GeekArthur: I am think of we may need to copy all contents under `server` because user may add more devDependencies for the tests, and devDependencies is in package.json under `server` not `server/test`, then for the following code, we need to do:
1) docker/kubectl cp firstly to copy contents under `server`
2) run `npm install --only=dev"` inside PFE container

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -55,14 +72,20 @@ function setup {
         rm -rf $CW_DIR/codewind-workspace/*
```
GeekArthur: Can we do the similar thing for Che (eg. remove /projects), because we have `run` mode which we don't reinstall Codewind on Che and we may have projects exist in `/projects`

GeekArthur:
Another comment in `test.sh` file that you didn't modify, in test.sh we have code to ensure Codewind container is up and running before running test cases: https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/server/test/test.sh#L38, you may need to add the same thing for Che if on `run` mode.

src/pfe/file-watcher/server/test/resources/workspace.json
```diff
@@ -0,0 +1,67 @@
+{
+	"apiVersion": "1.0.0",
+	"metadata": {
+		"name": "codewind-plugin-env"
+	},
+	"projects": [
```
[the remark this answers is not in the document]
maysunfaisal: I will remove all the projects from dev file. We will clone projects like local because Portal does initializing, ie; they modify charts when we import.. so we need to clone for Kube, we cannot bind with template projects because of this

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: same as above. will remove from dev file and clone like local because of portal initialization in chart

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -25,24 +25,41 @@ EOF
...
 
 function clone () {
-    curl -H "Content-type: application/json" -d '{"projectName": "'$1'", "parentPath": "'$2'", "url": "'$3'"}' "http://localhost:9090/api/v1/projects"
+    if [ $TEST_TYPE == "local" ]; then
+        curl -H "Content-type: application/json" -d '{"projectName": "'$1'", "parentPath": "'$2'", "url": "'$3'"}' "http://localhost:9090/api/v1/projects"
+    elif [ $TEST_TYPE == "kube" ]; then
+        kubectl exec -it $CODEWIND_POD_ID -- curl -H "Content-type: application/json" -d '{ "projectName": "'$1'" , "parentPath": "'$2'", "url": "'$3'" }' "https://localhost:9191/api/v1/projects" --insecure
+    fi
+
+    if [[ $? -ne 0 ]]; then
```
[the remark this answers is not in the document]
maysunfaisal: fixed

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: I just changed 

1. `TEST_DIR=$CW_DIR/src/pfe/file-watcher/server` to `TEST_DIR=$CW_DIR/src/pfe/file-watcher/server/test`
2. `docker cp $TEST_DIR/test $CODEWIND_CONTAINER_ID:$TEST_DIR_CONTAINER` to `docker cp $TEST_DIR $CODEWIND_CONTAINER_ID:$TEST_DIR_CONTAINER`

we were already copying test dir

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: I dont think we can do that in Che 🤔 

If we remove `/projects`, then we have to re-create workspace..  If we re-create workspace then all user projects will be lost.

maysunfaisal:
> Another comment in `test.sh` file that you didn't modify, in test.sh we have code to ensure Codewind container is up and running before running test cases: https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/server/test/test.sh#L38, you may need to add the same thing for Che if on `run` mode.

@jingfuwang yeah i think i can add that logic in

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: In that case, I think we have to skip this for Kube

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -25,24 +25,41 @@ EOF
...
+        exit 1
+    fi
 }
 
 function setup {
     DATE_NOW=$(date +"%d-%m-%Y")
     TIME_NOW=$(date +"%H.%M.%S")
-    BUCKET_NAME=turbine-$TEST_SUITE
-    TEST_DIR=$CW_DIR/src/pfe/file-watcher/server
+    BUCKET_NAME=turbine-$TEST_TYPE-$TEST_SUITE
```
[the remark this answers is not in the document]
GeekArthur: Sounds good, keep bucket name to make it align to dashborad

src/pfe/file-watcher/server/test/scripts/exec.sh
GeekArthur: Yes, your change is for copying `server/test` dir which is same as before.

But I mean we **not only**  copy the `server/test` dir, we copy all contents under `server` dir instead because `package.json` is under `server` dir and user may need `devDependencies` in `package.json` for installing new packages that user add for testing. 

So basically we can do two things:
1 Copy the whole contents under `server` dir
2 Run `npm install --only=dev` inside PFE container **after** `docker/kubectl cp` (currently we run `npm install --only=dev` inside PFE container **before** `docker/kubectl cp`)

src/pfe/file-watcher/server/test/resources/workspace.json
```diff
@@ -0,0 +1,23 @@
...
+			"type": "chePlugin",
+			"id": "https://raw.githubusercontent.com/eclipse/codewind-che-plugin/master/plugins/codewind/codewind-sidecar/latest/meta.yaml"
+		},
+		{
+			"alias": "codewind-theia",
+			"type": "chePlugin",
+			"id": "https://raw.githubusercontent.com/eclipse/codewind-che-plugin/master/plugins/codewind/codewind-theia/latest/meta.yaml"
+		}
+	]
+}
```
ssh24: Nitpick: Need new line here.

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -25,24 +25,36 @@ EOF
 }
 
 function clone () {
-    curl -H "Content-type: application/json" -d '{"projectName": "'$1'", "parentPath": "'$2'", "url": "'$3'"}' "http://localhost:9090/api/v1/projects"
+    if [ $TEST_TYPE == "local" ]; then
+        curl -H "Content-type: application/json" -d '{"projectName": "'$1'", "parentPath": "'$2'", "url": "'$3'"}' "http://localhost:9090/api/v1/projects"
+    elif [ $TEST_TYPE == "kube" ]; then
+        kubectl exec -i $CODEWIND_POD_ID -- curl -H "Content-type: application/json" -d '{ "projectName": "'$1'" , "parentPath": "'$2'", "url": "'$3'" }' "https://localhost:9191/api/v1/projects" --insecure
```
ssh24: I am thinking can we make a function get the curl call? So in the future if the project creation parameters change we only need to change it in one place. And also dependening on if test type is kube or local, it will dynamically select the protocol and ports.

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -25,24 +25,36 @@ EOF
...
+    mkdir -p $TEST_OUTPUT_DIR
+
+    if [ $TEST_TYPE == "local" ]; then
+        CODEWIND_CONTAINER_ID=$(docker ps | grep codewind-pfe-amd64 | cut -d " " -f 1)
+        docker cp $TURBINE_SERVER_DIR $CODEWIND_CONTAINER_ID:$TURBINE_DIR_CONTAINER \
+        && docker exec -it $CODEWIND_CONTAINER_ID bash -c "cd /file-watcher/server; npm install --only=dev"
+    elif [ $TEST_TYPE == "kube" ]; then
+        CODEWIND_POD_ID=$(kubectl get po --selector=app=codewind-pfe --show-labels | tail -n 1 | cut -d " " -f 1)
+        kubectl cp $TURBINE_SERVER_DIR $CODEWIND_POD_ID:$TURBINE_DIR_CONTAINER \
+        && kubectl exec -it $CODEWIND_POD_ID -- bash -c "cd /file-watcher/server; npm install --only=dev"
```
ssh24: Can we make the command `"cd /file-watcher/server; npm install --only=dev"` a variable and call it for both docker and kube exec? So in the future if it changes we only need to update it in one place

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -72,8 +92,13 @@ function setup {
 }
 
 function run {
-    docker exec -it $CODEWIND_CONTAINER_ID bash -c "cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml npm run $TEST_SUITE:test:xml" | tee $TEST_LOG
-    docker cp $CODEWIND_CONTAINER_ID:/test_output.xml $TEST_OUTPUT
+    if [ $TEST_TYPE == "local" ]; then
+        docker exec -it $CODEWIND_CONTAINER_ID bash -c "cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml npm run $TEST_SUITE:test:xml" | tee $TEST_LOG
+        docker cp $CODEWIND_CONTAINER_ID:/test_output.xml $TEST_OUTPUT
+    elif [ $TEST_TYPE == "kube" ]; then
+        kubectl exec -it $CODEWIND_POD_ID -- bash -c "cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml npm run $TEST_SUITE:test:xml" | tee $TEST_LOG
```
ssh24: We can also make `"cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml npm run $TEST_SUITE:test:xml"` a variable and call it in both docker and kube exec

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: tried it, it did not work

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: tried it, it did not work

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: no, the curl needs to run inside the kube bash terminal

ssh24: I meant something like this:
```
function getCurl() {
  protocol="http"
  port="9090"
  if [ $1 == "kube" ]; then
      protocol="https"
      port="9091"
      flags="--insecure"
  fi
  curl -H "Content-type: application/json" -d '{"projectName": "'$2'", "parentPath": "'$3'", "url": "'$4'"}' "$protocol://localhost:$port/api/v1/projects" $flags
}

Need to test the function out but hopefully you get the idea behind it?

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: What was the error?

maysunfaisal:
@ssh24 @jingfuwang pls help review when you have time, thx. I have updated the PR with feedback.

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: done

Issue #164, opened 2019-08-09
Performance Dashboard Design Enhancements and Bug Fixes - User Feedback


thisguy-1:
Run Load test needs to have its settings set the first time is set
- [ ] What the setting were for each run load test needs to be able to be looked up
- [ ] The colors associated with each measurement needs to be more obvious
- [ ]  What the setting were for the latest run test needs to be on screen
- [x] Solve the issue of when to lines overlap in the graph
- [x] "Unable to render" error
        - Fix bug
       - Trouble shoot instructions or link
        - Add refresh button

@tobespc @thisguy-1

markcor11:
PR: https://github.com/eclipse/codewind/pull/202

kimholmes:
@tobespc

mii-w:
In Progress Wire Frames for tomorrow's meeting reference
<img width="908" alt="Screen Shot 2019-12-09 at 4 32 40 PM" src="https://media.github.ibm.com/user/207956/files/b2338880-1aa1-11ea-88ac-194b9bd24ffc">
<img width="718" alt="Screen Shot 2019-12-09 at 4 32 49 PM" src="https://media.github.ibm.com/user/207956/files/b3fd4c00-1aa1-11ea-9480-1b3ca85bb4c6">
<img width="729" alt="Screen Shot 2019-12-09 at 4 32 56 PM" src="https://media.github.ibm.com/user/207956/files/b65fa600-1aa1-11ea-8915-710cf8a2f6e3">
<img width="728" alt="Screen Shot 2019-12-09 at 4 33 01 PM" src="https://media.github.ibm.com/user/207956/files/b8296980-1aa1-11ea-875a-fd59aab1f1ab">
<img width="723" alt="Screen Shot 2019-12-09 at 4 33 08 PM" src="https://media.github.ibm.com/user/207956/files/bb245a00-1aa1-11ea-807a-3ea4320211cd">
<img width="843" alt="Screen Shot 2019-12-09 at 4 33 18 PM" src="https://media.github.ibm.com/user/207956/files/bd86b400-1aa1-11ea-9e30-0dfa20001cad">
<img width="842" alt="Screen Shot 2019-12-09 at 4 33 26 PM" src="https://media.github.ibm.com/user/207956/files/bf507780-1aa1-11ea-81e1-4fa18344efd1">
[...]

PR #225, opened 2019-08-20
Handle path changes when updating load test preferences

markcor11:
Originated from Issue: https://github.com/eclipse/codewind/issues/164 

# Problem

After changing the PATH field under the load test preferences dialog,  clicking save closes the settings dialog box,  however the chart does will refresh and instead displays the red error handler "Unable to render". 

Screenshot 2019-08-20 at 15 40 00

# Workaround 

Manually refresh the browser page

# Solution

* This PR ensures that when changing the path,  the counter filter is re-applied using the new path value. 
* Since each project type produces path URLs formatted differently,  (Java for example includes protocol and port information, some include query strings etc where as NodeJS does not) the change also normalises the path into its simplest form by stripping off the server/port and query string.

# Additions 

With the chart now showing just HTTP plots for a specific path, these changes ensure non-HTTP counters such as CPU and Memory snapshots are also excluded from the charts if they were measured for an endpoint other than the filtered path. By synchronising the metrics across all types and filtering them by path, the charts behaviour is more predictable. 

Ran existing test bucket : 

Test Suites: 6 passed, 6 total
Tests:       41 passed, 41 total

### DIFF
--- a/src/performance/dashboard/src/modules/MetricsUtils.js
+++ b/src/performance/dashboard/src/modules/MetricsUtils.js
@@ -208,11 +208,7 @@ let getPathFromURL = function (url) {
  */
 let buildChartDataHTTP = function (params) {
 
-    const scaleFactor = params.scaleFactor;
-    const counterName = params.counterName;
-    const urlFilter = params.urlFilter;
-    const decimals = params.decimals;
-    const filteredData = params.filteredData;
+    const {scaleFactor, counterName, urlFilter, decimals, filteredData} = params;
 
     let columnData = [];
     columnData.push(counterName);
@@ -256,7 +252,8 @@ let getHTTPHitTimestamps = function (metrics, filteredUrl) {
             if (urlRow && urlRow.url) {
                 let uri = getPathFromURL(urlRow.url);
                 return getEndpoint(uri) === filteredUrl;
-            } return false;
+            } 
+            return false;
         });
         if (foundUrlSamples) {
             timestamps.push(snapshot.time);
--- a/src/performance/dashboard/src/modules/MetricsUtils.js
+++ b/src/performance/dashboard/src/modules/MetricsUtils.js
@@ -11,7 +11,7 @@
 
 /**
  * Returns the counter names for each fo the application metrics engines
- * @param {string} appMetricsName 
+ * @param {string} appMetricsName
  */
 let getLanguageCounters = function (appMetricsName) {
     switch (appMetricsName) {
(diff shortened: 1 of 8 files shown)


src/performance/dashboard/src/modules/MetricsUtils.js
```diff
@@ -218,125 +208,92 @@ let getPathFromURL = function (url) {
  */
 let buildChartDataHTTP = function (params) {
 
-    const mcAPIData = params.projectMetrics;
     const scaleFactor = params.scaleFactor;
```
rwalle61: nit: why not
```suggestion
    const { scaleFactor, counterName, urlFilter, decimals, filteredData } = params;
```
?

src/performance/dashboard/src/modules/MetricsUtils.js
```diff
@@ -218,125 +208,92 @@ let getPathFromURL = function (url) {
...
-            return snapshot["time"]
-        }));
+    const http = metrics.find(metric => { return metric.type === "http" });
+    http.metrics.forEach(snapshot => {
+        const urlList = snapshot.value.data;
+        const foundUrlSamples = urlList.find(urlRow => {
+            if (urlRow && urlRow.url) {
+                let uri = getPathFromURL(urlRow.url);
+                return getEndpoint(uri) === filteredUrl;
+            } return false;
```
rwalle61: nit:
```suggestion
            } 
            return false;
```

src/performance/dashboard/src/modules/MetricsUtils.js
```diff
@@ -218,125 +208,92 @@ let getPathFromURL = function (url) {
...
-
+    // merge all the metrics into the data model using a timestamp key
     metrics.forEach(metricType => {
         metricType.metrics.forEach(snapshot => {
-            let timestamp = model.find(element => { return element['time'] === snapshot['time'] })
-            let metricsGroup = {}
-            metricsGroup.type = metricType.type;
-            metricsGroup.value = snapshot;
-            timestamp[metricType.type] = metricsGroup;
+            const timestamp = model.find(element => { return element['time'] === snapshot['time'] });
```
rwalle61: nit: why not
```suggestion
            const timestamp = model.find(element => { return element.time === snapshot.time });
```

src/performance/dashboard/src/modules/MetricsUtils.js
markcor11: > question: why not

that would work too,  likely they didn't always have the same name

Issue #221, opened 2019-08-20
Add the portal functional tests to the Jenkinsfile


stalleyj:
**Description of the enhancement:**

There are currently no tests run during a PR build. 

**Proposed solution:** 

We should add running of the portal functional tests to the PR builds and not upload the docker images if the tests fail.

tobespc:
PR added that adds the tests to the jenkinsfile

PR #230, opened 2019-08-20
Enable tests to run in Jenkins on a PR

tobespc:
This PR enables the Jenkins build to run the tests associated with codewind when a PR is submitted

This is to resolve issue https://github.com/eclipse/codewind/issues/221


Jenkinsfile
```diff
@@ -81,6 +81,109 @@ pipeline {
...
+                        chmod +x ~/docker-compose
+
+                        # Run eslint on portal code
+                        cd src/pfe/portal
+                        npm install
+                        npm run eslint
+                        if [ $? -ne 0 ]; then
+                            exit 1
+                        fi
+                        cd ../../..
```
rajivnathan: Is there an environment variable for the codewind directory that we can use to change directory instead of relative paths? If not we can create one before `cd src/pfe/portal`. It'd help make the Jenkinsfile more robust in case other changes are added in between in the future.

Jenkinsfile
```diff
@@ -81,6 +81,109 @@ pipeline {
...
+
+                        # Check to see if any containers exited straight away
+                        printf "\n\n${BLUE}CHECKING FOR codewind CONTAINERS THAT EXITED STRAIGHT AFTER BEING RUN $RESET\n";
+                        EXITED_PROCESSES=$(docker ps -q --filter "name=codewind" --filter "status=exited"  | wc -l)
+                        if [ $EXITED_PROCESSES -gt 0 ]; then
+                          printf "\n${RED}Exited containers found $RESET\n";
+                          # docker ps --filter "name=codewind" --filter "status=exited";
+                          NUM_CODE_ZERO=$(docker ps -q --filter "name=codewind" --filter "status=exited" --filter "exited=0" | wc -l);
+                          NUM_CODE_ONE=$(docker ps -q --filter "name=codewind" --filter "status=exited" --filter "exited=1" | wc -l);
+                          if [ $NUM_CODE_ZERO -gt 0 ]; then
```
rajivnathan: Should this be `-eq` instead of `-gt`?

Jenkinsfile
```diff
@@ -81,6 +81,109 @@ pipeline {
...
+                            printf "\n\n${RED}FAILED TO START CONTAINERS $RESET\n";
+                            exit;
+                        fi
+
+                        printf "\n\nPausing for 20 seconds to allow containers to start\n";
+                        sleep 20;
+
+                        # Check to see if any containers exited straight away
+                        printf "\n\n${BLUE}CHECKING FOR codewind CONTAINERS THAT EXITED STRAIGHT AFTER BEING RUN $RESET\n";
+                        EXITED_PROCESSES=$(docker ps -q --filter "name=codewind" --filter "status=exited"  | wc -l)
```
rajivnathan: Think we need the `-a` flag to find exited containers here and in the other uses of `docker ps` below?

Jenkinsfile
tobespc: there isn't right now but I do agree we could improve on this.   Ideally, if we can get this in and working as is, we can iterate over it and make changes

Jenkinsfile
tobespc: These are the flags we are currently using in the build and test scripts (not made any changes)

tobespc: but we want to check for having 1 or more exited containers

Issue #211, opened 2019-08-19
Appsody Needs to be automatically installed


sghung:
**Codewind version:**
Master

**OS:**
Mac

**Description:**
This bug is to track the stop ship requirement of having the appsody extension automatically part of the codewind installation

**Steps to reproduce:**
Starting codewind does not have appsody or its stacks available

**Workaround:** 
Manual install, which is not the preferred user experience

sghung:
Changes have been merged into master

PR #231, opened 2019-08-20
Automatically install appsody into Codewind

sghung:
Downloads the appsody zip from the release. I don't use "latest" and choose a static version to ensure the version used in codewind is the version we want.

The zip is copied into the extensions folder into the pfe container. Logic already exists to pickup this zip and extract it into the codewind workspace .extensions directory.

Verified that with a deleted codewind workspace, the appsody extension gets installed with this code change.

@makandre can you help take a look too? Thank you

### DIFF
--- a/script/build.sh
+++ b/script/build.sh
@@ -56,7 +56,7 @@ cp -r $DIR/docs ${SRC_DIR}/pfe/portal/
 # Copy the appsody extension into portal. The zip file must have a version number e.g. codewind-appsody-extension-0.3.0.zip
 # in order for it to be accepted as a valid extension
 mkdir -p ${SRC_DIR}/pfe/extensions
-rm -f ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-0.3.0.zip
+rm -f ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-*.zip
 curl -Lo ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-0.3.0.zip https://github.com/eclipse/codewind-appsody-extension/archive/0.3.0.zip
 
 # BUILD IMAGES
--- a/script/build.sh
+++ b/script/build.sh
@@ -55,6 +55,7 @@ cp -r $DIR/docs ${SRC_DIR}/pfe/portal/
 
 # Copy the appsody extension into portal. The zip file must have a version number e.g. codewind-appsody-extension-0.3.0.zip
 # in order for it to be accepted as a valid extension
+echo -e "\n+++   DOWNLOADING EXTENSIONS   +++\n";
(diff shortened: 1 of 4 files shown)


script/build.sh
```diff
@@ -53,6 +53,12 @@ cp -r $DIR/NOTICE.md ${SRC_DIR}/performance/
 # Copy the docs into portal
 cp -r $DIR/docs ${SRC_DIR}/pfe/portal/
 
+# Copy the appsody extension into portal. The zip file must have a version number e.g. codewind-appsody-extension-0.3.0.zip
+# in order for it to be accepted as a valid extension
+mkdir -p ${SRC_DIR}/pfe/extensions
+rm -f ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-0.3.0.zip
```
makandre: should we use a wildcard here and not remove specific version zip?

script/build.sh
```diff
@@ -53,6 +53,12 @@ cp -r $DIR/NOTICE.md ${SRC_DIR}/performance/
...
+curl -Lo ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-0.3.0.zip https://github.com/eclipse/codewind-appsody-extension/archive/0.3.0.zip
```
makandre: I guess we need to update this file to pick up a newer release (as opposed to grabbing a latest release)

Don't really have have a strong opinion on this but just want to point it out.

src/pfe/Dockerfile_x86_64
```diff
@@ -159,6 +159,9 @@ RUN chmod -R +rx /file-watcher/idc
 
 WORKDIR /codewind-workspace
 
+WORKDIR /extensions
```
makandre: since you used absolute path below I don't think the `WORKDIR` change here is necessary

src/pfe/Dockerfile_x86_64
```diff
@@ -159,6 +159,9 @@ RUN chmod -R +rx /file-watcher/idc
...
+COPY /extensions /extensions
```
makandre: This `COPY` is failing in the build.  I think source path needs to be `src/pfe/extensions`?

sghung:
Thanks @makandre 

- Good point for removing a wildcard, I have made changes to do so
- The exact version is intentional. I don't think we should auto install the latest appsody extension by default in case there's any compatibility issues
- For the dockerfile, I moved the extensions copy to the top and removed the workdir call

makandre:
@sghung The `COPY` still appears to be failing.  I believe the source path is wrong

sghung:
I had missed updating the Jenkins file which mirrors the build.sh script. After adding that the build is green.

@tobespc can you review? Thank you

jgwest:
Issue is #211

PR #246, opened 2019-08-21
Fix circular dependency from logHelper to libertyProject

ssh24:
## Description

We should avoid the circular dependecy from `logHelper` module to `libertyProject` module. Moved the variable to exists in `logHelper` so that the dependency is one way only.


ssh24:
@rajivnathan Can you help merge this? Currently breaking our tests.

Issue #54, opened 2019-07-29
Auto Build is not working in Codewind for VSCode and Eclipse


johnmcollier:
**Codewind Version:** Latest Development Version
**Eclipse Version:** Eclipse IDE for Enterprise Java Developers 4.11.0
**VS Code Version:** Code 1.35.1

**Operating Systems:**
- Mac Mojave 10.14.5
- VM with Windows 10

In VS Code, if you click on "AutoBuild", stop Codewind and start Codewind again, you will receive some error messages indicating that no logs were found and a 404 error in the file watcher. After that, if you try to enable the “Auto Build” option again, it will not work. In addition, the project never builds again and remains in an “Unknown” state. 
![image](https://media.github.ibm.com/user/210489/files/6f69ef80-9432-11e9-95b8-2e0dba2eec86)

In Eclipse, the same thing happens, but there are no errors on the screen and the projects have a status of “No build information”.
![image](https://media.github.ibm.com/user/210489/files/77299400-9432-11e9-8d87-65e16a00c37c)

elsony:
The cause of this problem is due to createProject has never been called by portal to turbine on the project if autobuild has been turned off.  When autobuild is being turned on again after Codwind restart, portal just call build instead of createProject on the first time build. Based on the last design discussion with @tobespc, the conclusion is to handle it on the portal side to make sure the first time build (e.g. via manual build or switching autobuild from disable to enable), portal will call createProject on the first time build instead of just calling build if portal has never call createProject on turbine to register the project before. Therefore, I am removing the iterative-dev area label since this will be handled by portal component.

tetchel:
It doesn't even look like the user can recover from this; I tried disable/enable but it still got stuck Unknown, so you have to remove and then re-bind the project.

stalleyj:
I am working on a fix for this which is 90% of the way there but still has a few issues.  The logs/capabilities/build action apis are all working with projects that have autoBuild = false.

stalleyj:
PR is ready for review https://github.com/eclipse/codewind/pull/330

rajivnathan:
Discussed this today with Elson, Toby and Julie. Came to the consensus that when Codewind starts up again, all **enabled** projects will be built once irrespective of whether auto build is on/off. The auto build setting only applies to subsequent builds. Users still have the option of disabling a project if they do not want it to be built and started.

tobespc:
this is now delivered

tetchel:
[...]

PR #330, opened 2019-08-29
Fix problems with autobuild setting and affected APIs

stalleyj:
This PR fixes https://github.com/eclipse/codewind/issues/54 so that a project with autobuild set to false will behave as expected. 

The solution was to always call createProject on Turbine even when autobuild is false so that Turbine will set up project metadata and log directories but then only create a build operation when the flag is set to true.  This fixes problems with the capabilities/logs/build apis not working on projects that hadn't been built. problems were particularly encountered when restarting Codewind with a project having autobuild false. I have a clean run of the portal API tests and 4 failures on the Turbine unit tests but these also occurred on master branch.

### DIFF
--- a/src/pfe/portal/modules/User.js
+++ b/src/pfe/portal/modules/User.js
@@ -291,7 +291,7 @@ module.exports = class User {
       for (let i = 0; i < projectArray.length; i++) {
         let project = projectArray[i];
         // Don't request build if autoBuild is false
-        if (project.isOpen() && project.autoBuild) {
+        if (project.isOpen()) {
           this.buildAndRunProject(project);
         }
       }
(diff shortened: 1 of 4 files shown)


src/pfe/portal/modules/User.js
```diff
@@ -291,7 +291,7 @@ module.exports = class User {
       for (let i = 0; i < projectArray.length; i++) {
         let project = projectArray[i];
         // Don't request build if autoBuild is false
-        if (project.isOpen() && project.autoBuild) {
+        if (project.isOpen()) {
```
rwalle61: As far as I can tell, this change means that even if the user has `autoBuild` turned off, the projects will `buildAndRun` when PFE starts. Is that intended?

yangcao77:
This pr does not fix the issue. 
Turbine would call `update` for a manual build. If the project did not build on first time calling `projectCreate`. It would fail.

Portal should call `createProject` whenever the project requires a first time build. Simply set the autobuild flag to false won't work.

rajivnathan:
Discussed with Elson, Toby and Julie this morning. This is the summary: https://github.com/eclipse/codewind/issues/54#issuecomment-527507005

tobespc:
closing as this issue has had a rethink on design

PR #341, opened 2019-08-30
Docs: Create Project from Template 200 response fixes

mattcolegate:
This PR updates:
- The 200 response message to show it's the request that's successful, not the project creation
- The response bodies now have correct status, so that the `InitializeError` schema doesn't have a `status` with a value of `success` - most confusing!

Tested locally by using a Swagger viewer.


docs/openapi.yml
```diff
@@ -102,7 +102,7 @@ paths:
                   example: "https://github.com/microclimate-dev2ops/nodeExpressTemplate"
       responses:
         200:
-          description: Project creation successful
+          description: Project creation request sent successfully
```
tobespc: should this be request received successfully ?

mattcolegate: That's a better way of putting it, and I've gone one step further to 'Project creation request received and processed' because we've done the work and the user needs to check the response to see if it was successful or not

Issue #311, opened 2019-08-28
SVT:Open container shell does not work for some of appsody projects started in debug mode


jagraj:
**Codewind version:** 0.3
**OS:** WIndows 10

**Che version:**
**IDE extension version:** latest (0.3.0.v201908161735)
**IDE version:** 
Eclipse IDE for Enterprise Java Developers.
Version: 2019-06 (4.12.0)
Build id: 20190614-1200
OS: Windows 10, v.10.0, x86_64 / win32
Java version: 1.8.0_201

**Kubernetes cluster:**
**Description:**

Open container shell does not open if some of the appsody projects started in debug mode.

```
Error: No such container: 9baabef6358e806afaa16fee1788ff7fde01a75929be69188212e96d91531411
```

![Screen Shot 2019-08-28 at 10 59 22 AM](https://user-images.githubusercontent.com/713849/63867502-f41d4080-c982-11e9-90cf-7d62ab504772.png)

**Steps to reproduce:**
1.  Install Codewind with the latest images.
2. Create Appsody spring, nodejs projects.
3. Restart in debug mode.
4. Click on "Open container shell"  then you get message saying no container. 

**Workaround:**

jagraj:
When I click on Java appsody project and open container shell and it results no such container. Java Microprofile project says it is running but there is no container when I list the containers from command line.

```
Error: No such container: aec950565b6487b50c86d594d2248f5bbeb7270f123b33d32916a7d9aa0678a1
```

![image](https://user-images.githubusercontent.com/713849/64447068-78b93e80-d0a8-11e9-9285-c4c3d9b55bf8.png)


```
docker ps -a
CONTAINER ID        IMAGE                                                    COMMAND                  CREATED             STATUS              PORTS                                                                         NAMES
bdae1506045b        cw-lagomcw03svt-8632b130-cb1d-11e9-aad9-c5b40bd66875     "java -cp /deploymen…"   4 minutes ago       Up 4 minutes        0.0.0.0:32827->8778/tcp, 0.0.0.0:32826->9000/tcp, 0.0.0.0:32825->9779/tcp     cw-lagomcw03svt-8632b130-cb1d-11e9-aad9-c5b40bd66875
[...]

PR #348, opened 2019-09-03
Port Include new container id in restart event to 0.3.1

sghung:
Port of https://github.com/eclipse/codewind/pull/318/files

Tests done: verified that you can shell into the container after restarting various appsody projects into debug mode.


sghung:
This port is for fixing https://github.com/eclipse/codewind/issues/311 in 0.3.1

Issue #344, opened 2019-08-30
SVT: Intermittently Open liberty project fails on codewind che


jagraj:
**Codewind version:** 0.3
**OS:** RHEL

**Che version:** 0.3
**IDE extension version:** Theia 0.3
**IDE version:** Theia - RC4
**Kubernetes cluster:**  OpenShift

**Description:**
Open liberty project intermittently fails to build from Codewind che.  It tries to use lagom projects maven cache and fails to build. 

```
"projectID": "38d7b5b0-caab-11e9-ba5f-493060042e75",
  "appStatus": "stopped"
}
^[[91m[29/08/19 22:20:51 cw-olcc03svt-38d7b5b0-caab-11e9-ba5f] [ERROR]^[[39m ^[[31m[ File Name: /file-watcher/server/dist/projects/projectUtil.js | Function Name: containerBuildAndRun | Line Number: 1294] ^[[39mBuild image stage failed for olcc03svt
^[[91m[29/08/19 22:20:51 cw-olcc03svt-38d7b5b0-caab-11e9-ba5f] [ERROR]^[[39m ^[[31m[ File Name: /file-watcher/server/dist/projects/projectUtil.js | Function Name: containerBuildAndRun | Line Number: 1295] ^[[39m[object Object]
^[[91m[29/08/19 22:20:51 FileWatcher.js] [ERROR]^[[39m projectStatusChanged: failed (project: {
  "projectID": "38d7b5b0-caab-11e9-ba5f-493060042e75",
  "buildStatus": "failed",
  "detailedBuildStatus": "Creating image for build failed",
  "lastbuild": 1567117251564
})
^[[91m[29/08/19 22:20:51 olcc03svt] [ERROR]^[[39m containerBuildAndRun : 1287 | | The command /file-watcher/scripts/dockerScripts/docker-build.sh bud --format docker --layers --label builtBy=codewind -t cw-olcc03svt-38d7b5b0-caab-11e9-ba5f,/codewind-workspace/olcc03svt,/codewind-workspace/.logs/olcc03svt-38d7b5b0-caab-11e9-ba5f-493060042e75/docker.build.log,38d7b5b0-caab-11e9-ba5f-493060042e75 failed with exit code 1
^[[32m[29/08/19 22:20:51 olcc03svt] [INFO]^[[39m Build state changed for project: 38d7b5b0-caab-11e9-ba5f-493060042e75 from: inProgress, to: failed, with message: buildscripts.buildFail
[...]

PR #349, opened 2019-09-03
[0.3] Disable layer caching for Generic Docker Java projects

johnmcollier:
Fixes https://github.com/eclipse/codewind/issues/344

Due to a [bug in buildah](https://github.com/containers/buildah/issues/1830), we've found that when Open Liberty and Lagom projects are built in succession, buildah will use the wrong pom.xml and cause a build failure. This appears to be restricted to just Java generic Docker projects. Disabling the layer caching for the build prevents the error from showing up.

So, in order to resolve this issue and to keep layer caching for other build types (where this bug isn't seen), this PR disables layer caching on generic Docker projects with `Java` as the language.


johnmcollier:
Verified it fixed the issue seen in #344, and also verified that the layer caching for other projects are not affected

PR #367, opened 2019-09-04
Check debug port in getCapabilities

### LINKED ISSUE #363: Appsody python debug does not work and its option should be removed
**Codewind version:** 0.3.1
**OS:** Mac

**Description:**
Appsody added the python stack. For the debug scenario, debugging is not supported on python and its capabilities will need to be removed as part of https://github.com/eclipse/codewind/pull/367. The user can still restart in debug mode, but debugging will not actually work.

https://github.com/eclipse/codewind-appsody-extension/pull/20 was to set the debug and app port for the newly added python project, so the app would be detected as started. 

**Steps to reproduce:**
Create the python template, restart in debug mode, and debugging does not work

**Workaround:** 
None

makandre:

Resolves #363

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
+++ b/src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
@@ -277,7 +277,7 @@ export class ShellExtensionProject implements IExtensionProject {
      */
     getCapabilities = (): ProjectCapabilities => {
 
-        if (this.config.capabilities) {
+        if (this.config.capabilities && this.getDefaultDebugPort()) {
             return new ProjectCapabilities(
                 this.config.capabilities.startModes, this.config.capabilities.controlCommands);
         }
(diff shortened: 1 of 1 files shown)


makandre:
All the other appsody projects have debug ports that are defined in https://github.com/eclipse/codewind-appsody-extension/blob/master/.sh-extension#L13

elsony:
Restart is defined as part of the custom capability.  Does it mean Python will not support restart? Is this the expected behaviour?

makandre:
Yes, was making it on par with Codewind python projects (no restart option):

image

elsony:
As discussed, we don't want to disable general restart function to match the rest of the other Appsody project. Therefore, this PR will be changed to just disable the debug function on Appsody Python project instead of both debug and restart.

makandre:
@elsony requested changes are made.

PR #373, opened 2019-09-05
Update Appsody extension to 0.3.1

sghung:
Appsody added python and the extension needs to react to that change, otherwise the app port could not be detected. Once this is approved and merged, I will port the changes to master.


sghung:
@tobespc can you help review this one? Thank you

Issue #243, opened 2019-08-21
SVT: Add Existing Open Liberty project fails - missing files


mmulholl:
**Codewind version:** eclipse  - latest 
**OS:** Mac

**Description:**

If I create an open liberty project (openLiberty1) all is good. Remove the project from codewind but leave on disk. Now try adding it back using add existing project wizard but the project fails to build because of missing files. See this info in .projects folder:

  "validationStatus": "failed",
  "validationResults": [
    {
      "severity": "error",
      "filename": "Dockerfile-build",
      "filepath": "openLiberty1/Dockerfile-build",
      "type": "missing",
      "label": "Missing required file",
      "details": "Dockerfile-build is required but was not found.",
      "quickfix": {
        "fixID": "generateMissingFiles",
        "name": "Generate missing files",
        "description": "Generate missing required files."
      }
    },
    {
      "severity": "error",
      "filename": "pom.xml",
      "filepath": "openLiberty1/pom.xml",
      "type": "invalid",
      "label": "A maven profile named 'microclimate' was not found.",
      "details": "A microclimate build profile is recommended for Microprofile projects. See the project import documentation for details."
    }
  ],

**Steps to reproduce:**

1. Create an open liberty project
2. Remove project from codewind but leave on disk.
3. Use add existing project wizard to add project back to codewind.
 
**Workaround:** 

When binding the existing project again select "No" and then select "Other" for the project type.

mmulholl:
Added priorty/stopship:

Think this would be a common scenario. Work on the project, remove from codewind, sometime later start work on it again.

mmulholl:
Just noticed when adding an existing Liberty Profile project the type is set to  "Microprofile / Java EE" by default. If the default is left the problem happens. However if you set the type to "Other (Basic Container)" the project imports successfully.

johnmcollier:
@mmulholl I believe that's expected currently. The Open Liberty template can't currently be used as a `Microprofile / Java EE` project, but rather must be used as a generic Docker / `Basic container` project.

The `Microprofile / Java EE` project type requires WebSphere Liberty at the moment I believe.

johnmcollier:
I believe @rajivnathan was looking at getting it running as a Liberty project, but don't believe that's done yet. I'll let him confirm

rajivnathan:
[...]

PR #374, opened 2019-09-05
Openliberty project Recognition

ebuckle:
- Updating Liberty project detection to use dockerfile
- Updating project type detection tests to acknowledge new changes

As per issue #243


stalleyj:
@ebuckle I'm afraid this code now lies the cli, so you need to put a fix in there instead.

ebuckle:
No problem, implementing now

hhellyer:
This is actually needed in the installer now as this function has moved there.

Issue #125, opened 2019-08-07
Extend Turbine PFE Functional Tests - Maven Projects (Liberty and Spring) 


ssh24:
**Description of the enhancement:** Once https://github.com/eclipse/codewind/issues/48 has been completed for the docker projects, we can go ahead and extend the PFE functional tests for maven project types.

**Proposed solution:** The extended functional tests for maven project types should comply with the following list of functions (can change if new functions are added, check `index.ts` under `src/pfe/file-watcher/server/src/`) and work for both local and hybrid cases.
- [x] createProject
- [x] deleteProject
- [x] setLocale
- [x] setLoggingLevel
- [x] readWorkspaceSettings
- [x] testDeploymentRegistry
- [x] deploymentRegistryStatus
- [x] registerListener
- [x] getProjectTypes
- [x] getProjectCapabilities
- [x] projectAction
- [x] projectSpecification
- [x] logs
- [x] shutdown
- [x] updateStatus
- [x] checkNewLogFile
- [x] updateProjectForNewChange

PR #387, opened 2019-09-05
Extend Turbine Functional Test - Spring project

ssh24:
Related issue: #125

### Description

This PR extends the the turbine functional test suite for spring project type.

**Local result**:
Screen Shot 2019-09-05 at 3 31 17 PM
https://9.37.138.217/results.html?bucket=turbine-local-functional&build=test-190906045834

**K8 result**: Screen Shot 2019-09-05 at 3 49 11 PM
https://9.37.138.217/results.html?bucket=turbine-kube-functional&build=test-190906121242

**Unit test result**: 
Screen Shot 2019-09-05 at 3 51 00 PM

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
+++ b/src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
@@ -552,11 +552,12 @@ const changeMavenProfiles = async function (mavenProfiles: any, operation: Opera
 
     let arrayIsNotString = false;
     if (mavenProfiles instanceof Array) {
-        mavenProfiles.forEach(function(item: any): void {
-           if (typeof item !== "string") {
-               arrayIsNotString = true;
-           }
-        });
+        for (const profile of mavenProfiles) {
+            if (typeof profile !== "string") {
+                arrayIsNotString = true;
+                break;
+            }
+        }
     }
 
     if (mavenProfiles && arrayIsNotString) {
@@ -656,11 +657,12 @@ const changeMavenProperties = async function (mavenProperties: any, operation: O
 
     let arrayIsNotString = false;
     if (mavenProperties instanceof Array) {
-        mavenProperties.forEach(function(item: any): void {
-           if (typeof item !== "string") {
-               arrayIsNotString = true;
-           }
-        });
+        for (const property of mavenProperties) {
+            if (typeof property !== "string") {
+                arrayIsNotString = true;
+                break;
+            }
+        }
     }
 
     if (mavenProperties && arrayIsNotString) {
--- a/src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
+++ b/src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
@@ -550,7 +550,16 @@ const changeMavenProfiles = async function (mavenProfiles: any, operation: Opera
         return;
     }
 
-    if (mavenProfiles && !(mavenProfiles instanceof Array)) {
+    let arrayIsNotString = false;
+    if (mavenProfiles instanceof Array) {
+        mavenProfiles.forEach(function(item: any): void {
+           if (typeof item !== "string") {
+               arrayIsNotString = true;
+           }
+        });
+    }
+
+    if (mavenProfiles && arrayIsNotString) {
         const errorMsg = "BAD_REQUEST: mavenProfiles must be a string array.";
         logger.logProjectError("Failed to change the maven profile list: " + errorMsg, projectID);
 
@@ -645,7 +654,16 @@ const changeMavenProperties = async function (mavenProperties: any, operation: O
         return;
     }
 
-    if (mavenProperties && !(mavenProperties instanceof Array)) {
+    let arrayIsNotString = false;
+    if (mavenProperties instanceof Array) {
+        mavenProperties.forEach(function(item: any): void {
+           if (typeof item !== "string") {
+               arrayIsNotString = true;
+           }
+        });
+    }
+
+    if (mavenProperties && arrayIsNotString) {
         const errorMsg = "BAD_REQUEST: mavenProperties must be a string array.";
         logger.logProjectError("Failed to change the maven properties: " + errorMsg, projectID);
 
--- a/src/pfe/file-watcher/server/test/functional-test/configs/app.config.ts
+++ b/src/pfe/file-watcher/server/test/functional-test/configs/app.config.ts
@@ -28,7 +28,7 @@ export const projectConfigs = {
     "appDirectory": path.join(codewindWorkspaceDir, projectPrefix)
 };
 
-export const projectTypes = ["docker", "nodejs"];
+export const projectTypes = ["spring"];
 export const supportedDockerProjects = ["go", "lagom", "python"];
 
 export const templateNames: any = {
--- a/src/pfe/file-watcher/server/test/functional-test/configs/app.config.ts
+++ b/src/pfe/file-watcher/server/test/functional-test/configs/app.config.ts
@@ -28,14 +28,15 @@ export const projectConfigs = {
     "appDirectory": path.join(codewindWorkspaceDir, projectPrefix)
 };
 
-export const projectTypes = ["spring"];
+export const projectTypes = ["docker", "nodejs", "spring"];
 export const supportedDockerProjects = ["go", "lagom", "python"];
 
 export const templateNames: any = {
     "go": "microclimategotemplate",
     "lagom": "lagomjavatemplate",
     "python": "svtpythontemplate",
-    "nodejs": "node"
+    "nodejs": "node",
+    "spring": "springjavatemplate"
 };
 
 export const DEFAULT_LOG_LEVEL = "trace";
--- a/src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
+++ b/src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
@@ -13,6 +13,7 @@
  */
 import * as project from "../../../src/projects/Project";
 import * as nodeProject from "../../../src/projects/nodejsProject";
+import * as springProject from "../../../src/projects/springProject";
 
 // This config is to manual reset the running build queue and build status options for all project types.
 // We currently need to do that because project that relies on script and IDC code does not follow the typescript code and need to be manually reset.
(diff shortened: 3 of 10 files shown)


ssh24:
@maysunfaisal Could you PTAL?

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
```diff
@@ -550,7 +550,16 @@ const changeMavenProfiles = async function (mavenProfiles: any, operation: Opera
         return;
     }
 
-    if (mavenProfiles && !(mavenProfiles instanceof Array)) {
+    let arrayIsNotString = false;
+    if (mavenProfiles instanceof Array) {
+        mavenProfiles.forEach(function(item: any): void {
+           if (typeof item !== "string") {
+               arrayIsNotString = true;
```
maysunfaisal: can probably break here since we only need one element to be not a string

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
```diff
@@ -645,7 +654,16 @@ const changeMavenProperties = async function (mavenProperties: any, operation: O
         return;
     }
 
-    if (mavenProperties && !(mavenProperties instanceof Array)) {
+    let arrayIsNotString = false;
+    if (mavenProperties instanceof Array) {
+        mavenProperties.forEach(function(item: any): void {
+           if (typeof item !== "string") {
+               arrayIsNotString = true;
```
maysunfaisal: same as above

src/pfe/file-watcher/server/test/functional-test/configs/app.config.ts
```diff
@@ -28,14 +28,15 @@ export const projectConfigs = {
...
+export const projectTypes = ["docker", "nodejs", "spring"];
 export const supportedDockerProjects = ["go", "lagom", "python"];
 
 export const templateNames: any = {
     "go": "microclimategotemplate",
     "lagom": "lagomjavatemplate",
     "python": "svtpythontemplate",
-    "nodejs": "node"
+    "nodejs": "node",
+    "spring": "springjavatemplate"
```
maysunfaisal: Is there a way around to not hardcode these names? Will our tests break if template name changes?

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
```diff
@@ -23,8 +24,12 @@ export const needManualReset: any = {
     "nodejs": {
         "buildStatus": true
     },
+    "spring": {
+        "buildStatus": true
+    },
 };
 
+// project specific capabaility settings: start modes and control commands
```
maysunfaisal: capability typo

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
```diff
@@ -36,11 +41,27 @@ export const projectCapabilities: any = {
...
+        "local": {
+            "startModes": ["run", "debug", "debugNoInit"],
+            "controlCommands": ["restart"]
+        },
+        "kube": project.defaultProjectCapabilities
     }
 };
 
+// auto build enable/disable event capabilities: some projects emit the project changed event and some don't do anything
+export const autoBuildEventCapabailities: any = {
```
maysunfaisal: Should we be fixing this in our source code rather than working around it in the tests? I mean, if some projects dont emit events for the same action, then its a bug, is it not? 😆

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-action.ts
```diff
@@ -37,9 +37,9 @@ export function projectActionTest(socket: SocketIO, projData: ProjectCreation):
             "action": "disableautobuild",
             "returnKeys": ["statusCode", "status"],
             "statusCode": 200,
-            "socketEvent": projData.projectType.toLowerCase() === "docker" || process.env.IN_K8 ? [] : [eventConfigs.events.projectChanged],
-            "eventKeys": projData.projectType.toLowerCase() === "docker" || process.env.IN_K8 ? [] : [["operationId", "projectID", "ignoredPaths", "status", "host", "ports", "containerId", "logs"]],
-            "result": projData.projectType.toLowerCase() === "docker" || process.env.IN_K8 ? [] : [{
+            "socketEvent": process.env.IN_K8  || !project_configs.autoBuildEventCapabailities[projData.projectType] ? [] : [eventConfigs.events.projectChanged],
```
maysunfaisal: For `autoBuildEventCapabailities` like I mentioned above, dont think we should work around it for tests but rather fix it in src code. Probably ask Rajiv's input regd it...

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
```diff
@@ -431,7 +506,9 @@ export function projectSpecificationTest(socket: SocketIO, projData: ProjectCrea
 
             const dockerfile = path.join(projData.location, "Dockerfile");
             const fileOutput = await fs.readFileSync(dockerfile, {encoding: "utf-8"});
-            await fs.writeFileSync(dockerfile, fileOutput.replace(new RegExp(`EXPOSE ${defaultInternalPort}`, "g"), `EXPOSE ${testExposedPort}`), {encoding: "utf-8"});
+            if (projData.projectType != "spring") {
```
maysunfaisal: Cant we use your default one port exposed object in this if block condition rather than hardcoding spring

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
```diff
@@ -463,7 +540,9 @@ export function projectSpecificationTest(socket: SocketIO, projData: ProjectCrea
 
             const dockerfile = path.join(projData.location, "Dockerfile");
             const fileOutput = await fs.readFileSync(dockerfile, {encoding: "utf-8"});
-            await fs.writeFileSync(dockerfile, fileOutput.replace(new RegExp(`EXPOSE ${testExposedPort}`, "g"), `EXPOSE ${defaultInternalPort}`), {encoding: "utf-8"});
+            if (projData.projectType != "spring") {
```
maysunfaisal: same as above

src/pfe/iterative-dev/idc-java/IDC/src/org/eclipse/codewind/iterdev/IDCContext.java
```diff
@@ -164,7 +164,10 @@ public String getAppId() {
 	}
 	
 	public String getLocalWorkspaceOrigin() {
-		if(appDb.get(Constants.DB_WORKSPACE_ORIGIN) != null) {
+		if (System.getenv("NODE_ENV").equals("test")) {
```
maysunfaisal: Do you want to put this change in now or with the Liberty tests? asking because dont want src code to break since we havent started testing liberty projects

src/pfe/file-watcher/server/test/functional-test/configs/app.config.ts
ssh24: it is actually not the real template name. it is the folder under `<project_name>/chart`. i don't know how we can get it dynamically. if you have a better idea let me know and I am more than willing to make it dynamic

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
ssh24: updated it

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
ssh24: updated it

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
ssh24: Not really. This happens on disable and enable auto builds on node projects because of nodemon like Stephanie brought it up in today's meeting. It is not a bug on the source side. On the other having it as capabaility config is nice because in the future if any other project type has a functionaility like how nodemon does for node, it can be applied to it.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-action.ts
ssh24: Read comment https://github.com/eclipse/codewind/pull/387#discussion_r321511499

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
ssh24: I don't think so. This if statement is required for spring projects (and latter on for liberty) because neither of those project has `EXPOSE <PORT>` on their dockerfile. So we don't want to write to the dockerfile on those cases to update the port.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
ssh24: Read comment https://github.com/eclipse/codewind/pull/387#discussion_r321511853

src/pfe/iterative-dev/idc-java/IDC/src/org/eclipse/codewind/iterdev/IDCContext.java
ssh24: I think spring project also uses the same code path? Either ways I think we can use it because I will be working on Liberty next and it will be required for that

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
ssh24: done

src/pfe/file-watcher/server/test/functional-test/configs/app.config.ts
maysunfaisal: Not sure about that. We can discuss it later and think of a solution if needed. For now it's ok .

src/pfe/file-watcher/server/test/functional-test/configs/project.config.ts
maysunfaisal: Thx for the explanation.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
maysunfaisal: Understood

src/pfe/iterative-dev/idc-java/IDC/src/org/eclipse/codewind/iterdev/IDCContext.java
maysunfaisal: Spring does not use idc.java code path.

But since this is an if statement for test and you're going to use Liberty next, ok.

PR #390, opened 2019-09-05
Update to Appsody 0.3.1 release

sghung:
The appsody release was updated as part of the iFix and master needs to pick up the latest one we approved of.


sghung:
Not needed anymore, another PR updating it to 0.4.0 has been made into master

PR #440, opened 2019-09-13
{WIP} Enable remote mode by default.

hhellyer:
Remove the remote mode flag and enable the function by default.
Start codewind using a docker volume for codewind-workspace instead of a bind mount from the host OS.

Marked WIP so we can synchronise with changes to the installer and the editor plugins.


tobespc:
closing as no longer required due to https://github.com/eclipse/codewind/pull/672

PR #467, opened 2019-09-17
Add ability to run both local and kube tests simultaneously

ssh24:
## Description

This is an enhancement on our nightly cron scripts to run both local and kube test simulatenously since they don't overlap and can be kicked off at the same time.

This actually helps particularly instead of having two jobs because for example if the first job takes extremely long time to finish and overlaps with the next job it will cause test issues.


ssh24:
@maysun Could you please help review this?

PR #492, opened 2019-09-19
re enable tests

tobespc:


rajivnathan:
@stalleyj The results show some pending tests. Is that normal?
```
10:33:15    33 passing (50s)
10:33:15    17 pending
```

stalleyj:
@rajivnathan Yes, that's the correct results for now. i've reinstated passing tests but some are currently skipped because they need reworking.

PR #501, opened 2019-09-20
Improve sync in remote mode.

hhellyer:
Only clear files that don't exist in the users local project from pfe's remote copy of the project.
Requires updates to the hybrid branches for codewind-vscode and codewind-eclipse

Signed-off-by: Howard Hellyer <hhellyer@uk.ibm.com>


jcockbain96:
Tested with Howard's `hybrid-sync` vs-code branch and LGTM

Issue #292, opened 2019-08-27
Support binding existing project to appsody


makandre:
**Codewind version:** 0.3.0
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description of the enhancement:**

When binding (importing) a project in codewind, they should be able to choose appsody as a selectable type

**Proposed solution:** 

See https://github.com/eclipse/codewind/issues/292#issuecomment-526356002

makandre:
In order to bind a vanilla project as `appsody` (or similarly `odo`), user would need a secondary selection page to choose the desired `stack` (or `component` for `odo`) 

Today, both VS Code and Eclipse already kind of have that secondary selection page for bind, which is used to select the language:


![image](https://user-images.githubusercontent.com/17596240/63969283-8ac72b80-ca6f-11e9-9806-66c60b56d8a1.png)


I spoke with @eharris369, to populate that dialog, eclipse calls `GET /api/v1/templates` and then loop through to map type to available language.  VS Code probably does something similar (@tetchel please correct if I'm wrong)

# Proposal

There's a `GET /api/v1/project-types` API on portal that returns an array of types (strings); it does not appear to be called by anything currently.   We could change it to return something like this:

```json
[
  {
    "type": "docker",
    "subtypes": [ "go", "java", "python" ]
  },
  {
    "type": "appsodyExtension",
    "subtypes": [ "java-microprofile", "java-spring-boot2", "nodejs-express", ... ]
  },
  ...
]
```

Then IDE call this API instead, and populate the secondary selection with the subtypes. So no more need for IDEs to loop through templates to compute that, portal would figure it out.

For extension types such as `appsody` and the future `odo`, I propose to add a method to the `templatesProvider.js` that portal would call when forming the response for `GET /api/v1/project-types`:

```js
getSubTypes: async function() { ... }
```

Appsody extension would compute the values by calling `appsody list` (and `odo` extension will have their equivalent method).

The above proposal does not cover the actual bind operation yet; bind currently expects a `language` param which we may need to rename and expand to accept more than just language.  Then perhaps codewind CLI (which is going to do the bind) uses that value to invoke the `onProjectBind` command defined by extensions, as proposed in #270

cc: @tobespc @elsony @jingfuwang @ssh24

deboer-tim:
Sounds ok to me. The one thing I'm a little uneasy about is this:
[...]

PR #508, opened 2019-09-20
Change GET /api/v1/project-types

makandre:
For: https://github.com/eclipse/codewind/issues/292

Related PR: https://github.com/eclipse/codewind-appsody-extension/pull/30 

Sample output:

```json

  {
    "projectType": "appsodyExtension",
    "projectSubtypes": {
      "label": "Appsody stack",
      "items": [
        {
          "id": "appsodyhub/java-microprofile",
          "version": "0.2.14",
          "label": "Appsody java-microprofile",
          "description": "Eclipse MicroProfile on Open Liberty & OpenJ9 using Maven"
        }
      ]
    }
  },
  {
    "projectType": "docker",
    "projectSubtypes": {
      "items": [
        {
          "id": "go",
          "label": "go"
        },
        {
          "id": "java",
          "label": "java"
        },
        {
          "id": "python",
          "label": "python"
        }
      ]
    }
  }
]
```

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -1360,9 +1360,7 @@ components:
       type: object
       required:
         - id
-        - version
         - label
-        - description
       properties:
         id:
           type: string
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -1374,7 +1374,6 @@ components:
       description: Descriptor for subtypes of a project type
       type: object
       required:
-        - prompt
         - items
       properties:
         prompt:
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -1021,7 +1021,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: '#/components/schemas/ProjectType'
+                  $ref: '#/components/schemas/ProjectTypeDescriptor'
         204:
           description: No templates found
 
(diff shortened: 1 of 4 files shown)


GeekArthur:
@sghung Please add me as reviewer to ensure this works for the ODO case, thanks

GeekArthur:
Steven can't add me as reviewer. I think probably only committer can be added as the reviewer because reviewer needs approve/merge and only committer has the permission to approve/merge, I can add my comments directly since I already had the PR link.

makandre:
Project Types API tests will fail until appsody extension with https://github.com/eclipse/codewind-appsody-extension/pull/30 is picked up in PFE

sghung:
https://github.com/eclipse/codewind-appsody-extension/pull/30 has been delivered and the 0.5.0 release of appsody extension has been updated. The call to /api/v1/project-types will take around 3-4 seconds because the appsody CLI call for getting the list of projects takes time. There's not much we can do except display some status in the IDE. If the REST API takes more than a few seconds, does it need to go down the async API?

makandre:
I suspect the test is still failing because docker volumes are not purged so an older appsody extension is still being used.  I'm hoping that it will pass once https://github.com/eclipse/codewind/pull/581 is merged and I can rebase my branch with @josephca's fix.

src/pfe/portal/routes/projectTypes.route.js
```diff
@@ -15,15 +15,97 @@ const Logger = require('../modules/utils/Logger');
...
+    if (Array.isArray(types))
+      types.reduce(sanitizeProjectType, projectTypes);
+  }
+
+  return projectTypes;
+}
+
+function addLanguage(projectType, language) {
+  if (!projectType.projectSubtypes.items.find(item => item.id == language))
+    projectType.projectSubtypes.items.push({ id: language, label: language });
```
sghung: I printed out the output of template.label and template.language before this method call and saw output like:
language = java, label = Lagom Java. Should the addLanguage method also have the label passed into it as well? If the label is suppose to be the language, maybe add a comment explaining why, since it is a bit confusing.

makandre: I will add a comment.  The label (`Lagom`) is on the parent level project type, and the properties for the subtypes here is the same on purpose

makandre: Actually I misspoke.  The template's label is actually not used here in this API, it is used when we present the list of templates to user for project creation workflow.  In any case I'll add a comment

docs/openapi.yml
```diff
@@ -1344,6 +1344,45 @@ components:
...
+        description:
+          type: string
+    ProjectSubtypesDescriptor:
+      description: Descriptor for subtypes of a project type
+      type: object
+      required:
+        - items
+      properties:
+        prompt:
+          description: A prompt for user to select a subtype
```
elsony: We are not sure how else will the name of the subtype will need to be exposed on the UI in the future.  Instead of having a dedicated question for subtype selection prompt, should this be only the label of the subtype, e.g. Appsody stack, instead of a full question?  The problem with the full question prompt is that we need to come up with different strings in case we need to prompt users on subtype related things or error messages, we need to come up with unique strings for each one. We may be better off just provide the label of the subtype and let the IDE decide the full prompt by doing string substitution with the subtype lable only.

makandre: Sure, can do that.  And should we simply rename this to `label`?

makandre:
@elsony re the version, I forgot to edit the sample output.  It no longer includes the version in the label.

docs/openapi.yml
sghung: @elsony the change from prompt to label has been done. We have also tested this via the appsody extension via the merge of https://github.com/eclipse/codewind-appsody-extension/pull/32. Is there anything else required for a review approval? Thank you

GeekArthur:
LGTM, this can be applied to ODO extension bind case as well. Thanks!

PR #532, opened 2019-09-24
Allow developers to set appsody extension version

sghung:
Sometimes, we want to modify the appsody extension and not have it overwritten on a ./run.sh call.

If APPSODY_EXT_VERSION is unset, then the default will be downloaded
If APPSODY_EXT_VERSION is set, it will attempt to download that version. No checks are in place for invalid URLs, since this is for developers only. The name of the zip is displayed in the console in case they want to diagnose

An env variable is used, as that is what the IDE uses to specify which version of codewind to use. They use CW_TAGS, so we follow a similar approach.

### DIFF
--- a/script/build.sh
+++ b/script/build.sh
@@ -26,14 +26,6 @@ PERFORMANCE=performance;
 ARCH=`uname -m`;
 TAG=latest;
 REGISTRY=eclipse
-DEVMODE=false
-
-while [ "$#" -gt 0 ]; do
-  case $1 in
-    --dev) DEVMODE=true; shift 1;;
-    *) shift 1;;
-  esac
-done
 
 # On intel, uname -m returns "x86_64", but the convention for our docker images is "amd64"
 if [ "$ARCH" == "x86_64" ]; then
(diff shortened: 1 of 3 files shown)


script/build.sh
```diff
@@ -55,10 +63,28 @@ cp -r $DIR/docs ${SRC_DIR}/pfe/portal/
...
+if [ "$DEVMODE" == "true" ]; then
+  if [ "$APPSODY_EXT_VERSION" == "none" ]; then 
+    echo -e "\n+++   SKIPPING APPSODY EXTENSION DOWNLOAD DUE TO ENV VARIABLE APPSODY_EXT_VERSION=none +++\n";
+  else
+    if [ -z "$APPSODY_EXT_VERSION" ]; then 
+      APPSODY_EXT_VERSION=$DEFAULT_APPSODY_EXT_VERSION
+    fi
+
+    echo -e "\n+++   DOWNLOADING APPSODY EXTENSION: ${APPSODY_EXT_VERSION}  +++\n";
+    curl -Lo ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-${APPSODY_EXT_VERSION}.zip https://github.com/eclipse/codewind-appsody-extension/archive/${APPSODY_EXT_VERSION}.zip
```
makandre: Could we avoid duplicating these lines by checking if appsody version is set?  (then `--dev APPSODY_EXT_VERSION=none` can unset the version to skip the lines)

makandre:
Just want to state that this method works only for developers that build their own PFE.  This won't work for someone working just on an extension with a pre-built PFE; but this is probably ok, because now they can also prevent overwrite if they have a higher version of the extension than the one inside the PFE.

sghung:
@makandre thanks for the review. A few things:
- I have removed the need for `--dev` because as I was writing the readme, I was searching for other places `--dev` was used and it affects which docker compose file is used. It will use the dev version of the docker compose file, which currently has different configuration than the default docker compose file used when running without the flag. Since there may be differences, I decided to allow the environment variable to be used whenever. It is still safe as the Jenkins build goes down a different path, so it doesn't affect the build
- I have refactored the code to avoid duplicating the curl call. I also have a refactored version that takes `--dev` into account in case we still want to keep it as a `--dev` only feature
- Updated the README to include instructions on setting the version
- Correct, this is just for developers, not end users

sghung:
@tobespc can you help review? Thank you

elsony:
We are currently hardcoding Appsody specific information into PFE.  Ideally, the Appsody support is just one of the extensions so the Appsody specific information should go to the extension itself instead of burning into PFE.  Have we consider to abstract it out and put it somewhere as part of the extension and owned by the extension, e.g in the extension descriptor file?

sghung:
Had a talk with Elson and an improved way to do this (without hard coding the version) is outside the scope of this fix.

josephca:
Closing it without merge after discussion with @makandre

Issue #448, opened 2019-09-13
odo project type defintion


elsony:
Define new project type extension for odo implementation. This extension is only applicable on OpenShift Kube scenario, not applicable for local.

GeekArthur:
You can assign this issue to me.

GeekArthur:
@rajivnathan This one can be closed.

PR #601, opened 2019-10-01
ODO extension

GeekArthur:
Issue: https://github.com/eclipse/codewind/issues/448

This PR is used for implementing odo extension, it include features:
- Initialize odo components.
- Basic users scenarios: create, update, delete odo components.
- Generate build status and app status for odo components.
- Generate build log and app log for odo components.
- Dynamically get container name, app name and app port.

### DIFF
--- a/src/pfe/file-watcher/server/src/extensions/IExtensionProject.ts
+++ b/src/pfe/file-watcher/server/src/extensions/IExtensionProject.ts
@@ -45,7 +45,7 @@ export interface IExtensionProject {
      *
      * @param projectInfo <Required | ProjectInfo> - The metadata information for the project.
      */
-    start(projectInfo: ProjectInfo): Promise<void>;
+    start?(projectInfo: ProjectInfo): Promise<void>;
 
     /**
      * @function
@@ -53,7 +53,7 @@ export interface IExtensionProject {
      *
      * @param projectInfo <Required | ProjectInfo> - The metadata information for the project.
      */
-    stop(projectInfo: ProjectInfo): Promise<void>;
+    stop?(projectInfo: ProjectInfo): Promise<void>;
 
     /**
      * @function
@@ -130,7 +130,27 @@ export interface IExtensionProject {
      * @param projectID <Required | string> - An alphanumeric identifier for a project.
      * @param projectLocation <Required | string> - The project location directory.
      *
-     * @returns string
+     * @returns Promise<string>
      */
-    getContainerName?(projectID: string, projectLocation: string): string;
+    getContainerName?(projectID: string, projectLocation: string): Promise<string>;
+
+    /**
+     * @function
+     * @description Return a reproducible app name for the project.
+     *
+     * @param projectID <Required | string> - An alphanumeric identifier for a project.
+     *
+     * @returns Promise<string>
+     */
+    getAppName?(projectID: string): Promise<string>;
+
+    /**
+     * @function
+     * @description Get the internal app port that the project uses.
+     *
+     * @param projectID <Required | string> - An alphanumeric identifier for a project.
+     *
+     * @returns Promise<string>
+     */
+    getAppPort?(projectID: string): Promise<string>;
 }
--- a/src/pfe/file-watcher/server/src/extensions/IExtensionProject.ts
+++ b/src/pfe/file-watcher/server/src/extensions/IExtensionProject.ts
@@ -153,4 +153,14 @@ export interface IExtensionProject {
      * @returns Promise<string>
      */
     getAppPort?(projectID: string): Promise<string>;
+
+    /**
+     * @function
+     * @description Get the app ingress URL as app base URL.
+     *
+     * @param projectID <Required | string> - An alphanumeric identifier for a project.
+     *
+     * @returns Promise<string>
+     */
+    getAppBaseURL?(projectID: string): Promise<string>;
 }
--- a/src/pfe/file-watcher/server/src/extensions/projectExtensions.ts
+++ b/src/pfe/file-watcher/server/src/extensions/projectExtensions.ts
@@ -22,6 +22,7 @@ import * as swiftProject from "../projects/swiftProject";
 import * as nodeProject from "../projects/nodejsProject";
 import { DockerProject } from "../projects/DockerProject";
 import { ShellExtensionProject } from "../projects/ShellExtensionProject";
+import { OdoExtensionProject } from "../projects/OdoExtensionProject";
 import { ProjectInfo, ProjectCapabilities, defaultProjectCapabilities } from "../projects/Project";
 import { workspaceConstants } from "../projects/constants";
 
@@ -163,12 +164,14 @@ export function getAllProjectTypes(): Array<string> {
 async function getExtensionProjectHandler(projectInfo: ProjectInfo): Promise<any> {
 
     const key = projectInfo.projectID;
+    const projectName = path.basename(projectInfo.location);
     let handler = extensionProjectHandlers[key];
 
     // is there an extension handler for the project?
     if (!handler) {
 
         const fullPath = projectInfo.extensionID;
+        logger.logProjectInfo("Extension absolute path is: " + fullPath, key, projectName);
 
         // try to load extension handler if ID was provided
         if (fullPath) {
@@ -177,14 +180,17 @@ async function getExtensionProjectHandler(projectInfo: ProjectInfo): Promise<any
                 if (files) {
                     if (files.includes(".sh-extension")) {
                         handler = new ShellExtensionProject(projectInfo.projectType);
-                        await handler.init(projectInfo);
-                        extensionProjectHandlers[key] = handler;
+                    } else if (files.includes(".odo-extension-config.json")) {
+                        handler = new OdoExtensionProject(projectInfo.projectType);
                     }
+
+                    await handler.init(projectInfo);
+                    extensionProjectHandlers[key] = handler;
                 }
             }
             catch (err) {
-                logger.logError(`Failed to get extension project handler ${fullPath} for ${projectInfo.location}`);
-                logger.logError(err);
+                logger.logProjectError(`Failed to get extension project handler ${fullPath} for ${projectInfo.location}`, key, projectName);
+                logger.logProjectError(err, key, projectName);
                 return undefined;
             }
         }
--- a/src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
+++ b/src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
@@ -272,4 +272,42 @@ export class OdoExtensionProject implements IExtensionProject {
 
         return appPort;
     }
+
+    /**
+     * @function
+     * @description Get the app ingress URL as app base URL.
+     *
+     * @param projectID <Required | string> - An alphanumeric identifier for a project.
+     *
+     * @returns Promise<string>
+     */
+    getAppBaseURL = async (projectID: string): Promise<string> => {
+        let appBaseURL: string = undefined;
+        const projectInfo: ProjectInfo = await projectUtil.getProjectInfo(projectID);
+        const projectLocation = projectInfo.location;
+        const projectName = path.basename(projectLocation);
+        const args: string[] = [
+            projectLocation,
+            "",
+            "getURL",
+            "",
+            "",
+            "",
+            ""
+        ];
+
+        try {
+            const result = await processManager.spawnDetachedAsync(projectID, path.join(this.fullPath, "odo-extension-entrypoint.sh"), args, {});
+
+            if (result.exitCode != 0) {
+                logger.logProjectError(result.stderr, projectID, projectName);
+            } else {
+                appBaseURL = result.stdout;
+            }
+        } catch (err) {
+            logger.logProjectError(err, projectID, projectName);
+        }
+
+        return appBaseURL;
+    }
 }
--- a/src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
+++ b/src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
@@ -1,0 +1,275 @@
+/*******************************************************************************
+ * Copyright (c) 2019 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+"use strict";
+
+import fs from "fs-extra";
+import * as path from "path";
+
+import * as projectUtil from "./projectUtil";
+import * as processManager from "../utils/processManager";
+import * as logHelper from "./logHelper";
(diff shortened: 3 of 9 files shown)


GeekArthur:
#448

src/pfe/file-watcher/server/src/extensions/IExtensionProject.ts
```diff
@@ -130,7 +130,27 @@ export interface IExtensionProject {
...
+    /**
+     * @function
+     * @description Get the internal app port that the project uses.
+     *
+     * @param projectID <Required | string> - An alphanumeric identifier for a project.
+     *
+     * @returns Promise<string>
      */
-    getContainerName?(projectID: string, projectLocation: string): string;
+    getAppPort?(projectID: string): Promise<string>;
```
elsony: We are adding these functions. Does it mean that the Appsody needs to react to implement these as well to avoid breakage?

src/pfe/file-watcher/server/src/extensions/projectExtensions.ts
```diff
@@ -177,14 +180,17 @@ async function getExtensionProjectHandler(projectInfo: ProjectInfo): Promise<any
                 if (files) {
                     if (files.includes(".sh-extension")) {
                         handler = new ShellExtensionProject(projectInfo.projectType);
-                        await handler.init(projectInfo);
-                        extensionProjectHandlers[key] = handler;
+                    } else if (files.includes(".odo-extension-config.json")) {
```
elsony: Why are we removing the handler.init from the appsody case (I assume the above case is for Appsody)? Also, do we need to burn that info as part of the PFE code? Should this be defined as part of the extension manifest and load  that info from it?

src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    /**
+     * @function
+     * @description Delete operation for the project.
+     *
+     * @param projectInfo <Required | ProjectInfo> - The metadata information for the project.
+     */
+    deleteContainer = async (projectInfo: ProjectInfo): Promise<void> => {
+        await projectUtil.containerDelete(projectInfo, path.join(this.fullPath, "odo-extension-entrypoint.sh"));
```
elsony: My guess is this code is trying to delet the entrypoint.sh file during the container deletion operation. Why do we delete the entrypoint.sh during a containerDelete?

src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
```diff
@@ -0,0 +1,275 @@
...
+    /**
+     * @function
+     * @description Get the project capabilities of the project.
+     *
+     * @returns ProjectCapabilities
+     */
+    getCapabilities = (): ProjectCapabilities => {
+
+        if (this.config.capabilities) {
+            return new ProjectCapabilities(this.config.capabilities.startModes, this.config.capabilities.controlCommands);
```
elsony: I think the default capability supports startModes of debug and debugNoInit. Given that odo does not support debug, should we filter those out?

src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
```diff
@@ -0,0 +1,275 @@
...
+    /**
+     * @function
+     * @description Return a reproducible container name for the project.
+     *
+     * @param projectID <Required | string> - An alphanumeric identifier for a project.
+     * @param projectLocation <Required | string> - The project location directory.
+     *
+     * @returns Promise<string>
+     */
+    getContainerName = async (projectID: string, projectLocation: string): Promise<string> => {
```
elsony: If I remember correctly, we discussed before that we want to have the container name starts with "cw-". Please add.

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -186,7 +186,19 @@ export async function containerCreate(operation: Operation, script: string, comm
             args = [projectLocation, LOCAL_WORKSPACE, operation.projectInfo.projectID, command, operation.containerName,
                 String(operation.projectInfo.autoBuildEnabled), logName, operation.projectInfo.startMode, operation.projectInfo.debugPort,
                 (operation.projectInfo.forceAction) ? String(operation.projectInfo.forceAction) : "NONE", logDir, deploymentRegistry, userMavenSettings];
-    }
+        } else if (projectType == "odo") {
```
elsony: Why do we need to do this?  Does Appsody has equivalent?  We should try to avoid burning extension specific code into PFE.  Can this be move to part of the extension code?

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -287,6 +299,18 @@ export async function containerUpdate(operation: Operation, script: string, comm
         args = [projectLocation, LOCAL_WORKSPACE, operation.projectInfo.projectID, command, operation.containerName,
             String(operation.projectInfo.autoBuildEnabled), logName, operation.projectInfo.startMode, operation.projectInfo.debugPort,
             (operation.projectInfo.forceAction) ? String(operation.projectInfo.forceAction) : "NONE", logDir, deploymentRegistry, userMavenSettings];
+    } else if (projectType == "odo") {
```
elsony: Why do we need to do this? Does Appsody has equivalent? We should try to avoid burning extension specific code into PFE. Can this be move to part of the extension code?

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -618,10 +642,27 @@ export async function containerDelete(projectInfo: ProjectInfo, script: string):
     processManager.killRunningProcesses(projectInfo.projectID, projectName);
 
     try {
+        let args: any[] = [projectInfo.location, LOCAL_WORKSPACE, projectID, "remove", containerName, undefined, undefined, undefined, undefined, undefined, undefined, deploymentRegistry];
+
+        if (projectInfo.projectType == "odo") {
```
elsony: Why do we need to do this? Does Appsody has equivalent? We should try to avoid burning extension specific code into PFE. Can this be move to part of the extension code?

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -1172,9 +1225,24 @@ export async function runScript(projectInfo: ProjectInfo, script: string, comman
     const containerName = await getContainerName(projectInfo);
     const logName = getLogName(projectInfo.projectID, projectInfo.location);
     const logDir = await logHelper.getLogDir(projectInfo.projectID, projectInfo.projectName);
-    const args = [projectInfo.location, LOCAL_WORKSPACE, projectID, command, containerName, String(projectInfo.autoBuildEnabled), logName, projectInfo.startMode,
+    const projectName = path.basename(projectInfo.location);
+    let args = [projectInfo.location, LOCAL_WORKSPACE, projectID, command, containerName, String(projectInfo.autoBuildEnabled), logName, projectInfo.startMode,
         projectInfo.debugPort, "NONE", logDir];
 
+    if (projectInfo.projectType == "odo") {
```
elsony: Why do we need to do this? Does Appsody has equivalent? We should try to avoid burning extension specific code into PFE. Can this be move to part of the extension code?

src/pfe/file-watcher/server/src/utils/kubeutil.ts
```diff
@@ -115,14 +122,25 @@ export async function getApplicationContainerInfo(projectInfo: ProjectInfo, oper
         const internalPorts: Array<string> = [];
         const exposedPorts: Array<string> = [];
         const appPorts = projectInfo.appPorts;
-        const resp = await k8sClient.api.v1.namespaces(KUBE_NAMESPACE).services.get({ qs: { labelSelector: releaseLabel } });
+        let resp: any = undefined;
+        let deploymentconfig: string = undefined;
+
+        if (projectInfo.projectType == "odo") {
```
elsony: Why do we need to do this? Does Appsody has equivalent? We should try to avoid burning extension specific code into PFE. Can call into extension code for that?

src/pfe/file-watcher/server/src/utils/kubeutil.ts
```diff
@@ -115,14 +122,25 @@ export async function getApplicationContainerInfo(projectInfo: ProjectInfo, oper
...
+            resp = await k8sClient.api.v1.namespaces(KUBE_NAMESPACE).services.get({ qs: { labelSelector: releaseLabel } });
+        }
 
         // iterate through the available ports and store both the internal and exposed ports
         for ( let i = 0 ; i < resp.body.items.length ; i++ ) {
-            info.serviceName = resp.body.items[i].metadata.name;
-            for (let j = 0; j < resp.body.items[i].spec.ports.length; j++) {
-                internalPorts.push(String(resp.body.items[i].spec.ports[j].targetPort));
-                exposedPorts.push(String(resp.body.items[i].spec.ports[j].nodePort));
+            if (projectInfo.projectType != "odo" || (projectInfo.projectType == "odo" && resp.body.items[i].spec.selector.deploymentconfig == deploymentconfig)) {
```
elsony: Same here. We should delegate that to the extension code instead of hardcoding on this layer of the code.

src/pfe/file-watcher/server/src/utils/kubeutil.ts
```diff
@@ -185,9 +203,15 @@ export async function getApplicationContainerInfo(projectInfo: ProjectInfo, oper
  *
  * @returns Promise<any>
  */
-export async function isContainerActive(containerName: string): Promise<any> {
+export async function isContainerActive(containerName: string, projectInfo?: ProjectInfo): Promise<any> {
     try {
-        const releaseLabel = "release=" + containerName;
+        let releaseLabel = "release=" + containerName;
+        if (projectInfo.projectType == "odo") {
```
elsony: Same here. We should delegate that to the extension code instead of hardcoding on this layer of the code.

src/pfe/file-watcher/server/src/utils/locales/en/translation.json
```diff
@@ -26,7 +26,13 @@
       "missingDeploymentRegistry": "The project will not build due to missing Deployment Registry. Run the Set Deployment Registry command to set a new Deployment Registry to build projects.",
       "invalidDeploymentRegistry": "Codewind was unable to push an image to the Deployment Registry. Please make sure it is a valid Deployment Registry with the appropriate permissions.",
       "wrongDeploymentRegistry": "The project will not build with the new Deployment Registry. Please remove and re-add the project to deploy to the new Deployment Registry.",
-      "unexpectedError": "An unexpected error occurred"
+      "unexpectedError": "An unexpected error occurred",
+      "odoBuildCreateInProgressMsg": "Creating odo component",
+      "odoBuildCreateFailMsg": "Failed to create odo component",
+      "odoBuildPushInProgressMsg": "Building and deploying odo component",
```
elsony: We don't build and deploy the odo component. Do you mean application?

src/pfe/file-watcher/server/src/utils/locales/en/translation.json
```diff
@@ -26,7 +26,13 @@
...
+      "odoBuildPushFailMsg": "Failed to build or deploy odo component",
```
elsony: We don't build and deploy the odo component. Do you mean application?

src/pfe/file-watcher/server/src/utils/locales/en/translation.json
```diff
@@ -26,7 +26,13 @@
...
+      "odoBuildUrlInProgressMsg": "Creating URL for odo component",
```
elsony: We create the URL for the application, not the odo component.

src/pfe/file-watcher/server/src/utils/locales/en/translation.json
```diff
@@ -26,7 +26,13 @@
...
+      "odoBuildUrlFailMsg": "Fail to create URL for odo component"
```
elsony: We create the URL for the application, not the odo component.

src/pfe/file-watcher/server/src/extensions/IExtensionProject.ts
GeekArthur: Nope, we put `?` mark for the function, that means the function in the interface is optional.

src/pfe/file-watcher/server/src/extensions/projectExtensions.ts
GeekArthur: Actually we are not removing the handler.init, we just move the handler.init to the bottom since both Appsody and ODO can share the same code block for extension project init.

The extension manifest file is more about extension metadata (eg. extension name, extension project type, extension detection, etc), but this config file is more about the extension implementation (eg. build log name, app log name, capabilities) that uses for Turbine layer. ODO inherits the same design from Appsody.

src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
GeekArthur: We don't delete the entrypoint.sh, we just call the entrypoint.sh to do the container deletion operation, it's same as create and update cases, but for delete case we call entrypoint.sh with `remove` parameter here: https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/server/src/projects/projectUtil.ts#L624

src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
GeekArthur: We already filtered those out in ODO extension implementation here: https://github.com/eclipse/codewind-odo-extension/blob/master/.odo-extension-config.json#L9, then we load the capability in Turbine layer.

src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
GeekArthur: Yes, it's already added, actually the ODO "getContainerName" is different from other project types "getContainerName", other project types can specify container name through "getContainerName", but ODO can't due to ODO will generate container name automatically, we can only specify the component name, so I add the `cw-` here: https://github.com/eclipse/codewind/pull/601/files#diff-7597df0b79304bd2be18940009414316R800

src/pfe/file-watcher/server/src/utils/locales/en/translation.json
GeekArthur: The component (not component type) keyword that I use is from ODO man page: https://github.com/openshift/odo/blob/master/docs/cli-reference.adoc#create. But I can update it to application which makes more sense for our product.

src/pfe/file-watcher/server/src/utils/locales/en/translation.json
GeekArthur: The component (not component type) keyword that I use is from ODO man page: https://github.com/openshift/odo/blob/master/docs/cli-reference.adoc#create. But I can update it to application which makes more sense for our product.

src/pfe/file-watcher/server/src/extensions/projectExtensions.ts
elsony: As discussed, we'll need a second PR to clean up all odo extension code and do it in generic extension way for the project type specific code via the IExtensionProject interface.

Issue #586, opened 2019-09-30
Add 'name' attribute to POST /template/repositories


tetchel:
See https://github.com/eclipse/codewind/issues/541#issuecomment-535591155 and https://github.com/eclipse/codewind/issues/541#issuecomment-535658027

We want the user to able to set a `name` attribute on their template sources, outside of the `templates.json`, just like they can currently with `description`.

Also, both `name` and `description` should be optional, since they can be fetched from `templates.json` instead.

jcockbain96:
Taking this, could I be assigned?

jcockbain96:
Is this to follow that second comment? So we read the name from the template.json, and then allow a user to edit the name. 

Or do we just want to allow a 'name' parameter in the body of the API which will then be overwritten if a name does exist in the template.json?

tetchel:
I suppose if a name or description is set in the `repository_list`, then it should override the one in `templates.json` for that user.

PR #602, opened 2019-10-02
Adding 'name' parameter to POST /templates/repo

jcockbain96:
Adds an extra parameter in the request body of `POST /templates/repo` for repo name (`name`), this will take precedence over the name read from the `templates.json` in the repo. 

Also makes `name` and `description` both optional parameters for this API, with the user description taking precedence too.

Updates templates.service to match new template descriptions, which had been changed without updating the tests.

Resolves https://github.com/eclipse/codewind/issues/586.

Latest Commit - adds source to each template endpoint (if a repo name exists) - #608, and updates the default repo list to have a name for the Kabenero repo - #609.

Signed-off-by: James Cockbain <james.cockbain@ibm.com>

### DIFF
--- a/src/pfe/portal/modules/Templates.js
+++ b/src/pfe/portal/modules/Templates.js
@@ -289,7 +289,11 @@ function fetchAllRepositoryDetails(repos) {
 
 async function fetchRepositoryDetails(repo) {
   let newRepo = {...repo}
-  await readRepoTemplatesJSON(newRepo);
+
+  // Only set the name or description of the repo if not given by the user
+  if (!(repo.name && repo.description)){
+    await readRepoTemplatesJSON(newRepo);
+  }
 
   if (repo.projectStyles) {
     return newRepo;
@@ -326,8 +330,7 @@ async function readRepoTemplatesJSON(repository) {
   try {
     const templateDetails = JSON.parse(res.body);
     for (const prop of ['name', 'description']) {
-      // Only set the name or description of the repo if not given by the user
-      if (templateDetails.hasOwnProperty(prop) && !repository[prop]) {
+      if (templateDetails.hasOwnProperty(prop)) {
         repository[prop] = templateDetails[prop];
       }
     }
--- a/src/pfe/portal/modules/Templates.js
+++ b/src/pfe/portal/modules/Templates.js
@@ -326,6 +326,7 @@ async function readRepoTemplatesJSON(repository) {
   try {
     const templateDetails = JSON.parse(res.body);
     for (const prop of ['name', 'description']) {
+      // Only add the name or description if not set by the user on adding
       if (templateDetails.hasOwnProperty(prop) && !repository[prop]) {
         repository[prop] = templateDetails[prop];
       }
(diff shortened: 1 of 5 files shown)


src/pfe/portal/modules/Templates.js
```diff
@@ -325,7 +326,8 @@ async function readRepoTemplatesJSON(repository) {
   try {
     const templateDetails = JSON.parse(res.body);
     for (const prop of ['name', 'description']) {
-      if (templateDetails.hasOwnProperty(prop)) {
+      // Only set the name or description of the repo if not given by the user
+      if (templateDetails.hasOwnProperty(prop) && !repository[prop]) {
```
rwalle61: this should work, though performance could be improved if the `!repository[prop]` check were done before GETting the templates.json, to skip that request when name and desc are already set?

test/src/unit/template.test.js
```diff
@@ -399,10 +400,11 @@ describe('Templates.js', function() {
             it('succeeds', async function() {
```
rwalle61: the `describe`s now should have `<validName>` or something to similar effect

test/src/unit/template.test.js
```diff
@@ -412,6 +412,20 @@ describe('Templates.js', function() {
                 });
             });
         });
+        describe('(<validUrl>, <NoDesc> <NoName>)', function() {
+            it('succeeds, and gets the name and description from templates.json', async function() {
+                const url = 'https://raw.githubusercontent.com/kabanero-io/codewind-templates/master/devfiles/index.json';
+                const func = () => templateController.addRepository(url);
+                await (func().should.not.be.rejected);
+                templateController.repositoryList.should.deep.include({
```
rwalle61: can this instead use an object from `templates.service.js`? (To reduce maintenance)

src/pfe/portal/modules/Templates.js
jcockbain96: What if only one is set?

rwalle61: in that case we still need to get the one that isn't set by the user, right?

jcockbain96: Sorry, yeah understood what you meant so deleted the comment

rajivnathan:
@jcockbain96  The build for this PR seems to be green. Will this change fix the error I'm seeing in my PR build?

```
15:08:17    1) Batch enabling repositories
15:08:17         1 repo
15:08:17           returns 207 and sub-status 200 for each subrequest when batch disabling 1 repos:
15:08:17       Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/genie.codewind/jenkins-agent/workspace/Codewind_codewind_PR-604/test/src/API/templates/enableRepo.test.js)
```

jcockbain96:
@rajivnathan I get that one passing on master so I don't think so - were you referring to https://github.com/eclipse/codewind/pull/604? Is green now so may just be it's sometimes hitting that timeout? We can up it if so.

Issue #613, opened 2019-10-03
Turbine tests broken


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
`npm run unit:test` fails with 

```
/Users/maysun/dev/eclipse/codewind/src/pfe/file-watcher/server

/Users/maysun/dev/eclipse/codewind/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts:70
     *
                                                                                         ^
TypeError: projectUtil.getAllProjectInfo is not a function
    at initAppStateMap (/Users/maysun/dev/eclipse/codewind/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts:70:90)
    at Object.<anonymous> (/Users/maysun/dev/eclipse/codewind/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts:33:2)
    at Module._compile (internal/modules/cjs/loader.js:778:30)

```

**Steps to reproduce:**
Switching to commit before the ODO PR (`added a test for getting name and description from templates.json`)
```
git checkout f8bde84cb163a5791ca40ca70d037ece6ea8477b
```

-> Ran tests, it was fine.

Adding ODO PR first commit (`odo extension`)
```
git cherry-pick 82be6aaf7725bdf50382051cca695bec44fc2677
```

-> Ran tests, tests broken

you can see the commit hash from here https://github.com/eclipse/codewind/commits/master

maysunfaisal:
I didnt see it in the PFE log, so dont think this is an actual issue with the product

but only on nyc mocha execs:

```
maysuns-mbp:server maysun$ nyc mocha test/functional-test/functional.test.ts

/Users/maysun/dev/eclipse/codewind/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts:70
     *
                                                                                         ^
TypeError: projectUtil.getAllProjectInfo is not a function
    at initAppStateMap (/Users/maysun/dev/eclipse/codewind/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts:70:90)
    at Object.<anonymous> (/Users/maysun/dev/eclipse/codewind/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts:33:2)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Module.replacementCompile (/usr/local/lib/
```
```
[...]

PR #616, opened 2019-10-03
Fix Turbine Test - Circular Dependency on Modules

ssh24:
### Description

There was circular dependency introduced from https://github.com/eclipse/codewind/pull/601. This PR extracts the logic to set app name into the project info metadata and use that where needed and avoid circular dependencies.

Related to issue #613

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/projectUtil.ts
+++ b/src/pfe/file-watcher/server/src/projects/projectUtil.ts
@@ -1908,7 +1908,7 @@ async function getPODInfoAndSendToPortal(operation: Operation, event: string = "
         status: "failed"
     };
 
-    const projectInfo = operation.projectInfo;
+    let projectInfo = operation.projectInfo;
     const projectLocation = projectInfo.location;
     const projectID = projectInfo.projectID;
     const projectName = projectInfo.projectName;
@@ -1934,6 +1934,12 @@ async function getPODInfoAndSendToPortal(operation: Operation, event: string = "
         logger.logProjectInfo(`The container was started successfully for application ` + projectLocation, projectID, projectName);
 
         logger.logProjectInfo("The project location for " + projectID + " is " + projectLocation, projectID, projectName);
+
+        if (projectInfo.projectType == "odo") {
+            // this if loop is  used to get the odo app name to set the odo app name in the project info file and reload it back
+            await setOdoAppName(projectInfo);
+            projectInfo = await getProjectInfo(projectInfo.projectID);
+        }
 
         const containerInfo = await kubeutil.getApplicationContainerInfo(projectInfo, operation);
         projectEvent.status = "success";
--- a/src/pfe/file-watcher/server/src/projects/projectUtil.ts
+++ b/src/pfe/file-watcher/server/src/projects/projectUtil.ts
@@ -401,10 +401,12 @@ async function executeBuildScript(operation: Operation, script: string, args: Ar
 
                     logger.logProjectInfo("The project location for " + operation.projectInfo.projectID + " is " + projectLocation, projectID, projectName);
 
+                    // this if loop is used to get the odo app name and set the odo app name in the project info file and reload it back
                     if (operation.projectInfo.projectType == "odo") {
-                        // this if loop is  used to get the odo app name to set the odo app name in the project info file and reload it back
-                        await setOdoAppName(operation.projectInfo);
-                        operation.projectInfo = await getProjectInfo(operation.projectInfo.projectID);
+                        if (!operation.projectInfo.odoAppName) {
+                            await getAndSaveOdoAppName(operation.projectInfo);
+                            operation.projectInfo = await getProjectInfo(operation.projectInfo.projectID);
+                        }
                     }
 
                     const containerInfo = await kubeutil.getApplicationContainerInfo(operation.projectInfo, operation);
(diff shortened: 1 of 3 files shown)


ssh24:
@maysunfaisal Could you PTAL?

ssh24:
@jingfuwang Could you also test it out to see if the ODO projects work properly?

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -400,6 +400,13 @@ async function executeBuildScript(operation: Operation, script: string, args: Ar
                     projectInfo.status = "success";
 
                     logger.logProjectInfo("The project location for " + operation.projectInfo.projectID + " is " + projectLocation, projectID, projectName);
+
+                    if (operation.projectInfo.projectType == "odo") {
```
maysunfaisal: In all these if conditions, can you check if projectInfo.odoAppName is undefined, because we dont want to always set. We only want to do that, if odoAppName is not present

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -943,6 +957,11 @@ export async function isContainerActive(projectID: string, handler: any): Promis
         const containerName = await getContainerName(projectInfo);
         let containerState = undefined;
         if (process.env.IN_K8 === "true") {
+            if (projectInfo.projectType == "odo") {
```
maysunfaisal: same as abbove

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -830,6 +837,13 @@ export async function getContainerInfo(projectInfo: ProjectInfo, forceRefresh: b
     if (process.env.IN_K8 === "true") {
         const operation = new Operation("general", projectInfo);
         operation.containerName = containerName;
+
+        if (projectInfo.projectType == "odo") {
```
maysunfaisal: same as above

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -1897,6 +1935,12 @@ async function getPODInfoAndSendToPortal(operation: Operation, event: string = "
 
         logger.logProjectInfo("The project location for " + projectID + " is " + projectLocation, projectID, projectName);
 
+        if (projectInfo.projectType == "odo") {
```
maysunfaisal: same as above

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -967,6 +986,25 @@ export async function isContainerActive(projectID: string, handler: any): Promis
...
+ *
+ * @param projectInfo <Required | ProjectInfo> - The metadata information for a project.
+ *
+ * @returns Promsie<void>
+ */
+export async function setOdoAppName(projectInfo: ProjectInfo): Promise<void> {
+    const projectID = projectInfo.projectID;
+
+    const projectHandler = await projectExtensions.getProjectHandler(projectInfo);
+    const appName = await projectHandler.getAppName(projectID);
```
maysunfaisal: before calling getAppName can we also check if the func is present; because setOdoAppName can be called by any project.. so just want to safeguard...

src/pfe/file-watcher/server/src/projects/projectUtil.ts
ssh24: Why do we not want to set it even if it is present? It will just override the old one which is the same?

maysunfaisal: because once we have the name, is there a likelihood it will change? Otherwise we're doing an odo command execution everytime. The project handler getAppName calls an odo command to get the app name..

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -967,6 +986,25 @@ export async function isContainerActive(projectID: string, handler: any): Promis
     }
 }
 
+/**
+ * @function
+ * @description Set the ODO app name and save it to the project info file.
```
maysunfaisal: It should be get the ODO app name and save it...

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -967,6 +986,25 @@ export async function isContainerActive(projectID: string, handler: any): Promis
...
+ *
+ * @param projectInfo <Required | ProjectInfo> - The metadata information for a project.
+ *
+ * @returns Promsie<void>
+ */
+export async function setOdoAppName(projectInfo: ProjectInfo): Promise<void> {
```
maysunfaisal: Not sure, if I want to know call this getOdoAppName now; because the we are actually getting the name using the odo command.. Maybe call it getAndSaveOdoAppName?

src/pfe/file-watcher/server/src/projects/projectUtil.ts
ssh24: hmm, that makes sense. i will add that

ssh24:
@maysunfaisal Could you PTAL?

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -401,10 +401,12 @@ async function executeBuildScript(operation: Operation, script: string, args: Ar
 
                     logger.logProjectInfo("The project location for " + operation.projectInfo.projectID + " is " + projectLocation, projectID, projectName);
 
+                    // this if loop is used to get the odo app name and set the odo app name in the project info file and reload it back
                     if (operation.projectInfo.projectType == "odo") {
-                        // this if loop is  used to get the odo app name to set the odo app name in the project info file and reload it back
-                        await setOdoAppName(operation.projectInfo);
-                        operation.projectInfo = await getProjectInfo(operation.projectInfo.projectID);
+                        if (!operation.projectInfo.odoAppName) {
```
maysunfaisal: I would have liked it, if the if conditions were  in a single line like`if(operation.projectInfo.projectType == "odo" && !operation.projectInfo.odoAppName)` instead of two if's. But this works as well.

rajivnathan:
This change was adapted as part of https://github.com/eclipse/codewind/pull/623. Thanks @ssh24

PR #623, opened 2019-10-04
Fix Turbine Auto Tests

GeekArthur:
Issue: https://github.com/eclipse/codewind/issues/613

This PR is for:
- Fix the compilation errors in Turbine auto tests
- Ensure ODO extension works as it is
- Enable cache mechanism for Kube projects

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/projectUtil.ts
+++ b/src/pfe/file-watcher/server/src/projects/projectUtil.ts
@@ -77,6 +77,7 @@ export interface ProjectEvent {
     contextRoot?: string;
     isHttps?: boolean;
     appBaseURL?: string;
+    compositeAppName?: string;
 }
 
 export interface ProjectLog {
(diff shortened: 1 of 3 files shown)


src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -428,8 +429,14 @@ async function executeBuildScript(operation: Operation, script: string, args: Ar
...
                         const projectHandler = await projectExtensions.getProjectHandler(operation.projectInfo);
-                        const appBaseURL: string = await projectHandler.getAppBaseURL(projectID);
-                        projectInfo.appBaseURL = appBaseURL.trim();
+                        const projectinfo: ProjectInfo = operation.projectInfo;
+                        const appBaseURL: string = (await projectHandler.getAppBaseURL(projectID)).trim();
+                        const appName: string = (await projectHandler.getAppName(projectID)).trim();
+                        projectInfo.appBaseURL = appBaseURL;
+                        projectInfo.compositeAppName = appName;
+                        projectinfo.appBaseURL = appBaseURL;
+                        projectinfo.compositeAppName = appName;
```
rajivnathan: Do we need both `projectInfo` and `projectinfo`? In the current change they are both referencing the same object. If we really don't want to modify the original object then we should do `odoProjectInfo = Object.assign({}, projectInfo);

GeekArthur: It's very tricky, and it's another naming problem in our existing code. They are referencing different projects.

`projectInfo` uses `ProjectEvent` interface: https://github.com/eclipse/codewind/pull/623/files#diff-7597df0b79304bd2be18940009414316L339, and it's used for Portal socket event, I think change it from `projectInfo` to `projectEvent` makes more sense, but won't break more things still use it.

`projectinfo` uses `ProjectInfo` interface, and it's used for Turbine layer project info.

GeekArthur: Ideally I think, but may do that in separate improvement PR.
`projectEvent` uses `ProjectEvent` interface
`projectInfo` uses `ProjectInfo` interface

rajivnathan: You're right on both accounts. It is using the `projectEvent` type and it is *very* confusing haha. I think at a minimum for this change let's at least rename `projectinfo` to `odoProjectInfo` since it's only used in the scope of this if block.

GeekArthur: Sounds good, and yeah, I was walking the confusing path before you, haha

Issue #396, opened 2019-09-09
Turbine unit test failures


stalleyj:
When I run the Turbine unit tests against master, I consistently see four failures which all seem to be Appsody related. 

228 passing (29s)
  4 failing

  1) PFE - unit test
       actions module
         combinational testing of validate function
           combo24 => validate: good appsody project:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/stalleyj/Development/codewind/src/pfe/file-watcher/server/test/unit-test/unit.test.ts)
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)

  2) PFE - unit test
       actions module
         combinational testing of validate function
           combo25 => validate: appsody project with missing requiredFiles:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/stalleyj/Development/codewind/src/pfe/file-watcher/server/test/unit-test/unit.test.ts)
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)

  3) PFE - unit test
       projectExtensions module
         Combinational testing of getProjectHandler function
           combo6 => getProjectHandler & getProjectCapabilities: projectID dummyappsodyproject:
     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/stalleyj/Development/codewind/src/pfe/file-watcher/server/test/unit-test/unit.test.ts)
      at listOnTimeout (internal/timers.js:531:17)
      at processTimers (internal/timers.js:475:7)

  4) PFE - unit test
       projectExtensions module
         testing of removeProjectHandler function
           removeProjectHandler: dummyappsodyproject:

      AssertionError: expected 'docker' to equal 'appsodyExtension'
      + expected - actual

      -docker
      +appsodyExtension
      
      at Context.<anonymous> (test/unit-test/tests/projectExtensions.module.test.ts:431:53)

maysunfaisal:
I can take this issue and the area should be iterative dev since its iterative dev unit tests for all project types but since appsody extensions involve reading from files, the tests exceed the default mocha 2s timeout..

maysunfaisal:
[...]

PR #636, opened 2019-10-08
Fix unit test microprofile validate tests env variable

ssh24:
## Description 

There was a mismatch of env variable for the microprofile validate tests. This PR switches it from `HIDE_PFE_LOG` to `NODE_ENV` to make sure the logs are consistent and the error is caught.

See comment https://github.com/eclipse/codewind/issues/396#issuecomment-539533985

**Unit test**
Screen Shot 2019-10-08 at 12 39 03 PM

**Microprofile functional test**
Screen Shot 2019-10-08 at 12 38 08 PM

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/libertyProject.ts
+++ b/src/pfe/file-watcher/server/src/projects/libertyProject.ts
@@ -194,7 +194,6 @@ export async function validate(operation: Operation): Promise<void> {
  * @returns Promise<any>
  */
 export async function logBuildEvent(projectInfo: ProjectInfo, msg: String, isError: boolean): Promise<any> {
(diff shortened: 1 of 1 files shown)


ssh24:
@maysunfaisal Could you PTAL?

ssh24:
@rajivnathan I agree with you. The `NODE_ENV` logic was previously there and was changed to `HIDE_PFE_LOG` later on. IMO, it should never have had either of those two env variables in source code.

A better solution which I applied to is to check if the build log path is defined and if it is defined and exists, then append it to the file.

Test result
Screen Shot 2019-10-08 at 3 11 39 PM

PR #637, opened 2019-10-08
Set PFE as the owner of deployed projects and add projectID label

johnmcollier:
- Makes the PFE deployment the owner of all deployed projects in Codewind, rather than the Che workspace
    - PFE and the Performance dashboard are still owned by the Che workspace
    - Has the same effect as before, when PFE goes down, so do the projects
- Adds a projectID label to the project deployments and services, so that the deployment and service can be easily retrieved from processes that don't have access to the helm release name

### DIFF
--- a/src/pfe/file-watcher/scripts/kubeScripts/modify-helm-chart.sh
+++ b/src/pfe/file-watcher/scripts/kubeScripts/modify-helm-chart.sh
@@ -16,6 +16,7 @@ source /file-watcher/scripts/kubeScripts/kube-utils.sh
 deploymentFile=$1
 serviceFile=$2
 releaseName=$3
+projectID=$4
 
 function addOwnerReference() {
     local filename="$1"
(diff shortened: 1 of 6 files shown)


src/pfe/file-watcher/scripts/kubeScripts/modify-helm-chart.sh
```diff
@@ -16,18 +16,22 @@ source /file-watcher/scripts/kubeScripts/kube-utils.sh
...
     local filename="$1"
     local index=$(getIndex $filename metadata.ownerReferences[*].apiVersion)
     yq w -i $filename -- metadata.ownerReferences[+].apiVersion apps/v1
     yq w -i $filename -- metadata.ownerReferences[$index].blockOwnerDeletion true
     yq w -i $filename -- metadata.ownerReferences[$index].controller true
-    yq w -i $filename -- metadata.ownerReferences[$index].kind ReplicaSet
-    yq w -i $filename -- metadata.ownerReferences[$index].name $OWNER_REF_NAME
-    yq w -i $filename -- metadata.ownerReferences[$index].uid $OWNER_REF_UID
+    yq w -i $filename -- metadata.ownerReferences[$index].kind Deployment
+    yq w -i $filename -- metadata.ownerReferences[$index].name $PFE_NAME
```
rajivnathan: Are the `$OWNER_REF_NAME` and `$OWNER_REF_UID` still being set? If so, maybe they're not needed anymore? Alternatively, could change the values of those to match the PFE name and UID?

johnmcollier: They're still being set in the deploy-pfe code, but aren't being used. I changed the name since _technically_ the PFE_UID and PFE_NAME aren't the owner references for PFE

PR #671, opened 2019-10-14
Codewind license testing

ebuckle:
Test script that analyses the dependencies required in the codewind repo:

License checker:
- Performs a clean install `npm ci` of package dependencies when provided with a `package-lock.json` and `package.json`
- Performs a license check on each dependency 
- Clears any packages that have been approved in previous released of Codewind
- Checks an investigation list to either approve or fail packages that have been previously investigated
- Passes or fails remaining packages based on whether their license is eclipse approved
- Searches through dependency files for mentions of GPL to try and find any hidden licensing
- Returns a report containing: Packages Approved for Use, Packages that are Uncleared but contain no issues, Packages that are problematic and might not be suitable for use in Codewind

The current test script will only fail if the license check process itself fails to run. It will log an error if there are any uncleared or problematic files found and display the full report of those packages. 

Later iterations could have the build fail if there are any problematic packages contained.


ebuckle:
Closing until build issues resolved

PR #684, opened 2019-10-15
Add che access token for turbine test

ssh24:
### Description

Initialize access token for turbine functional test on kube with user account `admin/admin`.


ssh24:
@maysunfaisal Could you PTAL?

PR #706, opened 2019-10-16
[0.5.0]fix issue#688

yangcao77:
Fix the issue caused by buildah on K8, since the images listed format has changed from `{{.Repository}}` to `{{.Name}}`

And also the project image name listed by buildah is `localhost/<projectImageName>`, so when we check the existence of the project by checking the image name , we should use `contains` instead of `equals`

### DIFF
--- a/src/pfe/iterative-dev/idc-java/IDC/src/org/eclipse/codewind/iterdev/IDCContext.java
+++ b/src/pfe/iterative-dev/idc-java/IDC/src/org/eclipse/codewind/iterdev/IDCContext.java
@@ -58,6 +58,8 @@ public class IDCContext {
 	private final boolean isWin;
 
 	private final String imageCommand;
+
+	private final String imagesFormatString;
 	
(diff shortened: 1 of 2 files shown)


src/pfe/iterative-dev/idc-java/IDC/src/org/eclipse/codewind/iterdev/IDCContext.java
```diff
@@ -149,6 +151,14 @@ public IDCContext(String rootPassword, String localWorkspaceOrigin, String conta
 		else {
 			this.imageCommand = "docker";
 		}
+
+		if (this.isK8s) {
+			this.imagesFormatString = "\"{{.Name}}\"";
```
rajivnathan: Can you please add a comment here explaining that it's different in K8s because buildah's list format has changed from {{.Repository}} to {{.Name}}

rajivnathan:
Merging it based on the 2nd build being green: https://ci.eclipse.org/codewind/job/Codewind/job/codewind/view/change-requests/job/PR-706/2/

Only a comment was added and we need this to go in ASAP for 0.5.0.

Issue #757, opened 2019-10-21
Update ODO extension installer to accommodate ODO latest release format change


GeekArthur:
Currently ODO store the binaries into the OpenShift mirrors, we need to update the ODO extension installer to accommodate the changes.

GeekArthur:
Please assign this issue to me.

GeekArthur:
Download location before the change on ODO side:
```
https://github.com/openshift/odo/releases/latest/download/odo-linux-amd64
```

Download location after the change on ODO side:
```
https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/
```

rajivnathan:
Fix has been merged to 0.5.0

GeekArthur:
Verified, fix worked.

rajivnathan:
Reopening this one so that the fix can be ported to master as well.

GeekArthur:
We discuss the odo binary version today, we will use the latest version for odo binary on Codewind master branch, but will try to move it to build stage so that we don't need to specify version for each release.

PR #759, opened 2019-10-21
Update ODO binary download link

GeekArthur:
Issue: https://github.com/eclipse/codewind/issues/757

This PR is for updating ODO binary download link.

### DIFF
--- a/src/pfe/portal/modules/ExtensionList.js
+++ b/src/pfe/portal/modules/ExtensionList.js
@@ -24,7 +24,7 @@ const extensionsDir = '/extensions';
 const extensionsPattern = /^(\S+)-(\d+\.\d+\.\d+)\.zip$/; // e.g. extension-name-0.0.1.zip
 const suffixOld = '__old';
 const odoExtensionName = "codewind-odo-extension";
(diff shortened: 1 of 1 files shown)


src/pfe/portal/modules/ExtensionList.js
```diff
@@ -24,7 +24,7 @@ const extensionsDir = '/extensions';
 const extensionsPattern = /^(\S+)-(\d+\.\d+\.\d+)\.zip$/; // e.g. extension-name-0.0.1.zip
 const suffixOld = '__old';
 const odoExtensionName = "codewind-odo-extension";
-const odoBinarySource = "https://github.com/openshift/odo/releases/latest/download/odo-linux-amd64"
+const odoBinarySource = "https://mirror.openshift.com/pub/openshift-v4/clients/odo/latest/odo-linux-amd64"
```
rajivnathan: Can we specify a version instead of downloading latest?

GeekArthur: Good point, we can lock the version to use ODO 1.0.0 (which is a good stable version) for our 0.5.0, then let master uses the latest.

Issue #564, opened 2019-09-26
How to handle .appsody config and repos without codewind-workspace


makandre:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Question synopsis:**

Currently there are instructions telling users to make their local appsody CLI use the same configuration as Codewind's.  This allows them to add a repo in their local CLI, and it shows up in Codewind.

This is possible because Codewind's `.appsody` configuration folder is in `codewind-workspace` which is mounted from host system into PFE.  But this will becoming impossible when the mount goes away.

Questions:

1. Do we still support adding repo in CLI and it shows up in Codewind? (i.e. data flowing from Appsody CLI -> Codewind)

2. Or, now that there's a template manager, do we only support adding appsody repo via template manager?

3. If it is the second, there will be a problem for bind scenario (#292) if there's no data flowing from Codewind -> Appsody CLI.  Consider this scenario:
   1. user adds `repoX` in Codewind, which has `stackY` in it
   2. user wants to bind an existing non-appsody project to `stackY`
   3. during that bind, `codewind-installer` runs `appsody init stackY none`.  This will fail, because the Appsody CLI does not know about `repoX` nor `stackY`

tobespc:
We should be able to still do option 1 when this is all finished.  The smarts is all going to be in the inital creation of the build container but I don't think it will break current behaviour

makandre:
Discussed with @tobespc today.

We concluded there's no good way to continue support for option 1 once the codewind-workspace mount is gone.  Ultimately, to add a repo a user would need to add the `index.yaml` via Appsody CLI **and** add the corresponding `index.json` via Codewind CLI.

That leaves option 2, meaning users will be expected to only add repos via template manager in IDEs.  But additionally for Appsody style repos, IDE also need to add the corresponding `index.yaml` using Appsody CLI (or perhaps IDE makes Codewind CLI do this in a generic manner using extension command (like how it calls `appsody init`), but this needs to be investigated )

@deboer-tim @jopit your thoughts?

jopit:
Investigate if we can make the cli extensible so that it can invoke function in an extension in a generic way. 
IDEs would have to call template cli calls instead of the REST APIs (change already in flight for hybrid)
Portal to work with Andrew on how to get this working.

makandre:
For the proposal above IDEs need to call CLI for template repos management, which is tracked by #841

makandre:
[...]

PR #877, opened 2019-10-28
Notify extensions of add/remove template repository

makandre:
Part of the work for #564 

Requires an updated appsody extension with https://github.com/eclipse/codewind-appsody-extension/pull/53


makandre:
@sghung currently reviewing.  @tobespc would you mind taking a look as well?  Thanks.

Issue #819, opened 2019-10-23
Remote: project remains in PFE after user deletes their project


makandre:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**

Now that `codewind-workspace` is no longer mounted between the user's filesystem and PFE container, when the user deletes their local project, the copy inside PFE will still remain.

~This however does not appear to interfere with user's ability to create another project with the same name.~

**Steps to reproduce:**

**Workaround:**

makandre:
Correction: it does affect recreation of a project with the same name.  The new project is getting uploaded to a subdirectory under the first directory:

![image](https://user-images.githubusercontent.com/17596240/67415259-b0b20d80-f592-11e9-8995-23be2e5a7224.png)

makandre:
The nested dirs would be fixed by https://github.com/eclipse/codewind/pull/820.  However the old project contents will be merged with the new project contents as long as the old dir is not deleted.

maysunfaisal:
I have seen this as well #869 

Looks like the old projects goes away from the PFE container if I remove the volume `docker volume rm -f <volume>`

james-wallis:
@tobespc assign me please

james-wallis:
@makandre the PR has been merged for this, could you retest and close please.

makandre:
Verified it on latest code this morning.  Thanks for the fix!

yangcao77:
I'm still seeing this issue on che.

maysunfaisal:
This issue is still there if users dont delete their projects i.e.;

create proj xyz -> stop.sh -> docker prune -> run.sh -> proj xyz still exists

because the codewind docker volume is still present. Ideally IMO the PFE should not have it I am doing a prune and building from scratch. But I guess if I dont delete a proj, I should expect the docker volume to have it on a newer image. Tho this was not happening before we switched.

makandre:
`docker system prune` need to have `--volume` flag to delete volumes as well

But how does that remove the project code before the switch?

maysunfaisal:
Before we relied on bind mount ie my CW workspace on my laptop <-> PFE workspace, and thus if you had projects there, PFE would build it on restart. But if you check `docker-compose.yaml` we now have `volumes` since we no longer use bind mount.. This allows our project to be anywhere on our local disk

PR #892, opened 2019-10-29
Delete project directory in container

james-wallis:
### Summary
Added a line to remove the project directory from the PFE container when the user deletes a project.
Fix for #819

### Testing
I've manually tested that adding this line deletes the folder.
 

james-wallis:
@hhellyer made the point that `fs.remove` is slower than using exec and was replaced in microclimate - https://github.ibm.com/dev-ex/microclimate/pull/3099. 

I'm going to make the changes to this branch.

hhellyer:
Moved to WIP so we don't land this, there are known performance issues using `fs.remove()` on large directory trees.

PR #895, opened 2019-10-29
Use fs instead of exec commands

james-wallis:
### Summary
We already include the `fs-extra` NPM module in PFE which both adds more `fs` functions to remove directories etc and also add promise support to all existing `fs` commands. I've changed a lot of our `exec` commands to use the module which should improve code readability.

### Testing
I've manually testing the changes aside from the ones in `ExtensionList.js` as I'm not sure how to test them.


src/pfe/portal/modules/ExtensionList.js
```diff
@@ -68,8 +68,9 @@ module.exports = class ExtensionList {
               await exec(`unzip ${source} -d ${targetDir}`);
               
               if (name == odoExtensionName) {
-                await exec(`mkdir -p ${targetWithVersion}/bin`);
-                await exec(`mv ${odoBinarySource} ${targetWithVersion}/bin/odo && chmod +x ${targetWithVersion}/bin/odo`);
+                await fs.ensureDir(`${targetWithVersion}/bin`);
+                await fs.move(odoBinarySource, `${targetWithVersion}/bin/odo`); 
+                await fs.chmod(`${targetWithVersion}/bin/odo`, '755');
```
james-wallis: @rajivnathan could you review this please? I am unsure how I can test it. Thanks

rajivnathan: Sorry for the late reply, missed this initially. Thought I saw some comments in another issue that some `fs` functions were significantly slower than `exec`? If that's the case should we leave this as is?

rajivnathan: Found the comment: https://github.com/eclipse/codewind/pull/892#issuecomment-547405928

james-wallis: The issue with `fs.remove` was that it calls `rimraf` under the covers and deletes each individual file after the other which is why it has the performance hit. I've not seen anything else about other  `fs` commands having a performance hit.

rajivnathan: Jingfu & Len mentioned we can't test this change until https://github.com/eclipse/codewind/issues/897 is fixed

GeekArthur: @james-wallis We can definitely help review and test this PR for ODO extension, but since the changes are for ODO extension auto install, so we have to ensure the project creation is successful then we can test/approve/merge this PR. Please contact Toby to make #897 fix ASAP, thanks!

james-wallis: Ok understood, thanks. 
Depending on how long that fix will take (I'll check with Toby tomorrow) I may split this PR into the commands we can test and the three above.

james-wallis:
Linter errors are making this PR fail to pass the Jenkins test. Once 897 is resolved I'll fix them.

Issue #883, opened 2019-10-28
[Hybrid] Client adaptation to new log architecture


ssh24:
**Description:**
As a result of the hybrid changes the log architecture will change https://github.com/eclipse/codewind/issues/681. Portal will need to adapt to the new log architecture and the new event that will be emitted.

An example of the new project emittion object:
```
{
    "build": [
      {
        "origin": "container",
        "files": [
          "/home/default/logs/maven.build.log"
        ]
      },
      {
        "origin": "workspace",
        "files": [
          "/codewind-workspace/.logs/mylib-abebda90-f69d-11e9-954f-51fd5366b98f/docker.build.log"
        ]
      }
    ],
    "app": [
      {
        "origin": "workspace",
        "files": [
          "/codewind-workspace/.logs/mylib-abebda90-f69d-11e9-954f-51fd5366b98f/app.log"
        ]
      },
      {
        "origin": "container",
        "files": [
          "/home/default/logs/message.log",
          "/home/default/logs/console.log",
        ]
      }
    ]
  }
```

Once, portal fixes their streaming function be able to stream from both container and workspace logs, IDE should be able to show it as well. Though not sure how much work is required from the IDE side to adapt to the new change.

/cc @hhellyer @tetchel

tetchel:
@tobespc 👀

tetchel:
everything seems ok from my testing just now

ssh24:
Yes, I just saw it work after a while as well :\

Is this intermittent behavior from portal or IDE?

PR #899, opened 2019-10-29
New logs architecture

ssh24:
### Description

This PR enables the new log architecture required after the remote architecture. Now logs can be either in the workspace (PFE) or in the app container.

The returned lists of logs will show the log files in sorted order with the most recent logs on top of the array. Here is a snippet of the new logs data:
```
  "logs": {
    "build": [
      {
        "origin": "container",
        "files": [
          "/home/default/logs/maven.build.log"
        ],
        "containerName": "cw-mylib-5942e080-f991-11e9-8901-8f57ecbc22f8"
      },
      {
        "origin": "workspace",
        "files": [
          "/codewind-workspace/.logs/mylib-5942e080-f991-11e9-8901-8f57ecbc22f8/docker.build.log"
        ]
      }
    ],
    "app": [
      {
        "origin": "workspace",
        "files": [
          "/codewind-workspace/.logs/mylib-5942e080-f991-11e9-8901-8f57ecbc22f8/app.log"
        ]
      },
      {
        "origin": "container",
        "files": [
          "/home/default/app/mc-target/liberty/wlp/usr/servers/defaultServer/logs/console.log",
          "/home/default/app/mc-target/liberty/wlp/usr/servers/defaultServer/logs/messages.log"
        ],
        "containerName": "cw-mylib-5942e080-f991-11e9-8901-8f57ecbc22f8"
      }
    ]
  }
```

As seen above, previously the `build` or `app` log was just an object and the primary switch now has been to make it an **array of build/app objects**. Previously, we would only do one level of sorting, that is only sorting of the files in the `files` keyword but now the sorting is done on two levels: 1. as previously, with the `files` keyword and 2. within the object that can be either `container` or `workspace`.

For example, on the snippet above, you can see that there are two entries for build logs -> maven build and docker build. Previously, both maven build and docker build would reside in the workspace but now maven build resides in the app container whereas the docker build resides in workspace. And the order of the application built, docker build is done first, followed by the maven build. Hence, you can see the object containing maven.build is on top of the build logs array followed by the docker build log.

More interestingly if you look at the app log array, you will notice that there are 3 log files, 2 of them coming directly from the container and 1 from the workspace. And the order they are generated, app.log is the most recent log file. Messages.log was the first generated and then console.log was made available, thus showing the 2 levels of sorting as console.log is on top of the `files` keyword.

Once this PR lands, our client Portal will need to adapt to this https://github.com/eclipse/codewind/issues/883 for enabling the stream again.

**Required PR in Appsody**: https://github.com/eclipse/codewind-appsody-extension/pull/54
**Required PR in ODO**: https://github.com/eclipse/codewind-odo-extension/pull/17


src/pfe/file-watcher/server/src/projects/OdoExtensionProject.ts
```diff
@@ -23,14 +23,40 @@ import { ProjectInfo, BuildLog, AppLog, ProjectCapabilities, defaultProjectCapab
 import { Validator } from "./Validator";
 import { IExtensionProject } from "../extensions/IExtensionProject";
 
+// skeleton for the logs originated from the extension json
```
ssh24: @jingfuwang Could you please review the ODOExtensionProject part of this PR? Thanks.

src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
```diff
@@ -208,26 +234,35 @@ export class ShellExtensionProject implements IExtensionProject {
...
+     * @param containerName <Required | String> - The docker container name.
      *
-     * @returns Promise<BuildLog>
+     * @returns Promise<Array<AppLog | BuildLog>>
      */
-    getBuildLog = async (logDirectory: string): Promise<BuildLog> => {
-        return await logHelper.getBuildLogs(logDirectory, this.config.buildLogs);
+    getLogs = async (type: string, logDirectory: string, projectID: string, containerName: string): Promise<Array<AppLog | BuildLog>> => {
+        if (type.toLowerCase() != "build" && type.toLowerCase() != "app") return;
+        this.setLogsOriginFromExtension();
```
makandre: I think we discussed that this could be set right after we load the config since it won't change, so we don't need to run this every time `getLogs` is called.

src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
```diff
@@ -208,26 +234,35 @@ export class ShellExtensionProject implements IExtensionProject {
...
-     * @description Get the app log for the project.
-     *
-     * @param logDirectory <Required | String> - The log location directory.
-     *
-     * @returns Promise<AppLog>
-     */
-    getAppLog = async (logDirectory: string): Promise<AppLog> => {
-        return await logHelper.getAppLogs(logDirectory, this.config.appLogs);
+    setLogsOriginFromExtension(): void {
+        if (this.config.buildContainerLogs) {
```
makandre: Should we do some sanitization or error check on these properties here, to guard against extensions that define these incorrectly?  (if you're doing that in `logHelper.getLogs` then it's fine not to do it here)

ssh24: yup, its all typed against the interfaces in `logHelper`, so if the type check fails it will fail

src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
ssh24: Sure, that makes sense.

src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
makandre: But Typing in `ts` is only for compile time checking.  This doesn't guard at runtime.

src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
```diff
@@ -208,26 +235,35 @@ export class ShellExtensionProject implements IExtensionProject {
...
-     * @description Get the app log for the project.
-     *
-     * @param logDirectory <Required | String> - The log location directory.
-     *
-     * @returns Promise<AppLog>
-     */
-    getAppLog = async (logDirectory: string): Promise<AppLog> => {
-        return await logHelper.getAppLogs(logDirectory, this.config.appLogs);
+    setLogsOriginFromExtension(): void {
+        if (this.config.buildContainerLogs && (this.config.buildContainerLogs.files || this.config.buildContainerLogs.dirs)) {
```
makandre: It's better, but if `files` or `dirs` are of the wrong type (for example, it's a number or boolean), would `logHelper` tolerate it?

It may be better to consolidate the sanitization into a private function within the `logHelper` itself

makandre: I see, tho just to be safe, could we add a runtime type check here:

https://github.com/eclipse/codewind/pull/899/files#diff-44a052830863e32d7b0f6a81d8d83584R264

`const logs = Object.keys(logsOrigin);` 
(make logs an empty array if `logsOrigin` is not an object)

ssh24: Not sure what the runtime check do though? If `logsOrigin` is not of type 
```
interface ILogOriginFolderOrFiles {
    [x: string]: string;
}
```
it will throw an error, even on runtime

makandre: `typeof logsOrigin === "object"`?

If being the wrong type is causing a runtime error to be thrown, that's the kind of problem we're trying to guard against isn't it?

ssh24: Yup. Good catch. Added that fix:
```
const logs = (typeof logsOrigin === "object" && logsOrigin !== null) ? Object.keys(logsOrigin) : [];
```

ssh24: Actually, improved the condition even better. Thanks to https://stackoverflow.com/a/52478680

```
const logs = (typeof logsOrigin == "object" && logsOrigin instanceof Object && !(logsOrigin instanceof Array)) ? Object.keys(logsOrigin) : [];
```

This will trim it down and only accept `{}` as an object.

hhellyer:
@elsony @ssh24 - This PR should probably have been squashed before committing, additionally lines like:
https://github.com/eclipse/codewind/pull/899/files#diff-895a69e551a600295f919319ea90f10fR109
seems to contain spurious changes that shouldn't have been merged.

ssh24:
Not sure how those extra spaces were generated. I will fix them on separate PR

PR #905, opened 2019-10-29
Expose projects over ingress

johnmcollier:
This PR updates Turbine so that after a project is deployed, it creates an ingress resource exposing that project. This allows projects to be accessed in the browser when NodePorts aren't necessarily exposed (such as on OCP 4.x or GKE).

The following is done in this PR to achieve that:
- `projectID` label added to every project deployment and service (via `modify-helm-chart.sh`), allowing for an easy way to retrieve project resources
- If the project is successfully deployed **and** an ingress for the project doesn't already exist:
     - Retrieves the PFE ingress domain (from the poorly named `CHE_INGRESS_HOST`) and appends it to the project name to create a new unique ingress URL for the project
         - Unique because the PFE ingress domain is per-workspace
     - Generates an ingress resource, setting the project's ingress domain and pointing it to the service
         - The project's deployment is the `ownerReference` for the ingress
    - Deploys the ingress resource

Unrelated to the ingress support, this PR also pulls in the changes from https://github.com/eclipse/codewind/pull/637


src/pfe/file-watcher/server/src/utils/kubeutil.ts
```diff
@@ -333,6 +333,98 @@ export async function installChart(projectID: string, deploymentName: string, ch
...
+    } catch (err) {
+        logger.logProjectError("Unable to retrieve project deployment or service", projectID);
+        throw err;
+    }
+
+    try {
+
+        // Create and deploy the ingress on Kube if one for the project doesn't already exist
+        const resp = await k8sClient.apis.extensions.v1beta1.namespaces(KUBE_NAMESPACE).ingresses.get({ qs: { labelSelector: "projectID=" + projectID } });
+        if (resp.body.items.length == 0) {
```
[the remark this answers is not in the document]
johnmcollier: Now that I think about it, I think we'll need to refresh the ingress each time, in case the port does change.

This should be a relatively quick change.

src/pfe/file-watcher/server/src/utils/kubeutil.ts
```diff
@@ -333,6 +333,98 @@ export async function installChart(projectID: string, deploymentName: string, ch
...
 }
 
+export async function exposeOverIngress(projectID: string, projectName: string, ingressDomain: string, isHTTPS: boolean): Promise<string> {
+    let ownerReferenceName: string;
+    let ownerReferenceUID: string;
+    let serviceName: string;
+    let servicePort: number;
+
+    try {
+        let resp: any = undefined;
```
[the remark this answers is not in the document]
johnmcollier: D'oh, good catch. I'll fix that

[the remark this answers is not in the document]
johnmcollier: Actually, since we'll be refreshing the ingress each time, this shouldn't be needed.

I did change the code that retrieves the service and deployment so that we would error out if it couldn't be found, as we'd run into issues trying to create an ingress pointing to a blank service.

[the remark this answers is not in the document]
rajivnathan: Ok, makes sense. Can we remove the `deployments.get` call though? It looks like we're only using it to get the ownerReferenceName and ownerReferenceUID for which both should be available from the service which is the next call.

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -440,6 +440,17 @@ async function executeBuildScript(operation: Operation, script: string, args: Ar
                         odoProjectInfo.compositeAppName = appName;
                         await projectsController.saveProjectInfo(projectID, odoProjectInfo, true);
                     }
+                    else if (operation.projectInfo.projectType != "appsody") {
```
makandre: Just a small reminder here, we can treat appsody the same (if all goes as planned)

johnmcollier: Want me to change it now, or after the appsody projectID change is delivered?

johnmcollier: Don't want to break appsody on Kube if the needed change in Appsody isn't there yet.

I can open another PR once the CLI is updated

makandre: Can change it now if not finding the appropriate service is handled gracefully

johnmcollier: KK, fixed

rajivnathan:
John mentioned he'd like to cover a bit more testing before this goes in so added the WIP to the title

Issue #915, opened 2019-10-30
Turbine Unit Test Broken


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
```

/root/codewind/src/pfe/file-watcher/server/node_modules/ts-node/src/index.ts:240
    return new TSError(diagnosticText, diagnosticCodes)
           ^
TSError: ⨯ Unable to compile TypeScript:
test/unit-test/tests/logHelper.module.test.ts:269:39 - error TS2339: Property 'split' does not exist on type 'LogFiles'.

269                 const tokens = result.split("/");
```

**Steps to reproduce:**

**Workaround:**

ssh24:
I am looking into this. @rajivnathan Could you please assign this to me?

ssh24:
@maysunfaisal Could you please verify and close this?

maysunfaisal:
verified

PR #921, opened 2019-10-30
Fix turbine unit tests

ssh24:
### Description

This PR fixes the turbine unit test failures. It also takes care about the check on `process.env.IN_K8` becaue in typescript all env variables are processed as strings.

Related to https://github.com/eclipse/codewind/issues/915

Screen Shot 2019-10-30 at 12 41 43 PM


ssh24:
@maysunfaisal Could you please review?

PR #940, opened 2019-10-31
Add the node attach to the launch.json and change the npm run dev command

james-wallis:
### Summary
Add VSCode utils to development code so its there for everyone in development and there won't be any worry about doing a `git add .` and adding them.

Also removes some linting errors from the tests.


src/pfe/portal/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "start": "./npm-start.sh",
     "prod": "node server.js",
-    "dev": "nodemon server.js",
+    "dev": "node --inspect=0.0.0.0:9777 server.js",
```
rajivnathan: Can we keep nodemon?

james-wallis: I've changed it to keep nodemon and use your `devfile.env`.

PR #952, opened 2019-11-04
[WIP] Change directory used inside the Docker volume to be project.name+project.id

james-wallis:
### Summary
This PR changes the directory that is used inside the Docker volume (`/codewind-workspace` in PFE) to take the form of project.name-project.id (`mondaytest-f7cb5c90-ff05-11e9-8761-9bf62d92b58b`). This will allow the user to have multiple projects of the same name.

### Testing
I've run manual tests to see that project creation works + changing code.
I've added unit tests for Project.js which check that the directory is being set correctly.

### Other bits
* Re-enabled a few tests and moved from NOTRUNNING folder. - Now run 148 tests.
* `npm run test` now runs unit tests as well as API tests


test/src/unit/modules/Project.test.js
```diff
@@ -0,0 +1,150 @@
...
+            project.loadConfig = obj.loadConfig;
+    
+            project.should.containSubset(obj);
+    
+            const json = project.toJSON();
+            json.should.not.containSubset(obj);
+        });
+    })
+    describe('getPort', () => {
+        it('Checks that the port returned is internal as RUNNING_IN_K8S is false', () => {
```
jcockbain96: You could separate these into 2 describe blocks. i.e 1 for `RUNNING_IN_K8S` is false, and the other for the reverse. I'd say that'd make the logs a bit easier to follow, and mean adding new tests simpler

test/src/unit/modules/Project.test.js
```diff
@@ -0,0 +1,150 @@
...
+        });
+    })
+    describe('projectPath', () => {
+        it('Checks that projectPath() is workspace+directory', () => {
+            const project = new Project({ name: 'dummy', directory: 'directory' }, '/codewind-workspace/');
+            const projectPath = project.projectPath();
+            projectPath.should.equal('/codewind-workspace/directory');
+        });
+    })
+    describe('isValidName', () => {
```
jcockbain96: I think this is now an obsolete function, we moved to an API `/invalidNameChers`.

jcockbain96: That might not be relevant, but if it's not used we may want to get rid.

james-wallis:
Although its been discussed that we don't want duplicate names and so there is no need to move to an ID I believe this code simplifies the process of getting the project directory and temp directory so I'll keep it up to date and have it ready for post decut. 

As it is this branch needs to be rebased.

james-wallis:
Stale so closing.

Issue #908, opened 2019-10-29
Add/update Turbine tests for the new log architecture


ssh24:
### Description

Once https://github.com/eclipse/codewind/issues/681 goes in we need to update the test automation to work with the new log architecture.

ssh24:
@rajivnathan Could you please assign this to me?

Issue #907, opened 2019-10-29
[Iterative-dev]Update Turbine Win Func Tests


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
Currently we are using a cwctl installer from ages ago.

To pick up the hybrid/remote changes from the cwctl; we need to download and install a new one during every Turbine Win cronjob..

**Steps to reproduce:**

**Workaround:**

jgwest:
@maysunfaisal In case you hadn't already seen this, you can steal [cli-pull.sh from here](https://github.com/eclipse/codewind-che-plugin/blob/master/codewind-che-sidecar/scripts/cli-pull.sh) to automaticallly download the latest cwctl. No parameters needed, just invoke with `CW_CLI_BRANCH=master ./cli-pull.sh` (env var is optional).

maysunfaisal:
sweet, i did see that change, thx for reminding~

ssh24:
@rajivnathan I am looking into this. Could you please assign this to me?

ssh24:
This is now working. @maysunfaisal Could you close this issue?

PR #967, opened 2019-11-04
Add tests on new log architecture

ssh24:
### Description

This PR:
- [x] adds the tests on new log architecture
- [x] enables the cronjob on windows to pull down and use the latest installer
- [x] enables the linux/macos to pull down and use the latest installer from the vm instead of inside PFE

Related to https://github.com/eclipse/codewind/issues/908 and https://github.com/eclipse/codewind/issues/907

### DIFF
--- a/src/pfe/file-watcher/server/src/utils/dockerutil.ts
+++ b/src/pfe/file-watcher/server/src/utils/dockerutil.ts
@@ -314,7 +314,7 @@ export async function getFilesOrFoldersInContainerWithTimestamp(projectID: strin
         return [];
       } else {
         logger.logInfo("At least one file was found");
-        const lines =  data.split("\n");
+        const lines = data.split("\n");
         const fileIndex = 8; // file is the 9th argument from the above command
         const filesTimestamp: Array<logHelper.LogFiles> = [];
         for (let line of lines) {
--- a/src/pfe/file-watcher/server/src/utils/dockerutil.ts
+++ b/src/pfe/file-watcher/server/src/utils/dockerutil.ts
@@ -314,20 +314,23 @@ export async function getFilesOrFoldersInContainerWithTimestamp(projectID: strin
         return [];
       } else {
         logger.logInfo("At least one file was found");
-        const files = data.replace(/\n/g, " ").replace(/[^ -~]+/g, "").split(" ").filter((entry: string) => { return entry.trim() != ""; });
+        const lines =  data.split("\n");
         const fileIndex = 8; // file is the 9th argument from the above command
         const filesTimestamp: Array<logHelper.LogFiles> = [];
-        for (let index = fileIndex; index < files.length; index = index + fileIndex + 1) {
-          const file = files[index];
-          const dateString = files[index - 3] + " " + files[index - 2] + " " + files[index - 1];
-          if (!folderName) {
-            filesTimestamp.push({file: path.join(fileLocation, file), time: +new Date(dateString)});
-          } else {
-            if (file.toLowerCase() === folderName.toLowerCase()) {
-              filesTimestamp.push({file: path.join(fileLocation, folderName), time: +new Date(dateString)});
-              break;
+        for (let line of lines) {
+            line = line.replace(/[^ -~]+/g, "").replace(/\s+/g, " ").trim();
+            if (line != "") {
+                const file = line.split(" ")[fileIndex];
+                const dateString = line.split(" ")[fileIndex - 3] + " " + line.split(" ")[fileIndex - 2] + " " + line.split(" ")[fileIndex - 1];
+                if (!folderName) {
+                    filesTimestamp.push({file: path.join(fileLocation, file), time: +new Date(dateString)});
+                } else {
+                    if (file.toLowerCase() === folderName.toLowerCase()) {
+                        filesTimestamp.push({file: path.join(fileLocation, folderName), time: +new Date(dateString)});
+                        break;
+                    }
+                }
             }
-          }
         }
         return filesTimestamp;
       }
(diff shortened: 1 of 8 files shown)


ssh24:
@maysunfaisal Could you please review?

src/pfe/file-watcher/server/src/utils/dockerutil.ts
```diff
@@ -314,20 +314,23 @@ export async function getFilesOrFoldersInContainerWithTimestamp(projectID: strin
...
-        for (let index = fileIndex; index < files.length; index = index + fileIndex + 1) {
-          const file = files[index];
-          const dateString = files[index - 3] + " " + files[index - 2] + " " + files[index - 1];
-          if (!folderName) {
-            filesTimestamp.push({file: path.join(fileLocation, file), time: +new Date(dateString)});
-          } else {
-            if (file.toLowerCase() === folderName.toLowerCase()) {
-              filesTimestamp.push({file: path.join(fileLocation, folderName), time: +new Date(dateString)});
-              break;
+        for (let line of lines) {
```
maysunfaisal: Can you explain what changes are being made here?

src/pfe/file-watcher/server/src/utils/kubeutil.ts
```diff
@@ -227,23 +227,26 @@ export async function getFilesOrFoldersInContainerWithTimestamp(projectID: strin
...
-                    filesTimestamp.push({file: path.join(fileLocation, folderName), time: +new Date(dateString)});
-                    break;
-                }
+                logger.logInfo("At least one file was found");
+                const lines = data.stdout.split("\n");
+                const fileIndex = 8; // file is the 9th argument from the above command
+                const filesTimestamp: Array<logHelper.LogFiles> = [];
+                for (let line of lines) {
+                    line = line.replace(/[^ -~]+/g, "").replace(/\s+/g, " ").trim();
+                    if (line != "") {
```
maysunfaisal: Can you explain what changes are being made here?

src/pfe/file-watcher/server/test/cronjob/cronjob-win.sh
```diff
@@ -95,16 +97,33 @@ function checkExitCode() {
...
+   ./$EXECUTABLE_NAME project create --url $1 $2
+}
+
+function copyToPFE() {
+   docker cp $1 $CW_CONTAINER:"/codewind-workspace"
+}
+
+echo -e "${BLUE}>> Downloading latest installer ... ${RESET}"
+EXECUTABLE_NAME="cwctl"
+curl -X GET http://download.eclipse.org/codewind/codewind-installer/master/latest/cwctl-win.exe --output $EXECUTABLE_NAME
```
maysunfaisal: Since we're doing a curl output to current location, will the next cron run overwrite the old executables?

src/pfe/file-watcher/server/test/functional-test/configs/log.config.ts
```diff
@@ -14,13 +14,69 @@
...
 
 export const logTypes = ["build", "app"];
+const logOrigins = ["container", "workspace"];
 
 export const logFileMappings: any = {
     "docker": {
-        [logTypes[0]]: [log_names[0]],
-        [logTypes[1]]: [log_names[3]],
-    }
+        [logTypes[0]]: [{
```
maysunfaisal: nice

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -24,29 +24,49 @@ Options:
...
-        kubectl exec -i $CODEWIND_POD_ID -- bash -c "$DOWNLOAD_CMD"
-        kubectl exec -i $CODEWIND_POD_ID -- bash -c "chmod +x /usr/local/bin/$EXECUTABLE_NAME"
+    HOST_OS=$(uname -a)
+    if [[ "$HOST_OS" =~ "Darwin" ]]; then
+        extension="macos"
+    elif [[ "$HOST_OS" =~ "Linux" ]]; then
+        extension="linux"
     fi
+    echo "Extension is $extension"
+    curl -X GET http://download.eclipse.org/codewind/codewind-installer/master/latest/cwctl-$extension --output $EXECUTABLE_NAME
```
maysunfaisal: switched back to saving cwctl to pwd?

src/pfe/file-watcher/server/src/utils/dockerutil.ts
ssh24: This is a change to read the `ls` output line by line instead of reading them all together and indexing them. Reading them all together and then indexing them may causes issues when the last column on the ls output have spaces...for example, the file might be symlinked to another file and have spaces in the output causing the indexing to be messed up. So a better solution is to use it line by line.

src/pfe/file-watcher/server/src/utils/kubeutil.ts
ssh24: Same as dockerutils.

src/pfe/file-watcher/server/test/cronjob/cronjob-win.sh
ssh24: we remove it after the test run

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: yes because this will now be on the users machine (i.e on the test VM) and we should not be writing and installing cwctl to their /usr/local/bin folders. So better to just do it within our test folder and also gitignored it so it does not get comitted accidentally. We also remove it after the test run.

Issue #943, opened 2019-10-31
Separate "sync" and "build" ignore filters for filewatcherd


makandre:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description of the enhancement:**

@jgwest can explain this better.

There is currently one set of `ignorePath` filter, which controls what files should not trigger a build, and hence should not be synced from the filesystem to PFE.

But this does not support the use case of extension projects (such as Appsody) that perform their own file monitoring to determine when to build.

In these cases, all files should be synced to PFE, but they should **not** trigger Codewind to build the project.

So there should be separate filters to control the "sync" and "build" behaviour

Related: #818 #939

**Proposed solution:**

jgwest:
Exactly: previously we had only one filter, which controlled (only) filewatcher's file change detection.

However, if Appsody is going to use a `*/*` filter to prevent builds, we need to separate the filters as Andrew described above (likely this work would be in PFE).

For more detail and a diagram of the workflow, see my comment here:
https://github.com/eclipse/codewind/issues/818#issuecomment-545573418

yangcao77:
I can work on this issue.

elsony:
One thing that I have discussed on a meeting with @makandre  is that if we can stay on only one list instead of creating a new list for it.  At the end of the day, it is the Appsody extension implementation that decides that it doesn't need build.  Andrew will take a look at to see if it is possible to setup the properly list of files for the sync build do a no op when Appsody get called for a build.

makandre:
Yes I checked the code, and for a "build" the extension already does very little; we don't take the container down.  But minimally we still had to issue some build/app status change events (this was needed to address some state display issues in IDE).  So in turn that would cause project status events etc. to be emitted by Turbine.

It's not the end of the world but it does result in some unnecessary event and state changes.  If we had a filter that allowed us to not trigger a Codewind build on file changes, that would eliminate those events altogether.

yangcao77:
In my opinion, Turbine does not need a separate filter. Since for other project types, the filter will be exactly the same, and the filtering is already done by cwctl/filewatcherd. Appsody project is the only exception. 
In addition, the `ignoredPaths` filter is controller by user in the `.cw-settings` file; however, appsody implementation does not support that, and still decides whether to build or not by itself. 

[...]

PR #1016, opened 2019-11-06
Add another property to all project handlers to decide build or not

yangcao77:

For issue: https://github.com/eclipse/codewind/issues/943

This PR adds a property called `builtByExtension` to all project handlers to indicate if  the build of this project type should be done by the project extension or not.

### DIFF
--- a/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
+++ b/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
@@ -78,6 +78,8 @@ export async function updateProjectForNewChange(projectID: string, timestamp: nu
         let isSettingFileChanged = false;
         let isProjectBuildRequired = false;
 
+        const projectHandler = await projectExtensions.getProjectHandler(projectInfo);
+        const builtByExtension = projectHandler.builtByExtension;
         try {
             for (let i = 0; i < eventArrayLength; i++) {
                 if (isSettingFileChanged && isProjectBuildRequired) {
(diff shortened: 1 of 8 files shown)


yangcao77:
@rajivnathan @makandre Can you help review the change?

src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
```diff
@@ -78,6 +78,8 @@ export async function updateProjectForNewChange(projectID: string, timestamp: nu
         let isSettingFileChanged = false;
         let isProjectBuildRequired = false;
 
+        const projectHandler = await projectExtensions.getProjectHandler(projectInfo);
+        const builtByExtension = projectHandler.builtByExtension;
```
makandre: Could we skip the `try` block below entirely, then `isProjectBuildRequired` will never be set to `true`

yangcao77: We need to find out if the `.cw-settings` file has been changed, and need to proceed project settings change.

src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
```diff
@@ -78,6 +78,8 @@ export async function updateProjectForNewChange(projectID: string, timestamp: nu
         let isSettingFileChanged = false;
         let isProjectBuildRequired = false;
 
+        const projectHandler = await projectExtensions.getProjectHandler(projectInfo);
```
makandre: Since you got the project handler here, you may not need to get it again later inside this function (I think there are 2 places below that calls `await projectExtensions.getProjectHandler`

src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
```diff
@@ -86,6 +86,7 @@ export class ShellExtensionProject implements IExtensionProject {
 
     supportedType: string;
     // defaultIgnoredPath: string[] = ["*/*"];
+    builtByExtension: boolean = true;
```
makandre: thanks for this change, could you remove the commented out line above as well?

src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
yangcao77: updated

src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
yangcao77: updated

Issue #1009, opened 2019-11-06
Filewatcher.projectChanged throwing error before project starts


ssh24:
**Description:**
If I create a project, the `projectChanged` function is throwing error just before the app starts. Though there is no issue with the app starting or build affecting it, not sure what might be causing this:

```
[06/11/19 16:10:26 myspr] [INFO] Build completed for 4bb21420-00af-11ea-8d51-9982ecc43880
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is .classpath projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is .cw-settings projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is .gitignore projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is .project projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is Dockerfile projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is Dockerfile-tools projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is README.md projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is bin/.project projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is chart/springjavatemplate/Chart.yaml projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is chart/springjavatemplate/bindings.yaml projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is chart/springjavatemplate/templates/basedeployment.yaml projectRoot is /root/app
[06/11/19 16:10:29 /portal/routes/projects/remoteBind.route.js] [INFO] project is [object Object] file is chart/springjavatemplate/templates/deployment.yaml projectRoot is /root/app
[...]

PR #1030, opened 2019-11-07
Fix timestamp check condition on updateProjectForNewChange

ssh24:
### Description

The timestamp check condition on `updateProjectForNewChange` function would fail if the number was 0. This was because it would literally treat it as boolean. Changed the condition to take care of the null cases as well as the case when it is not a number.

Related to https://github.com/eclipse/codewind/issues/1009


ssh24:
@rajivnathan Could you PTAL?

PR #1077, opened 2019-11-11
Codewind license testing

ebuckle:
Test script that analyses the dependencies required in the codewind repo:

License checker:
- Performs a clean install npm ci of package dependencies when provided with a package-lock.json and package.json
- Performs a license check on each dependency
- Clears any packages that have been approved in previous released of Codewind
- Checks an investigation list to either approve or fail packages that have been previously investigated
- Passes or fails remaining packages based on whether their license is eclipse approved
- Searches through dependency files for mentions of GPL to try and find any hidden licensing
- Returns a report containing: Packages Approved for Use, Packages that are Uncleared but contain no issues, Packages that are problematic and might not be suitable for use in Codewind

The current test script will only fail if the license check process itself fails to run. It will log an error if there are any uncleared or problematic files found and display the full report of those packages.

Later iterations could have the build fail if there are any problematic packages contained.


ebuckle:
Jenkins build is failing due to a test timeout. This runs perfectly fine locally so I'm not sure what is causing the issue.

Screen Shot 2019-11-11 at 15 48 24

test/src/license.test.js
```diff
@@ -0,0 +1,52 @@
...
+        const packageJsonPath = path.resolve(allPackageJsonPaths[key], 'package.json');
+        let packageLockJson = null;
+        if (fs.existsSync(path.resolve(allPackageJsonPaths[key], 'package-lock.json'))) {
+            const packageLockJsonPath = path.resolve(allPackageJsonPaths[key], 'package-lock.json');
+            packageLockJson = JSON.parse(fs.readFileSync(packageLockJsonPath));
+        }
+        const packageJson = JSON.parse(fs.readFileSync(packageJsonPath));
+
+        describe(`${key}`, function() {
+            it('perform license check', async function() {
```
jcockbain96: could this it block message context, as to what it's checking ?

test/src/license.test.js
```diff
@@ -0,0 +1,52 @@
+/* eslint-disable no-loop-func */
```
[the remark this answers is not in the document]
jcockbain96: This currently won't be run as part of PR builds / npm test ? Is this intended ?

test/src/license.test.js
```diff
@@ -0,0 +1,52 @@
...
+
+const allPackageJsonPaths = require('../config/index').packageJsons;
+const fs = require('fs-extra');
+const path = require('path');
+const mlog = require('mocha-logger');
+const chai = require('chai');
+const performCheck = require('../modules/license.service').performCheck;
+
+describe('check node package licenses', function() {
+    this.timeout('5m');
```
jcockbain96: In other tests we use `config.testTimeout` object

test/modules/license.service.js
```diff
@@ -0,0 +1,265 @@
...
+    const unclearedPackages = {};
+    const problemPackages = {};
+
+    await CompareResultsWithExistingClearance(acceptedPackages, unclearedPackages, licenseCheckerOutput);
+    await FilterInvestigatedPackages(acceptedPackages, unclearedPackages, problemPackages);
+    await CompareResultsWithApprovedLicenses(unclearedPackages, problemPackages);
+    const grepResults = await PerformGrep(packageDirectory);
+    await CompareResultsWithGPLMentions(grepResults, unclearedPackages, problemPackages, packageId);
+
+    const report = {};
```
jcockbain96:
```suggestion
    const report = { packageId, acceptedPackages, unclearedPackages, problemPackages };
```

test/src/license.test.js
```diff
@@ -0,0 +1,52 @@
...
+                } else {
+                    mlog.error('WARN - Contains packages that have not yet been approved:');
+                    mlog.log(JSON.stringify(results.unclearedPackages, null, 2));
+                }
+        
+                if (Object.keys(results.problemPackages).length === 0) {
+                    mlog.success('Contains no problematic packages');
+                } else {
+                    mlog.error('ERR - Could contain problematic packages');
+                    mlog.log(JSON.stringify(results.problemPackages, null, 2));
```
jcockbain96: whitespace on the end of this line

ebuckle:
Cheers James, I'll get that tidied up

test/modules/license.service.js
```diff
@@ -0,0 +1,265 @@
...
+                delete packageJson.dependencies[dependency];
+            }
+        }
+    }
+}
+
+async function NodePackageInstall(packageJson, packageDirectory, packageLockJson) {
+    let npmInstallCommand;
+    const packageJsonPath = path.join(packageDirectory, 'package.json');
+    let packageLockJsonPath;
```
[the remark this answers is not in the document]
ebuckle: If I initialize these as const then I wouldn't be able to set the value in the following if/else block. 

If I initialized as const in the if/else then they would be local variables in the block and would be unreachable by the await exec() command at the end of the function.

test/src/license.test.js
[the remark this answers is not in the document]
ebuckle: Was leftover from a previous version of the file, removed.

ebuckle:
Changes made from your review

ebuckle:
Yep, I see. Change made

PR #1088, opened 2019-11-12
Project.js tests + enable code coverage

james-wallis:
#### Project.js
* Add tests (Apart from the socket functions)
* Reorder two functions to move them out of the constructor
* Remove functions that are not used anywhere in the codebase
  * Tests still pass + no mentions anywhere

#### Enable test code coverage
* Add NYC which works well with Mocha to report code coverage 
  * Reports in terminal after testing
  * Creates HTML page to open in browser
* Keeps old coverage generator to be removed later rather than disrupt current way of testing 

#### Additional
* Remove `repository_list.json` as it shouldn't have been committed in the first place (deleted by tests)

### DIFF
--- a/test/src/unit/modules/Project.test.js
+++ b/test/src/unit/modules/Project.test.js
@@ -525,4 +525,18 @@ describe('Project.js', () => {
             descPostChange.should.equal(description);
         });
     });
+    describe('getComparison()', () => {
+        it('gets comparison', async() => {
+            const project = new Project({ name: 'dummy' }, global.codewind.CODEWIND_WORKSPACE);
+            project.loadTestPath = loadTestResources;
+            const comparision = await project.getComparison();
+            comparision.should.have.length(4);
+            comparision.should.containSubset([
+                { type: 'cpu', delta: {} },
+                { type: 'gc', delta: {} },
+                { type: 'memory', delta: {} },
+                { type: 'http', delta: {} },
+            ]);
+        });
+    });
 });
--- a/test/src/unit/modules/Project.test.js
+++ b/test/src/unit/modules/Project.test.js
@@ -18,6 +18,8 @@ const chaiAsPromised = require('chai-as-promised');
 
 const Project = rewire('../../../../src/pfe/portal/modules/Project');
 const ProjectError = require('../../../../src/pfe/portal/modules/utils/errors/ProjectError');
+const ProjectMetricsError = require('../../../../src/pfe/portal/modules/utils/errors/ProjectMetricsError');
+
 const { suppressLogOutput } = require('../../../modules/log.service');
 
 chai.use(chaiSubset);
@@ -33,9 +35,9 @@ describe('Project.js', () => {
         global.codewind = { 
             RUNNING_IN_K8S: false,  
             CODEWIND_WORKSPACE: `${__dirname}/project_temp/`, 
-            CODEWIND_TEMP_WORKSPACE: '${__dirname}/project_temp/temp/',
+            CODEWIND_TEMP_WORKSPACE: `${__dirname}/project_temp/temp/`,
         };
-        fs.ensureDirSync(global.codewind.CODEWIND_WORKSPACE);
+        fs.ensureDirSync(global.codewind.CODEWIND_TEMP_WORKSPACE);
     });
     after(() => {
         fs.removeSync(global.codewind.CODEWIND_WORKSPACE);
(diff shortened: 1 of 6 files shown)


test/src/unit/modules/Project.test.js
```diff
@@ -95,9 +108,82 @@ describe('Project.js', () => {
...
+                project.should.be.an('object');
+                const packageSwiftPath = path.join(project.projectPath(), 'Package.swift');
+                await fs.ensureFile(packageSwiftPath);
+                await fs.writeFile(packageSwiftPath, 'SwiftMetrics.git');
+                const areMetricsAvailable = await project.checkIfMetricsAvailable();
+                areMetricsAvailable.should.be.true;
+            });
+        });
+        describe('Checks if metrics are available for Appsody projects', () => {
+            it('Node.js', async() => {
```
jcockbain96: Would prefer these to be whole sentences, even if some message is repeated

test/src/unit/modules/Project.test.js
```diff
@@ -95,9 +108,82 @@ describe('Project.js', () => {
...
+                const project = new Project({ name: 'dummy', language: 'java' }, global.codewind.CODEWIND_WORKSPACE);
+                project.should.be.an('object');
+                const pomXmlPath = path.join(project.projectPath(), 'pom.xml');
+                await fs.ensureFile(pomXmlPath);
+                await fs.writeFile(pomXmlPath, 'javametrics');
+                const areMetricsAvailable = await project.checkIfMetricsAvailable();
+                areMetricsAvailable.should.be.true;
+            });
+            it('Swift', async() => {
+                const project = new Project({ name: 'dummy', language: 'swift' }, global.codewind.CODEWIND_WORKSPACE);
```
jcockbain96: You could combine these 2 lines into a `createProjectAndCheckItsAnObject(projectParams)`
function, and call that at the start of each it block

jcockbain96: Would save 1/2 lines in lots of places

test/src/unit/modules/Project.test.js
```diff
@@ -116,32 +202,536 @@ describe('Project.js', () => {
...
+            const project = new Project({ name: 'dummy', language: 'java' }, global.codewind.CODEWIND_WORKSPACE);
+            const metricsRoot = project.getMetricsContextRoot();
+            metricsRoot.should.equal('javametrics');
+        });
+        it('Gets metrics root for a Swift project', () => {
+            const project = new Project({ name: 'dummy', language: 'swift' }, global.codewind.CODEWIND_WORKSPACE);
+            const metricsRoot = project.getMetricsContextRoot();
+            metricsRoot.should.equal('swiftmetrics');
+        });
+        it('Gets a blank string root for an invalid Project', () => {
```
jcockbain96: blank string metrics root?

jcockbain96: Also lower case project

test/src/unit/modules/Project.test.js
```diff
@@ -116,32 +202,536 @@ describe('Project.js', () => {
...
+            const metricsRoot = project.getMetricsContextRoot();
+            metricsRoot.should.equal('');
+        });
+        it('Gets a blank string root for a Project with no language', () => {
+            const project = new Project({ name: 'dummy' }, global.codewind.CODEWIND_WORKSPACE);
+            const metricsRoot = project.getMetricsContextRoot();
+            metricsRoot.should.equal('');
+        });
+    });
+    describe('projectPath(_)', () => {
```
jcockbain96: underscore in params here?

test/src/unit/modules/Project.test.js
```diff
@@ -116,32 +202,536 @@ describe('Project.js', () => {
...
+                .should.be.eventually.rejectedWith(`Unable to find metrics for project ${project.projectID}`)
+                .and.be.an.instanceOf(ProjectMetricsError)
+                .and.have.property('code', 'NOT_FOUND');
+        });
+        it('Gets a metrics file using a valid time stamp (String)', async() => {
+            const project = new Project({ name: 'dummy' }, global.codewind.CODEWIND_WORKSPACE);
+            project.loadTestPath = loadTestResources;
+            const metricsFile = '20190326154749';
+            const metricsTypes = await project.getMetricsByTime(metricsFile);
+            metricsTypes.should.have.property('id').and.equal(0);
```
jcockbain96: this could be array of properties which you loop through, to save some duplication

test/src/unit/modules/Project.test.js
james-wallis: Thats how the function is in Project.js

Issue #257, opened 2019-08-22
SVT:Project Build rank info displayed with wrong data.


jagraj:
**Codewind version:** 0.3
**OS:** WIndows 10

**Che version:**
**IDE extension version:** latest (0.3.0.v201908161735)
**IDE version:** 
Eclipse IDE for Enterprise Java Developers.
Version: 2019-06 (4.12.0)
Build id: 20190614-1200
OS: Windows 10, v.10.0, x86_64 / win32
Java version: 1.8.0_201

**Kubernetes cluster:**

**Description:** 
Projects build Rank displaying wrong data in plugins.  It says [Build queued: Rank 12/12] ,  [Build queued: Rank 13/13]

Here is the screenshot..
![image](https://user-images.githubusercontent.com/713849/63520542-e1a69100-c4c2-11e9-831e-4120fe3c7f6a.png)

**Steps to reproduce:**
1. Install codewind from the plugins.
2. Create all project types.
3. Builds get into queue and displays incorrect information.

**Workaround:**

ssh24:
I am also able to reproduce this. I had these projects created and then rebuilt codewind and enabled them but the ranks don't seem correct
![Uploading Screen Shot 2019-11-11 at 6.26.52 PM.png…]()

ssh24:
After debugging this and I found out that Turbine is emitting the event correctly and it is the IDE that does not reflect the new ranks. I was able to see the new rank by clicking on `Local` -> `Refresh` and the new rank would show correctly on the IDE. Below is the logs from Turbine where you can see the ranks are emitted correctly from us:
```
[12/11/19 12:34:40 Turbine] [TRACE] Key translated. Returning message: Rank 1/2
[12/11/19 12:34:40 Turbine] [TRACE] Key translated. Returning message: Rank 2/2
[12/11/19 12:34:40 Turbine] [DEBUG] Running Builds queue: ["39be6fc0-00af-11ea-8d51-9982ecc43880","e1367560-04d1-11ea-93c7-877e1ecd4578"]
[12/11/19 12:34:40 Turbine] [DEBUG] Builds in progress: 2
[12/11/19 12:34:40 mynod] [TRACE] Setting rank for projectID d58d1390-04d1-11ea-93c7-877e1ecd4578: 1/2
[12/11/19 12:34:40 myspr] [INFO] Build completed for 4bb21420-00af-11ea-8d51-9982ecc43880
[...]

PR #1091, opened 2019-11-12
Fix update build status with correct detailed message to emit

ssh24:
Related to https://github.com/eclipse/codewind/issues/257

The update build status function was using the wrong detailed message and hence was not able to emit on changes of the detailed message, for example: project rank changes. This PR uses the correct detailed message to emit the events and store the build state map.

You can watch the screen recording here on how it is done now:
[Screen Recording 2019-11-12 at 11.57.52 AM.zip](https://github.com/eclipse/codewind/files/3837308/Screen.Recording.2019-11-12.at.11.57.52.AM.zip)

### DIFF
--- a/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts
+++ b/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts
@@ -262,7 +262,7 @@ export async function updateProjectStatus(type: string, projectID: string, statu
             if (buildImageLastBuild) {
                 data.buildImageLastBuild = buildImageLastBuild;
             }
-            buildStateMap.set(projectID, new ProjectState(newState, newDetailedState, timestamp, appImageLastBuild, buildImageLastBuild));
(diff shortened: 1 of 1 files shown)


ssh24:
@elsony Could you please help me review this PR?

rajivnathan:
Adding some additional explanation from Sakib since I didn't understand how this change is meant to fix projects reporting ranks of 12/12 or 13/13, etc. simultaneously.

Paraphrasing from discussion with Sakib:
This change will ensure that the build state map is always up-to-date with what's been passed to the updateProjectStatus() function.

In https://github.com/eclipse/codewind/blob/6f1df458f5da74d87709ea12c80970172b3d89c4/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts#L250-L252 we're passing in already translated messages originating from the emitProjectRanks() function here: https://github.com/eclipse/codewind/blob/9f4812f1ea7babd6861a96e73a52ca0a28fee335/src/pfe/file-watcher/server/src/controllers/projectsController.ts#L540

The translated messages previously never made it to the build state map and so the map would contain outdated values that would cause the comparison between newDetailedState and oldDetailedState to match and prevent the status from being sent.

PR #1136, opened 2019-11-14
0.6.0: 777 on project

### LINKED ISSUE #1133: Project folder in PFE not world writable
**Codewind version:** 0.6.0
**OS:**

**Che version:** 7.3.1
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:** OCP 4.2

**Description:**

This would address issue https://github.com/kabanero-io/kabanero-foundation/issues/129, where the Kabanero springboot stack is failing to run in Codewind on Che.  We narrowed it down to a permissions problem:

- kabanero stacks run as non-root (they run as `java_user`)
- when project is created in PFE, it is owned by root and has 775 permissions (user/group write only)
- when the project dir is mounted to the app container (as `/project/user-app`), that folder has the same ownership and permissions as in PFE
- therefore, `java_user` cannot write to `/project/user-app`

Tested that if the project dir is made world writable in PFE, then the project can run

**Steps to reproduce:**

**Workaround:**

makandre:

Fixes #1133

### DIFF
--- a/src/pfe/portal/modules/utils/sharedFunctions.js
+++ b/src/pfe/portal/modules/utils/sharedFunctions.js
@@ -109,6 +109,7 @@ module.exports.updateObject = function updateObject(objectToUpdate, fieldsToAddT
 module.exports.copyProject = async function copyFile(fromProjectPath, toProjectPath) {
   log.debug(`copyProject fromPath: ${fromProjectPath}, toPath: ${toProjectPath}`);
   await fs.copy(fromProjectPath, toProjectPath);
+  await fs.chmod(toProjectPath, "777");
   // now clean up old directory
   await module.exports.forceRemove(fromProjectPath)
 }
--- a/src/pfe/portal/modules/utils/sharedFunctions.js
+++ b/src/pfe/portal/modules/utils/sharedFunctions.js
@@ -106,10 +106,11 @@ module.exports.updateObject = function updateObject(objectToUpdate, fieldsToAddT
   return objectToUpdate;
 }
 
-module.exports.copyProject = async function copyFile(fromProjectPath, toProjectPath) {
+module.exports.copyProject = async function copyFile(fromProjectPath, toProjectPath, mode) {
   log.debug(`copyProject fromPath: ${fromProjectPath}, toPath: ${toProjectPath}`);
   await fs.copy(fromProjectPath, toProjectPath);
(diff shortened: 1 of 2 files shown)


makandre:
This was tested on Che and does resolve the issue for the kabanero springboot stack

maysunfaisal:
Do we want to do this for all projects tho?

makandre:
@maysunfaisal gd point, I can scope it down

BarDweller:
`chmod 777` is like using a sledgehammer to crack a nut here! 

Have you considered the impact to the application? If you 777 the entire dir, you just altered all the users file permissions, blindly adding execute/read/write bits for user/group/world .. what if the user had project content that relied on the correct permission bits being set, you just obliterated them. Eg, read-only config just became _world_ writable, that's horrific!! Even in a container environment!

BarDweller:
> Do we want to do this for all projects tho?

Given the issue is related to the use of APPSODY_USER_RUN_AS_LOCAL, it would make sense to limit fixes to any stack encountered using that flag with the value `true`

makandre:
This only changes the project dir permission inside PFE, which had 775 permissions, so in effect it only adds the world-write bit.  It is also not a recursive set.

BarDweller:
>This only changes the project dir permission inside PFE, which had 775 permissions, so in effect it only adds the world-write bit. It is also not a recursive set.

So this fix only works because the only write the current stack & sample template makes are to _new_ files, directly within the project dir. (eg, creation project/target). 

If during the process of a maven build, the steps of which are entirely in the users hands, project using the stack were to try to modify any existing file, it would still break, because the file would be owned by the wrong user. (Eg if a maven build wanted to update a build counter stored in a text file within the project, or wanted to copy or create resources within the project itself, rather than within the target folder). 

Given you know the user id you are executing the stack as, why can't you chown rather than chmod ?

makandre:
As we're trying to shutdown 0.6.0, you could say this is a tactical fix.

True, a more exhaustive fix may be to chown the folder.  However, Codewind does not actually know the user id (unless we get into the business of inspecting the stack image I suppose, but I don't think that something Codewind should be doing)

@chilanti Is there anything the appsody CLI could do here to chown the project?

makandre:
@tobespc @james-wallis Could you have a look at this PR?  It's a quick fix for #1133

Issue #1072, opened 2019-11-08
Docker cp of mc-target is to a wrong folder in windows


ssh24:
**Description:**
Similar to https://github.com/eclipse/codewind/issues/1071, the liberty path in windows should be `/tmp` instead of `/home/default/app/mc-target`
```
[08/11/19 20:48:47 saklib] [TRACE] Docker copy command to copy the mc-target folder: docker cp cw-saklib-a3d2ffb0-0268-11ea-bc65-e5ab01f015cd:/home/default/app/mc-target /codewind-workspace/saklib

[08/11/19 20:48:47 saklib] [ERROR] executeBuildScript : 404 | Error: No such container:path: cw-saklib-a3d2ffb0-0268-11ea-bc65-e5ab01f015cd:/home/default/app/mc-target

[08/11/19 20:48:47 saklib] [ERROR] executeBuildScript : 404 | Failed to copy mc-target folder to the PFE container
```

ssh24:
So it does look like the `docker cp` back to PFE was added initially in the hybrid stages to have a snapshot of the logs working. It can safely be removed. Verified it works after removing.

ssh24:
This is fixed now. **Verified**

PR #1151, opened 2019-11-15
Remove docker cp of mc-target to PFE container

ssh24:
### Description

So it does look like the `docker cp` back to PFE was added initially in the hybrid stages to have a snapshot of the logs such as console, message log available. It can safely be removed. Verified it works after removing.

Related to https://github.com/eclipse/codewind/issues/1072


ssh24:
@rajivnathan Could you PTAL?

Issue #1180, opened 2019-11-18
Gatekeeper server won't start


tetchel:
after `cwctl install remote`,
```
[ /tim/programs/codewind ] 07 (master) $ kw po
kubectl get po @ 15:07:34
NAME                                             READY   STATUS             RESTARTS   AGE
codewind-gatekeeper-k34uy7if-86d688b784-xc4mv    0/1     CrashLoopBackOff   4          2m10s
codewind-keycloak-k34uy7if-94b4d8d5c-gg6c5       1/1     Running            0          2m56s
codewind-performance-k34uy7if-6cf566f659-f7p7h   1/1     Running            0          2m10s
codewind-pfe-k34uy7if-6d64b969d9-9s4tj           1/1     Running            0          2m11s
^C
[ /tim/programs/codewind ] 07 (master) $ k logs -f po/codewind-gatekeeper-k34uy7if-86d688b784-xc4mv

> gatekeeper@1.0.0 start /usr/src/app
> node server.js

internal/modules/cjs/loader.js:797
    throw err;
    ^

Error: Cannot find module 'express'
Require stack:
- /usr/src/app/server.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:794:15)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:849:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/usr/src/app/server.js:1:17)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1025:10) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/usr/src/app/server.js' ]
}
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! gatekeeper@1.0.0 start: `node server.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the gatekeeper@1.0.0 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?

npm ERR! A complete log of this run can be found in:
[...]

PR #1185, opened 2019-11-18
Add npm install step to gatekeeper

tetchel:
https://github.com/eclipse/codewind/issues/1180

Signed-off-by: Tim Etchells <timetchells@ibm.com>

the container can run successfully now with this change

### DIFF
--- a/src/gatekeeper/Dockerfile
+++ b/src/gatekeeper/Dockerfile
@@ -18,6 +18,7 @@ RUN apk add --no-cache --update openssl
 # Create UI app directory and install source code
 WORKDIR /usr/src/app
 COPY . .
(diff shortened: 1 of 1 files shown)


tetchel:
@markcor11

tobespc:
@tetchel Thanks for investigating and putting this PR up but, after discussion we are going to fix this in a slightly different way.  @markcor11 will put a fix in shortly

markcor11:
Lets not install the dependancies this way... we should use the package-lock to get the exact versions of the modules and that needs to be .git ignored to bypass the defaults.  I've put up a PR with those additions

tetchel:
that is a better way, thats why i opened #1186 :+1:

PR #1200, opened 2019-11-19
fix failing templates test

tobespc:

The repos needed updating to the latest versions to be used


rajivnathan:
I believe the name of the skaffold template changed to `Kabanero Node.js Express scaffold template`?

PR #1220, opened 2019-11-20
Update docs to include bind/upload

stalleyj:
I've added project/bind/start, project/bind/end, project/upload and project/upload/end. 
Also removed some leftover initialize doc.

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -197,14 +197,14 @@ paths:
         500:
           description: Internal error
 
-  /api/v1/projects/bind:
+  /api/v1/projects/bind/start:
     post:
-      summary: Bind a project to Codewind
+      summary: Begin binding a project to Codewind
       description: >
-        The bind api will be passed the location on disk of an existing project so that Codewind
-        adds it to the list of projects it monitors.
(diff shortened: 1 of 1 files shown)


docs/openapi.yml
```diff
@@ -245,6 +246,80 @@ paths:
...
+          schema:
+            $ref: '#/components/schemas/ProjectID'
+          required: true
+          description: id of the project        
+      responses:
+        202:
+          description: >
+            Bind end request accepted. Will attempt to complete binding the Codewind project.
+        404:
+          description: Project with given projectiD was not found.
```
hhellyer: `projectiD` - I think you've managed to find a formatting of projectId we haven't used before. ;-)

docs/openapi.yml
```diff
@@ -1445,43 +1520,45 @@ components:
...
+          type: array
+          example: ["/project/modifiedfile1"]
+      
+    Upload:
       type: object
       required:
-        - status
-        - projectPath
+        - path
+        - msg
```
hhellyer: We do have an optional `mode` parameter now so we can get the file permissions right as well as the path and the contents.

Issue #1161, opened 2019-11-15
Turbine build queue may not be working properly


rajivnathan:
**Codewind version:** 0.6.0
**OS:** Mac OSX

**Che version:** 7.3.1
**IDE extension version:** 0.6.0
**IDE version:** 7.3.1
**Kubernetes cluster:** OKD 3.1.1

**Description:**
I stopped and started a Che workspace that had many projects in it (17 in this case). The view in Theia shows 3 builds in progress and the rest in the queue which is what we'd expect but when I check the PFE container there are many build processes running. 3 were running the `create` command on the build script and the rest were running `update`.

If there are only 3 builds running I'd expect only 3 processes to be running but clearly something is triggering more build processes.

Also note that the status shows `Running` and `Build Queued` at the same time?

**Codewind view in Theia**

![image](https://user-images.githubusercontent.com/20015929/68972401-b8dc2200-07b9-11ea-9796-e48c46cc578e.png)

**PFE logs**
[pfe.log](https://github.com/eclipse/codewind/files/3853006/pfe.log)

**Output of `ps aux` command**
```
USER        PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root          1  0.0  0.0  11684  1468 ?        Ss   19:34   0:00 sh -c  /file-watcher/scripts/root-watcher.sh ${HOST_WORKSPACE_DIRECTORY} ${CONTAINER_WORKSPACE_DIRECTORY} ; cd /portal ; npm start ;
root        104  0.1  0.1 668728 30508 ?        Sl   19:34   0:00 npm
root        119  0.0  0.0  11828  1500 ?        S    19:34   0:00 /bin/bash ./npm-start.sh
root        120  0.0  0.1 668648 30820 ?        Sl   19:34   0:00 npm
root        134  3.9  0.5 954528 96944 ?        Sl   19:34   0:23 node server.js
root        186  0.0  0.0  12060  1908 ?        Ss   19:34   0:00 /bin/bash /file-watcher/scripts/spring-container.sh /codewind-workspace/spring6 /codewind-workspace 4c3f5550-0718-11ea-a90f-55d080717017 create cw-spring6-4c3f5550-0718-11ea-a90f true cw-4c3f5550-0718-11ea-a90f-55d080717017-65f226471c4ae run 7777 NONE spring6-4c3f5550-0718-11ea-a90f-55d080717017 docker.io/jeevandroid
[...]

PR #1231, opened 2019-11-20
Prevent build running while project is queued

rajivnathan:
Related to #1161 this fixes the problem where an update build is running while the create is still scheduled on the build queue.
Signed-off-by: Rajiv Senthilnathan <rajivsen@ca.ibm.com>


yangcao77:
LGTM

Issue #1246, opened 2019-11-22
App status flipping between stopping stopped and starting during update


rajivnathan:
**Codewind version:** latest
**OS:** any

**Description:**
```
[34m[21/11/19 23:36:23 server.js] [TRACE][39m POST /internal/api/v1/projects/updateStatus called with 
 req.body { projectID: 'f5324df0-0cb5-11ea-9ec8-b78f8be2e20d',
  status: 'stopping',
  type: 'appState' }
[36m[21/11/19 23:36:23 /portal/routes/projects/internal.route.js] [DEBUG][39m Internal API: updateStatus
[32m[21/11/19 23:36:23 User.js] [INFO][39m Updating status for project f5324df0-0cb5-11ea-9ec8-b78f8be2e20d.
[36m[21/11/19 23:36:23 FileWatcher.js] [DEBUG][39m Portal's event listener: projectStatusChanged
[36m[21/11/19 23:36:23 FileWatcher.js] [DEBUG][39m Portal's event listener: {"projectID":"f5324df0-0cb5-11ea-9ec8-b78f8be2e20d","appStatus":"stopping"}
[36m[21/11/19 23:36:23 FileWatcher.js] [DEBUG][39m projectStatusChanged: stopping (project f5324df0-0cb5-11ea-9ec8-b78f8be2e20d)
[34m[21/11/19 23:36:23 FileWatcher.js] [TRACE][39m projectStatusChanged: stopping (project: {
  "projectID": "f5324df0-0cb5-11ea-9ec8-b78f8be2e20d",
  "appStatus": "stopping"
})
[36m[21/11/19 23:36:23 FileWatcher.js] [DEBUG][39m Received filewatcher module response: 
[36m[21/11/19 23:36:23 FileWatcher.js] [DEBUG][39m {"statusCode":200}
[34m[21/11/19 23:36:23 server.js] [TRACE][39m responded to POST /internal/api/v1/projects/updateStatus with status 200 and body { projectID: 'f5324df0-0cb5-11ea-9ec8-b78f8be2e20d',
  status: 'stopping',
  type: 'appState' }
[34m[21/11/19 23:36:23 server.js] [TRACE][39m Checking origin host  matches expected host localhost:9090
[34m[21/11/19 23:36:23 server.js] [TRACE][39m POST /internal/api/v1/projects/updateStatus called with 
 req.body { projectID: 'f5324df0-0cb5-11ea-9ec8-b78f8be2e20d',
  status: 'stopped',
[...]

PR #1255, opened 2019-11-22
Revert back change to the pinging timeout

yangcao77:

This PR is for issue: https://github.com/eclipse/codewind/issues/1246
Reverting back the change to the pinging timeout. It may take longer than 1 min to start the application if it's on a slow cluster, or if the project is big. 
With the hardcoded 1 min pinging timeout, user would see the status flips from `starting` -> `stopped` -> `started`, which is confusing. 
Remove the timeout in 0.6.0, and will add warning message instead post GA.


yangcao77:
@rajivnathan Can you help review this PR?

Issue #1149, opened 2019-11-15
Way to copy portal source into PFE for debugging (for devs)


makandre:
**Codewind version:** 0.7.0
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description of the enhancement:**

Recently, the `start.sh` script used by developers was changed to use `cwctl` to start up codewind https://github.com/eclipse/codewind/pull/1085

However, now it seems that it's no longer possible to add the portal source folder as a mount into PFE; that ability was very useful for developers as changes made to the source would trigger `nodemon` to restart the process and pick up the changes, without needing to rebuild and restart.

**Proposed solution:** 

A way to pass in the location of portal source folder (possibly only when `--debug` is used) and mount it to PFE when starting it

makandre:
Instead of mount, @rajivnathan suggested to copy files into PFE instead.  That's what's done in #1288

PR #1275, opened 2019-11-26
allow debug mount of portal source

makandre:

Resolves https://github.com/eclipse/codewind/issues/1149
Requires PR https://github.com/eclipse/codewind-installer/pull/240

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -45,17 +45,6 @@ rm $GIT_CONFIG
 git config -f $GIT_CONFIG --add user.name "`git config --get user.name || echo 'codewind user'`"
 git config -f $GIT_CONFIG --add user.email "`git config --get user.email || echo 'codewind.user@localhost'`"
 
-
-# Set docker-compose file
-DOCKER_COMPOSE_FILE="docker-compose.yaml -f docker-compose-remote.yaml"
-
-
-if [ "$DEVMODE" = true ]; then
-  printf "\nDev mode is enabled\n";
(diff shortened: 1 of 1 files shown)


makandre:
@tobespc @rajivnathan Could you please take a look at this PR?

rajivnathan:
Do we really need a mount for dev? Part of the reason for making the switch to cwctl is to bring the dev environment closer to the prod environment. The main difference between dev and prod currently is nodemon and other dev dependencies that stick around. Adding a mount to the container sounds like a significant departure from that.

Also, you don't need to rebuild and restart because we already have nodemon running in dev mode. You'd only need to copy the source into the container for nodemon to pick it up. For Turbine development we have `push` scripts to handle that. See https://github.com/eclipse/codewind/blob/755a13a4aa067e9b2790ffc3c87943e0ae0fffd5/src/pfe/file-watcher/server/package.json#L22

Think we just need something similar for portal?

makandre:
If you’re copying the entire dist folder over, how is it that much different than mounting the folder? (Tho I didn’t know about this push script in turbine, that’s useful)

Well I did make this available only if —dev is specified, so it’s not going to do the mount by default. I guess a mount is just a convenience (tho especially when working on extension code since I can just save a file to trigger the restart and reload the extension)

But ok, if turbine is doing push local style then we should keep it consistent. I’ll change it to that.

makandre:
Superseded by #1288

Issue #669, opened 2019-10-11
Enhance the Turbine Che test scripts


ssh24:
### Description

We need to enhance the turbine che test scripts by utilizing chectl and by utilizing the che api's where possible. It can be then used for the turbine functional test on kube and the sidecar tests.

ssh24:
@rajivnathan I am looking into this. Can you please assign me?

PR #1278, opened 2019-11-26
Che setup script

ssh24:
### Description

Related to https://github.com/eclipse/codewind/issues/669

This che setup script can be used to deploy a che instance and also install codewind. The inital requirement was for our test automation but can also be used for dev purposes. The script is currently stored under `/file-watcher/server/test/scripts/` but can be moved around as it can run standalone.

Example script run:
```
./che-setup.sh --cluster-ip=<cluster_ip> --cluster-user=<cluster_user>  --cluster-pass=<cluster_pass>  --che-ns=che --che-version=stable --clean-deploy --default-registry --install-codewind
```

**PS.** The setup is only for OpenShift kube. Further enhancement can be made in the future for multi-kube support.

### DIFF
--- a/src/pfe/file-watcher/server/test/scripts/.gitignore
+++ b/src/pfe/file-watcher/server/test/scripts/.gitignore
@@ -1,0 +1,3 @@
+codewind-che-plugin
+codewind-odo-extension
+chectl-install.log
--- a/src/pfe/file-watcher/server/test/scripts/che-setup.sh
+++ b/src/pfe/file-watcher/server/test/scripts/che-setup.sh
@@ -21,6 +21,8 @@ RESET='\033[0m'
 CODEWIND_CHE="codewind-che-plugin"
 CODEWIND_ODO_EXTENSION="codewind-odo-extension"
 
+CURR_DIR=$(pwd)
+
 CLUSTER_IP=
 CLUSTER_USER=
 CLUSTER_PASS=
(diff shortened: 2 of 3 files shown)


ssh24:
@maysunfaisal Could you PTAL?

ssh24:
I don't think it matters. But to be fair, I like it on the codewind repo because it is easy to just deploy che from one place instead of cloning the codewind-che-plugin repo. It is a script for improving developer experience to deploy a che automatically

ssh24:
I was thinking of putting the script in https://github.com/eclipse/codewind/tree/master/script instead so that it everyone can access it instead of going under `/file-watcher/server/test/scripts/`

maysunfaisal:
@ssh24 hmm but i think it makes more sense to have it in the other repo. Because i think we should make more ppl be aware of that repo and is specific to Che. 

@rajivnathan any thoughts?

ssh24:
Sure, I don't mind moving it to the other repo. But let's get everyone's opinion in the devstat.

rajivnathan:
The Codewind install docs for Che ask users to clone the codewind-che-plugin repo so it probably makes sense to have the install script there? If we really want a way to install it easily from this repo we could have a separate script for this repo that pulls down the latest version of the install script and installs?

This script appears to reference the other repo `OPERATOR_YAML="$CURR_DIR/$CODEWIND_CHE` already so does that mean the other repo would need to be cloned anyway?

ssh24:
Yeah, it requires the codewind-che-plugin repo for the default cluster yaml. Sure, I can move the script over there

Issue #1270, opened 2019-11-25
Update and run the Kube functional tests on OCP4


johnmcollier:
We're currently running our nightly Turbine and Codewind Che functional tests for Kube on OKD 3.11. We need to get them running on OCP4.

Before we can run the tests on OCP4, we need to update the test scripts to remove references to `nip.io` and `CLUSTER_IP` (see https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/server/test/scripts/setup.sh#L48) when accessing the Che / Keycloak routes, both of which won't work on OCP4. It's probably easiest to dynamically retrieve the Che and Keycloak URLs and then pass them into the test scripts:
```
CHE_URL=$(kubectl get routes --selector=component=che -o jsonpath="{.items[0].spec.host}")
KEYCLOAK_URL=$(kubectl get routes --selector=component=keycloak -o jsonpath="{.items[0].spec.host}")
```

ToDo:
- [ ] Update test scripts to dynamically retrieve Che and Keycloak routes where needed
- [ ] Provision an internal OCP4 test cluster
   - 3 Masters, 3 Workers should suffice
- [ ] Run the nightly tests on the OCP4 cluster

jgwest:
@maysunfaisal: GitHub didn't allow me to assign the following users: ssh24

Note that only org members, repo collaborators and people who have commented on this issue/PR can be assigned.

ssh24:
Our tests are now running on OKD4. Closing this as done.

/close

PR #1283, opened 2019-11-26
Fix test script to dynamically get che endpoint

ssh24:
Related to https://github.com/eclipse/codewind/issues/1270

Tested with @maysunfaisal and it works


ssh24:
@maysunfaisal PTAL

PR #1288, opened 2019-11-27
pushLocal.sh script

### LINKED ISSUE #1149: Way to copy portal source into PFE for debugging (for devs)

makandre:

Resolves #1149 
This PR supersedes #1275 

Adding a `pushLocal` script as suggested by @rajivnathan 

@tobespc @rajivnathan pls review, thanks.

### DIFF
--- a/src/pfe/portal/package.json
+++ b/src/pfe/portal/package.json
@@ -10,7 +10,7 @@
     "coverage": "nyc node server.js",
     "local": "node server.js --local",
     "eslint": "eslint .",
-    "pushLocal": "./pushLocal.sh"
+    "pushLocal": "npm run eslint && ./pushLocal.sh"
   },
   "author": "",
   "license": "ISC",
--- a/src/pfe/portal/package.json
+++ b/src/pfe/portal/package.json
@@ -9,7 +9,8 @@
     "dev": "nodemon --inspect=0.0.0.0:9777 server.js",
     "coverage": "nyc node server.js",
     "local": "node server.js --local",
-    "eslint": "eslint ."
+    "eslint": "eslint .",
+    "pushLocal": "./pushLocal.sh"
   },
   "author": "",
   "license": "ISC",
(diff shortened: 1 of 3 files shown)


src/pfe/portal/package.json
```diff
@@ -9,7 +9,8 @@
     "dev": "nodemon --inspect=0.0.0.0:9777 server.js",
     "coverage": "nyc node server.js",
     "local": "node server.js --local",
-    "eslint": "eslint ."
+    "eslint": "eslint .",
+    "pushLocal": "./pushLocal.sh"
```
rajivnathan: You may want to run eslint as part of the pushLocal

johnmcollier:
Wondering if we should move the script one level up into the PFE folder rather than storing it directly as part of the Portal container's code?

makandre:
added to `.dockerignore`

src/pfe/portal/pushLocal.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+docker cp config/. codewind-pfe:/portal/config
+docker cp middleware/. codewind-pfe:/portal/middleware
+docker cp modules/. codewind-pfe:/portal/modules
+docker cp routes/. codewind-pfe:/portal/routes
+docker cp server.js codewind-pfe:/portal/server.js
```
hhellyer: Do you need to add the `openapi.yml` file here? It's copied in when we do a build and is actually used by the `validateReq` middleware function?

makandre:
@hhellyer Ok added the docs folder and will mention this script in the next playback

Issue #665, opened 2019-10-11
Create docker registry secret in Codewind


tetchel:
The user should not have to do step 1) of https://github.com/tetchel/codewind-without-che#install-steps

We should create the secret for the user when they set their docker registry credentials.

This could be done in PFE upon POST /registry.

This could also be done using a cwctl command, if the cwctl had access to the current cluster context (which it should).

tobespc:
The docs in that link are not valid anymore with the work Mark has done, rather than go through tech topics lets just have a chat when you get in to sort this out

jopit:
TODO: iterative-dev to investigate if this could be done in PFE upon POST /registry.

maysunfaisal:
I can take a look at this.

Even tho this is tagged as iterative-dev; this issue will need new sets of Portal API for setting up docker registries.

I have already laid the foundation of the new API in my branch, will update this issue with API info tomorrow.

maysunfaisal:
cc: @tetchel @jopit

maysunfaisal:
POST `/api/v1/dockerregisrty` will take `url`,`username`,`password` in it's body
GET `/api/v1/dockerregisrty` will return the list of created Docker registry - `url` and `username`
DELETE `/api/v1/dockerregisrty` will take `url` as the input and delete the docker registry from PFE

rajivnathan:
Talked to Maysun and Elson, given we're working on shutdown we'll defer this to the next tech-topic call.

maysunfaisal:
To use the docker registries on Kube/Che, we will need to enter information regarding our docker registries like we do in Che:

![Screen Shot 2019-11-06 at 7 02 10 PM](https://user-images.githubusercontent.com/31771087/68348485-f5c45c80-00c7-11ea-9ba6-5516cf3a1152.png)


## Existing Logic

### PFE Mount from Che secret

- When deploying PFE on Che, we mount the Che secret to `/tmp/secret` https://github.com/eclipse/codewind-che-plugin/blob/master/codewind-che-sidecar/deploy-pfe/pkg/codewind/util.go#L176
- The secrets are then copied over from the PFE mount to the appropriate place like either `/root/.docker/config.json` or `/root/.dockercfg` during PFE startup in `root-watcher.sh` https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/scripts/root-watcher.sh#L95

## Proposed APIs

The proposed APIs will do things similar(almost exactly) to what Che currently does

[...]

PR #1302, opened 2019-11-28
Update and apply PR feedback for /api/v1/registrysecrets

maysunfaisal:

PR review continuation from https://github.com/eclipse/codewind/pull/1118 for issue #665

### DIFF
--- a/src/pfe/portal/modules/User.js
+++ b/src/pfe/portal/modules/User.js
@@ -680,6 +680,14 @@ module.exports = class User {
     const registrySecretList = [];
     
     try {
+      // Handle dockerhub specifically on local since docker.io does not work
+      if (url === "docker.io" && !global.codewind.RUNNING_IN_K8S) {
+        // eslint-disable-next-line no-param-reassign
+        url = "https://index.docker.io/v1/";
+      } else if (!url.startsWith("https://") && !url.startsWith("http://")) {
+        // eslint-disable-next-line no-param-reassign
+        url = "https://" + url ;
+      }
       const isDockerConfigFilePresent = await cwUtils.fileExists(this.dockerConfigFile)
       const encodedAuth = Buffer.from(username + ":" + password).toString("base64");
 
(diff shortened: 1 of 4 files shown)


maysunfaisal:
@hhellyer pls help review, thx!

I will do a final test in my morning, so pls wait and do not merge :)

src/pfe/portal/modules/User.js
```diff
@@ -876,13 +852,8 @@ module.exports = class User {
...
       log.error(msg);
-      const retval = {
-        statusCode: 500,
-        body: msg
-      };
-      return retval;
+      throw err;
     }
     
     const retval = {
```
hhellyer: Again, this can just return the list. `GET /api/v1/registrysecrets` knows that's a 200 and it makes it easier to re-use this function elsewhere.

src/pfe/portal/modules/User.js
```diff
@@ -962,13 +924,9 @@ module.exports = class User {
       return retval;
```
hhellyer: Again, this can just return the list. `DELETE /api/v1/registrysecrets` knows that's a 200.

src/pfe/portal/modules/User.js
```diff
@@ -736,30 +725,18 @@ module.exports = class User {
...
-      };
-      return retval;
     }
 
-    if (isServiceAccountPatched) {
+    if (isServiceAccountPatched || !global.codewind.RUNNING_IN_K8S) {
+      // when either the service account is patched successfully or we are in local Codewind, we can return
       log.debug("Codewind Docker Registry List: " + JSON.stringify(registrySecretList));
       const retval = {
         statusCode: 201,
```
hhellyer: I think you can just return the list here. The `POST /api/v1/registrysecrets` route can decide that's a 201 response. That way any other part of pfe that wants the list of registry secrets can call this function and get a list. (And this function doesn't need to know anything about http response codes.)

src/pfe/portal/modules/User.js
maysunfaisal: updated

src/pfe/portal/modules/User.js
maysunfaisal: updated

maysunfaisal:
@hhellyer thx very much for the feedback. You're right, it looks cleaner and nicer. :D 

I have also put a 2nd commit where i am appending `https` to registry secret addresses if absent, very much like how Che does https://github.com/eclipse/che/blob/490413cd482f61878050accec0bf2d9ad094456d/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/ImagePullSecretProvisioner.java#L136

I have also handled conversion of `docker.io` to `https://index.docker.io/v1/` for local CW because `docker.io` does not work as an address on local but does work on kube because buildah can handle it. Issue comment here https://github.com/eclipse/codewind/issues/665#issuecomment-559233164 and i tested it manually and my commit solves the issue.

Thx again for reviewing!

maysunfaisal:
@hhellyer the Jenkins kick off has passed, pls merge when you have time, thx

Issue #1289, opened 2019-11-27
File rename is not handled correctly by sync


makandre:
**Codewind version:** master
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**

When a file is renamed, inside PFE the file with the new name is created along-side the file with the old name; the old file is not deleted.

**Steps to reproduce:**

1. rename a file in a project (I used VS code)
2. look inside the project inside `codewind-pfe:/codewind-workspace`, you'll see both old and new file there

**Workaround:**

tobespc:
interesting, the logic we have in there should prevent this happening

james-wallis:
Recreated using `cwctl` bind and then the VSCode extension to do the sync. Looks like when I delete files they're not being deleted either so I suspect its a problem with the deleting.

PR #1310, opened 2019-11-28
Fix file not being deleted on rename + other `sync/end` bugs

james-wallis:
Fixes https://github.com/eclipse/codewind/issues/1289 (File not being deleted on rename)
### Summary
* Fixed issue with file not being deleted after a rename
* Fixed unknown issue about file not being deleted when no file is changed / a new file is uploaded
  * If a file was not uploaded with sync then the `upload/end` API would return 404 without checking whether the modified list had any files inside thus skipped deleting files.
* Fixed unknown issue about file not being deleted when there is a space in the filename - changed sharedFunctions.forceRemove 

### Testing
* Added unit tests for functions introduced into `remoteBind.route.js` to sanity check them.
* Added unit tests for an existing function.
* Added unit tests for `sharedFunctions.forceRemove()`


src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -212,24 +212,23 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
...
       } else {
-
-        const currentFileList = await listFiles(pathToTempProj, '');
-
-        const filesToDeleteSet = new Set(currentFileList);
-        keepFileList.forEach((f) => filesToDeleteSet.delete(f));
-        const filesToDelete = Array.from(filesToDeleteSet);
+        const pathToProj = path.join(global.codewind.CODEWIND_WORKSPACE, project.name);
+        const currentFileList = await listFilesInDirectory(pathToProj);
+        // don't use pathToTempProj use the actual path ..... obv
```
hhellyer: The meaning of obv is obvious obviously...
(Tidy up the comment?)

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -334,6 +342,10 @@ async function listFiles(absolutePath, relativePath) {
   return fileList;
 }
 
+function listFilesInDirectory(absolutePathToDirectory) {
```
hhellyer: You could have achieved the same effect by setting a default value in listFiles:
```
async function listFiles(absolutePath, relativePath = '') {
```

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -212,24 +212,23 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
...
+        log.info('Temporary project directory doesn\'t exist and modified list is empty, not syncing any files');
+        res.status(404).send('No files have been synced');
       } else {
-
-        const currentFileList = await listFiles(pathToTempProj, '');
-
-        const filesToDeleteSet = new Set(currentFileList);
-        keepFileList.forEach((f) => filesToDeleteSet.delete(f));
-        const filesToDelete = Array.from(filesToDeleteSet);
+        const pathToProj = path.join(global.codewind.CODEWIND_WORKSPACE, project.name);
```
hhellyer: Should you use project.projectPath() here? I think there may be cases the project directory doesn't match the name. They are certainly different fields in the project object.

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -212,24 +212,23 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
     const project = user.projectList.retrieveProject(projectID);
     if (project) {
       const pathToTempProj = path.join(global.codewind.CODEWIND_WORKSPACE, global.codewind.CODEWIND_TEMP_WORKSPACE, project.name);
+      const tempProjectExists = await fs.pathExists(pathToTempProj);
       // eslint-disable-next-line no-sync
```
rwalle61: don't need this anymore

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -320,7 +329,6 @@ async function listFiles(absolutePath, relativePath) {
     const nextRelativePath = path.join(relativePath, f);
```
rwalle61: perhaps out of scope but fyi there are open source node modules that do this function (better than us) e.g. https://www.npmjs.com/package/recursive-readdir-async

test/src/unit/routes/projects/remoteBind.route.test.js
```diff
@@ -0,0 +1,164 @@
...
+
+function createDirectoriesFromArray(directory, dirArray) {
+    const promiseArray = dirArray.map(dir => {
+        return fs.ensureDir(path.join(directory, dir));
+    });
+    return Promise.all(promiseArray);
+}
+
+function assertFilesExist(directory, fileArray) {
+    for (let i = 0; i < fileArray.length; i++) {
```
rwalle61:
```suggestion
    for (const relativePath of fileArray) {
```

PR #1313, opened 2019-11-29
trigger build on metrics injection

tobespc:

When the injection api is called to either enable of disable the injection, we need to trigger a build straight away. This will require the project code in the pfe container to be synced to the build container and a build then triggered

### DIFF
--- a/src/pfe/portal/routes/projects/metrics.route.js
+++ b/src/pfe/portal/routes/projects/metrics.route.js
@@ -194,8 +194,6 @@ router.post('/api/v1/projects/:id/metrics/inject', validateReq, async function (
     // We now need to synch files in to the build container
     copyProjectToBuild(project);
 
-    user.buildProject(project, "build");
-
     res.sendStatus(202);
 
   } catch (err) {
@@ -213,6 +211,7 @@ async function copyProjectToBuild(project){
       globalProjectPath,
       projectRoot
     );
+    user.buildProject(project, "build");
   } else {
     // if a build is in progress, wait 5 seconds and try again
     await cwUtils.timeout(5000)
--- a/src/pfe/portal/routes/projects/metrics.route.js
+++ b/src/pfe/portal/routes/projects/metrics.route.js
@@ -192,7 +192,7 @@ router.post('/api/v1/projects/:id/metrics/inject', validateReq, async function (
 
     await metricsService.injectMetricsCollectorIntoProject(project.projectType, path.join(project.workspace, project.directory));
     // We now need to synch files in to the build container
-    copyProjectToBuild(project);
+    copyProjectToBuild(project, user);
 
     res.sendStatus(202);
 
@@ -202,7 +202,7 @@ router.post('/api/v1/projects/:id/metrics/inject', validateReq, async function (
   }
 });
 
-async function copyProjectToBuild(project){
+async function copyProjectToBuild(project, user){
   const globalProjectPath = path.join(project.workspace, project.name);
   const projectRoot = cwUtils.getProjectSourceRoot(project);
   if (project.buildStatus != "inProgress") {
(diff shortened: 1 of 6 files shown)


src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -254,24 +256,21 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
...
 
-async function syncToBuildContainer(project, filesToDelete, pathToTempProj, modifiedList, timeStamp, IFileChangeEvent, user, projectID) {
+async function syncToBuildContainer(project, filesToDelete, modifiedList, timeStamp, IFileChangeEvent, user, projectID) {
   // If the current project is being built, we do not want to copy the files as this will
   // interfere with the current build
   if (project.buildStatus != "inProgress") {
     const globalProjectPath = path.join(project.workspace, project.name);
     // We now need to remove any files that have been deleted from the global workspace
     await Promise.all(filesToDelete.map(oldFile => cwUtils.forceRemove(path.join(globalProjectPath, oldFile))));
     // now move temp project to real project
```
rwalle61: could you delete this comment as well as the line below? (It's now misleading)

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -231,15 +231,17 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
           filesToDelete.map(oldFile => cwUtils.forceRemove(path.join(pathToTempProj, oldFile)))
         );
         res.sendStatus(200);
-
+ 
+        await cwUtils.copyProject(pathToTempProj, path.join(project.workspace, project.directory), getMode(project));
```
rwalle61: Looks like a good change - I don't know how metrics injection was working previously, when this line was in `syncToBuildContainer`. (How did we inject metrics before the project was copied from `cw-temp` to `codewind-workspace`?)

src/pfe/portal/routes/projects/metrics.route.js
```diff
@@ -197,6 +204,21 @@ router.post('/api/v1/projects/:id/metrics/inject', validateReq, async function (
...
+  const projectRoot = cwUtils.getProjectSourceRoot(project);
+  if (project.buildStatus != "inProgress") {
+    await cwUtils.copyProjectContents(
+      project,
+      globalProjectPath,
+      projectRoot
+    );
+  } else {
+    // if a build is in progress, wait 5 seconds and try again
+    await cwUtils.timeout(5000)
```
hhellyer: Not sure what to do about this but if a long build (liberty) is in progress when metrics injection is turned on then `/api/v1/projects/:id/metrics/inject` could take a really long time to return.

src/pfe/portal/routes/projects/metrics.route.js
```diff
@@ -187,6 +190,10 @@ router.post('/api/v1/projects/:id/metrics/inject', validateReq, async function (
       injectMetrics: injectMetrics
     });
 
+    await metricsService.injectMetricsCollectorIntoProject(project.projectType, path.join(project.workspace, project.directory));
+    // We now need to synch files in to the build container
+    copyProjectToBuild(project);
```
rwalle61: if we don't `await` this (and `user.buildProject(...)` below), then our try-catch block won't catch any errors from them. To achieve Howard's  requested change, maybe we should keep this `await`, add an `await` onto `buildProject`, then move `res.sendStatus(202)` above them?

I'd be tempted to add a response body informing the user why they're getting a 202 rather than 200 (e.g. `res.status(202).send('Metrics collector successfully injected. Project rebuild queued.');`

rwalle61: if we don't `await` this (and `user.buildProject(...)` below), then our try-catch block won't catch any errors from them. To achieve Howard's  requested change, maybe we should keep this `await`, add an `await` onto `buildProject`, then move `res.sendStatus(202)` above them?

I'd be tempted to add a response body informing the user why they're getting a 202 rather than 200 (e.g. `res.status(202).send('Metrics collector successfully injected. Project rebuild queued.');`

Issue #1062, opened 2019-11-08
Turbine Test: Win Liberty Project Create Failures


maysunfaisal:
**Codewind version:** latest
**OS:** Win

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
The Win Turbine func tests seem to be failing on Liberty create
```

[34m[08/11/19 06:01:27 codewindtestliberty] [TRACE][39m  ---> 1adafcb9c97d

[34m[08/11/19 06:01:27 codewindtestliberty] [TRACE][39m Step 24/24 : USER default

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m  ---> Running in fe249875871d

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m Removing intermediate container fe249875871d

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m  ---> 4f239e7ed4ac

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m Successfully built 4f239e7ed4ac

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m Successfully tagged cw-codewindtestliberty-liberty1573190740157:latest

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m Successfully built cw-codewindtestliberty-liberty1573190740157

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m Container image has been built successfully
Updating build state for project codewindtestliberty(liberty1573190740157) to: inProgress

[91m[08/11/19 06:01:28 codewindtestliberty] [ERROR][39m executeBuildScript : 48 | Update build state request failed for project: codewindtestliberty(liberty1573190740157), with code: 404, and message: Not Found

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m Container built
* Starting container. 

[34m[08/11/19 06:01:28 codewindtestliberty] [TRACE][39m Start the container
Run command to start the container: /file-watcher/idc/artifacts/run_docker.sh cw-codewindtestliberty-liberty1573190740157 cw-codewindtestliberty-liberty1573190740157 "-p 127.0.0.1::7777 -P" /codewind-workspace/.idc/docker/docker-c6f926dbd9ee2b9006ad19e42c84133e14bfafec /codewind-workspace "codewindtestliberty-liberty1573190740157"
[...]

PR #1338, opened 2019-12-02
Fix run docker statement

ssh24:
Potential fix for https://github.com/eclipse/codewind/issues/1062

I also saw this on linux and this PR this solved it.


ssh24:
@maysunfaisal Could you PTAL?

maysunfaisal:
sorry for double approval/comment. Thought it was the master PR

ssh24:
Still think it should go into 0.6.1 because it was seen consistently when doing performance testing over a long run period

rajivnathan:
I think it should be done for all projects and for both 0.6.1 and master @eyuen thoughts?

rajivnathan:
As discussed, we'll target this fix only for master since it turned out to be too late for 0.6.1

PR #1349, opened 2019-12-03
bump to centos8 to pick up latest python

tobespc:

This pr bumps the level of centos to be centos8.  This is to pick up the latest Python release as the 2.7 version in centos7 has security issues.   

Centos8 no longer package 2 of the modules required by buildah so I have added them manually to the docker-compose file

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,8 @@ pipeline {
     }
     
     triggers {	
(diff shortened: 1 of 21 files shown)


johnmcollier:
@tobespc Buildah can be rather finicky, have you been able to verify on Kube that it still works after adding the packages that it requires?

Issue #1320, opened 2019-11-29
Move Codewind to Helm 3


johnmcollier:
Helm 3 has officially GA'd and has removed the tiller! 

Codewind currently uses Helm 2.9.1 which is quite back level and doesn't work on newer Kubernetes (such as 1.16), as it uses resources that have since been deleted from the Kube API. While we are investigating having Codewind deploy projects without Helm, that's still a ways away and we're still very much reliant on Helm for deploying Codewind-style applications.

We should see if we can move Codewind up to Helm 3, so that:
- We can function on Kube 1.16
- We can remove our dependency on the Helm tiller
- We can move away from an old, unsupported Helm release

Existing charts should be compatible with Helm so I wouldn't expect any migration needed for old projects.

johnmcollier:
Closing, as Helm 3 has been delivered into master.

PR #1358, opened 2019-12-03
Issue #1320: Update to Helm 3 for Codewind

johnmcollier:
This PR updates Codewind to use Helm 3 to deploy Codewind-style applications:
- Pulls down the Helm 3.0.0 binary and updates the SHA in the PFE dockerfile accordingly
- Removes the code that initializes the tiller:
   - Helm 3 has no tiller, thus no tiller to initialize
- Removes the outdated helm-wrapper.sh script.
    - It was only being used when `USE_HELM_TLS` was set to true, and with the tiller gone, it's no longer applicable.
- Updates `helm template` and `helm upgrade` commands as the format has changed slightly
   - Need to use `helm template <name> <chart-dir> <flags>` and `helm upgrade <name> <chart-dir> <flags>`
   - The `--name` flag has also been removed from both commands, as it doesn't exist in Helm 3 any more (see previous point)
- Removes the `--purge` flag from all `helm delete` commands, as it too has been removed.
- Removes any parameters from `helm list` commands, as it now only accepts flags

This PR relies on https://github.com/eclipse/codewind-che-plugin/pull/123 being merged first.


johnmcollier:
Since I don't want a change like this accidentally being delivering in 0.7.0, I'm keeping this PR in draft-mode until we've branched off and will deliver in master then.

johnmcollier:
0.7.0 has been cut, so removing the `draft` status from this PR

johnmcollier:
Functional Test logs look fine to me: 
[functional-test-helm.log](https://github.com/eclipse/codewind/files/3953103/functional-test-helm.log)

Issue #1412, opened 2019-12-06
Fix registrysecrets API error handling


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
When service account patch fails, GET API returns the new list from Docker Config which it should not because the Codewind Secret create and/or Service Account patch failed.

Revert changes to the Docker config when this happens.

Applies to POST and DELETE.

![image (6)](https://user-images.githubusercontent.com/31771087/70337609-71c1d980-1819-11ea-90a2-515f1b843aaa.png)

@tetchel 

**Steps to reproduce:**

**Workaround:**

jgwest:
@maysunfaisal: Could not find label: kind/hot

tetchel:
well, it should return the new list, but the new list should equal the old list because we failed to modify it :P

maysunfaisal:
@tetchel changes have been merged

tetchel:
neato

malincoln:
@tetchel @maysunfaisal can you verify and close?  Thanks

Issue #1300, opened 2019-11-27
Update /api/v1/registry to better handle /api/v1/registrysecrets


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description of the enhancement:**

Update /api/v1/registry to better handle /api/v1/registrysecrets

**Proposed solution:** 

- [x] Rename `/api/v1/registry` to `/api/v1/imagepushregistry` for better semantics
- [x] Update `POST /api/v1/registry` https://eclipse.github.io/codewind/#/paths/~1api~1v1~1registry/post to have `address` and `namespace` instead of `deploymentRegistry` (reason see next point)
- [x] Update `GET /api/v1/registry` https://eclipse.github.io/codewind/#/paths/~1api~1v1~1registry/get to have `imagePushRegistry` instead of `deploymentRegistry` and when `true` it will include `address` & `namespace` which IDE can use to match Registry Secrets from `/api/v1/registrysecrets`
- [x] Update `POST /api/v1/registry` ie; when an API call is made with `address: https://index.docker.io/v1/`, PFE will store the value in metadata(`/codewind-workspace/.config/settings.json`) but when using it to push application images to the registry namespace, PFE will use `docker.io` instead
- [x] Rename all registry APIs`url` property to `address`
- [x] Rename all occurances of `xyzdeploymentxyz` property to `xyzimagePushxyz`
- [x] Remove temporary API `GET /api/v1/registry` which always returns true
- [x] Cleanup `getWorkspaceSettingsInfo(true)`
- [x] Set Codewind registry secret owner to CW PVC instead of CW Deployment

maysunfaisal:
@rajivnathan can you pls move this to in progress, thx

PR #1416, opened 2019-12-06
Handle Docker Config if Secret create and/or Service Acc patch fails

maysunfaisal:

Issue #1412 
If Secret create and/or Service Account patch fails, revert changes made to the Docker Config

Issue https://github.com/eclipse/codewind/issues/1300
Removes the temp API `GET /api/v1/registry` which always returned True

### DIFF
--- a/src/pfe/portal/modules/User.js
+++ b/src/pfe/portal/modules/User.js
@@ -667,7 +667,8 @@ module.exports = class User {
   }
 
   /**
-   * Function to setup the Docker config file and Kube secret
+   * Function to setup the Docker config file. In Kubernetes, the function also creates the Codewind secret from
+   * the Docker config and patches the Service Account with the created Codewind Secret. On local, this is skipped.
    */
   async setupRegistrySecret(credentials, address) {
     const credentialsJSON = JSON.parse(Buffer.from(credentials, "base64").toString());
@@ -737,8 +738,10 @@ module.exports = class User {
 
     // Only update the Kube secret when running in K8s
     let isServiceAccountPatched = false;
+    let serviceAccountPatchData;
     if (global.codewind.RUNNING_IN_K8S) {
-      isServiceAccountPatched = await this.updateServiceAccountWithDockerRegisrySecret();
+      serviceAccountPatchData = await this.updateServiceAccountWithDockerRegisrySecret();
+      isServiceAccountPatched = serviceAccountPatchData.isServiceAccountPatched;
     }
 
     if (isServiceAccountPatched || !global.codewind.RUNNING_IN_K8S) {
@@ -747,24 +750,25 @@ module.exports = class User {
       return registrySecretList;
     }
 
-    const msg = "Failed to create the Codewind Secret and/or patch the Service Account, reverting changes to the Docker Config";
+    const msg = "Reverting changes to the Docker Config";
     log.error(msg);
     
-    // Since patching the Service Account failed, we need to revert the update to the Docker Config
+    // Since patching the Service Account failed, we need to revert the update to the Docker Config and patch the Service Account again
     delete jsonObj.auths[address];
     await fs.writeJson(this.dockerConfigFile, jsonObj);
+    await this.updateServiceAccountWithDockerRegisrySecret();
 
-    throw new RegistrySecretsError("SERVICE_ACCOUNT_PATCH_FAILED", "for address " + address);
+    throw new RegistrySecretsError(serviceAccountPatchData.error, "for address " + address);
   }
 
   /**
-   * Function to create the Kubernetes secret and patch the Service Account
+   * Function to create the Kubernetes Secret and patch the Service Account with the created Secret
    */
   async updateServiceAccountWithDockerRegisrySecret() {
-    try {
-      log.info("Creating a Secret and patching the Service Account");
-      const isDockerConfigFilePresent = await cwUtils.fileExists(this.dockerConfigFile)
-      if (isDockerConfigFilePresent) {
+    log.info("Creating a Secret and patching the Service Account");
+    const isDockerConfigFilePresent = await cwUtils.fileExists(this.dockerConfigFile);
+    if (isDockerConfigFilePresent) {
+      try {
         log.info("The Docker config file exists, reading contents");
         const jsonObj = await fs.readJson(this.dockerConfigFile);
         const encodedDockerConfig = Buffer.from(JSON.stringify(jsonObj)).toString("base64");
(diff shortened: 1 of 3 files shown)


maysunfaisal:
@rajivnathan @tobespc @hhellyer pls help review, thx

src/pfe/portal/modules/User.js
```diff
@@ -746,8 +747,13 @@ module.exports = class User {
       return registrySecretList;
     }
 
-    const msg = "Failed to patch the Service Account";
+    const msg = "Failed to create the Codewind Secret and/or patch the Service Account, reverting changes to the Docker Config";
     log.error(msg);
+    
+    // Since patching the Service Account failed, we need to revert the update to the Docker Config
+    delete jsonObj.auths[address];
+    await fs.writeJson(this.dockerConfigFile, jsonObj);
```
rajivnathan: I'm not sure this will revert the file to the original content prior to this request. Looks like it will only remove the address from the file. If the file existed before we write to it then should we keep the contents of the file in memory? If `updateServiceAccountWithDockerRegisrySecret()` returns false we write the original content back to the file?

src/pfe/portal/modules/User.js
```diff
@@ -746,8 +747,13 @@ module.exports = class User {
       return registrySecretList;
     }
 
-    const msg = "Failed to patch the Service Account";
+    const msg = "Failed to create the Codewind Secret and/or patch the Service Account, reverting changes to the Docker Config";
```
rajivnathan: 1. Looks like `updateServiceAccountWithDockerRegisrySecret()` will output the failure messages so can this message just be to indicate the changes are being reverted?

2. In `updateServiceAccountWithDockerRegisrySecret()` can we separate the secret POST and the service account PATCH calls into separate blocks so that we can have specific errors for each?

src/pfe/portal/modules/User.js
```diff
@@ -746,8 +747,13 @@ module.exports = class User {
...
     log.error(msg);
+    
+    // Since patching the Service Account failed, we need to revert the update to the Docker Config
+    delete jsonObj.auths[address];
+    await fs.writeJson(this.dockerConfigFile, jsonObj);
+
```
rajivnathan: Should `updateServiceAccountWithDockerRegisrySecret()` be called again after reverting the changes to the file?

src/pfe/portal/modules/User.js
rajivnathan: Maysun noted that an error will be returned if the address already exists since update is not supported. So it's just the case where adding a new entry to an existing registry secret fails that we need to handle.

src/pfe/portal/modules/User.js
```diff
@@ -737,8 +737,10 @@ module.exports = class User {
 
     // Only update the Kube secret when running in K8s
     let isServiceAccountPatched = false;
+    let serviceAccountPatchData;
     if (global.codewind.RUNNING_IN_K8S) {
-      isServiceAccountPatched = await this.updateServiceAccountWithDockerRegisrySecret();
+      serviceAccountPatchData = await this.updateServiceAccountWithDockerRegisrySecret();
```
rajivnathan: Can you please add some comments to the description of the `setupRegistrySecret` function that explains the usage of this function for the non-k8s cases? I'm sure others would wonder in the future.

src/pfe/portal/modules/User.js
```diff
@@ -736,8 +737,10 @@ module.exports = class User {
 
     // Only update the Kube secret when running in K8s
     let isServiceAccountPatched = false;
+    let serviceAccountPatchData;
```
rajivnathan: Does the error on line 735 need a specific message key like the other errors that are thrown in this function?

maysunfaisal: Not necessarily, this is for the overall try catch for updating docker config. If there is an exception on a place where i am not throwing, then this will return the error to the IDE caller.

src/pfe/portal/modules/User.js
maysunfaisal: done

src/pfe/portal/modules/User.js
maysunfaisal: done

src/pfe/portal/modules/User.js
```diff
@@ -922,9 +959,15 @@ module.exports = class User {
...
 
-    const msg = "Failed to patch the Service Account";
+    const msg = "Reverting changes to the Docker Config";
     log.error(msg);
-    throw new RegistrySecretsError("SERVICE_ACCOUNT_PATCH_FAILED", "for address " + address);
+
+    // Since patching the Service Account failed, we need to revert the delete from the Docker Config and patch the Service Account again
+    jsonObj.auths[address] = registrySecretToBeDeleted;
+    await fs.writeJson(this.dockerConfigFile, jsonObj);
+    await this.updateServiceAccountWithDockerRegisrySecret();
```
hhellyer: I might be missing something kubernetes related here but if we fail to patch the service account:
- Do we need to repatch it back to the old settings? Won't they still be set?
- Will we be able to change them? The patch operation just failed once, why will it work this time?

src/pfe/portal/modules/User.js
```diff
@@ -746,19 +750,25 @@ module.exports = class User {
...
+
+    throw new RegistrySecretsError(serviceAccountPatchData.error, "for address " + address);
   }
 
   /**
-   * Function to create the Kubernetes secret and patch the Service Account
+   * Function to create the Kubernetes Secret and patch the Service Account with the created Secret
    */
   async updateServiceAccountWithDockerRegisrySecret() {
```
hhellyer: This function is a little odd now, it either returns an object that contains `true` or one that contains `false` and an error message. Would it be easier to return an error message on failure and `undefined`/`null` on success?
You could also throw the error but I can see why that might be tricky with lots of different callers.

It's a bit odd to return an error or nothing in Node.js but C does it all the time with non-zero return codes. ;-)

src/pfe/portal/modules/User.js
maysunfaisal: 1. It may or may not, I think in the case of a failed API operation, since we're rolling back to the previous state, does it not make sense to have the secret re-created with the prev docker config and patch the SA?

2. Yes it may have failed in the API operation(could be secret creation failed or patching SA failed), but we should at least attempt to restore it back to what it was.

What do you think?

src/pfe/portal/modules/User.js
maysunfaisal: I'll update it to be either undefined or ERR_CODE

src/pfe/portal/modules/User.js
rajivnathan: Actually, I see Howard's point. If the patch fails then the only the docker config file and secret need to be rolled back. What we have should be harmless though because the patch to roll back should just fail and if it somehow works it'll match up with the secret.

Issue #1401, opened 2019-12-05
Appsody extension should pass env to Appsody CLI via --docker-options


makandre:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description of the enhancement:**

Appsody CLI supports passing `--docker-options` to it.  Codewind should use that for passing environment, either using `--env` or `--env-file` (TBD) to Appsody

**Proposed solution:**

makandre:
One other thing needed is for Codewind to re-build the project (i.e. re-invoke `appsody run`) when the `env.list` file changes.  But currently `detectChangeByExtension = true` for `ShellExtensionProject` means that this does not happen.  This would need to be enhanced to allow certain files to trigger Codewind re-build.

https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts#L88

makandre:
Discuss with appsody team; they made `-e`/`--env` flag take effect on local and k8s, however it's preferable for us to use `--env-file` (PR https://github.com/eclipse/codewind-appsody-extension/pull/73 passes `--env-file`)

Opened followup issue https://github.com/appsody/appsody/issues/784 on appsody

micgibso:
Epic #934 , this not yet ready for doc

makandre:
The name of the file is being changed from `env.list` to `env.properties`.  See https://github.com/eclipse/codewind/issues/1692

sghung:
TODO: draft up the content for ID to pick up. TBD by Sarah as to where this content will be placed.

Remember to add note to not check this into a repository. May want to put example of what an env.properties file looks like.

makandre:
Draft doc: @sishida 

### Defining Environment Variables for Appsody Projects

To define environment variables that would take effect in an appsody application:

1.  Create a file named `env.properties` in the root of the appsody project
2. Define your envirnoment variables in this file using standard properties format, i.e. one `name=value` entry per line
3. If autobuild is enabled, Codewind will automatically rebuild the project to pick up the environment values.  Otherwise, the new values will take effect the next time the project is rebuilt.

> Note: You should avoid committing the `env.properties` file to your source repository if it contains sensitive information such as passwords

sishida:
@makandre Could you please review this PR and add your approval if it looks good to go? I just added some edits to the content that you originally provided. Thank you! :)

sishida:
[...]

PR #1422, opened 2019-12-06
Allow project extension to specify certain files that should trigger a Codewind rebuild

makandre:

This PR is for eclipse/codewind#1401.  In that issue, we are adding ability to pass an `env.list` file to the Appsody CLI.  When this file changes, we would require Codewind to rebuild the project (reruns `appsody run`) to pick up new environment values.

But currently, all file changes in `ShellExtensionProjects` do not trigger Codewind rebuild since it is assumed the extension will watch and handle all changes.

This PR allows project extensions to at least specify that some files should trigger Codewind to rebuild.

PR on appsody extension side is https://github.com/eclipse/codewind-appsody-extension/pull/73


src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
```diff
@@ -42,6 +42,30 @@ const workspaceSettingsInfo =  workspaceSettings.workspaceSettingsInfoCache ? JS
...
+/**
+ * Determine if Codewind should handle a detected file change, or if it is handled by an extension
+ *
+ * @param detectChangeByExtension property from the project handler
+ * @param path the path of detected file change
+ */
+function shouldHandle(detectChangeByExtension: boolean | string[], path: string): boolean {
+
+    // clearly Codewind should handle when detectChangeByExtension is false
+    if (!detectChangeByExtension) {
```
elsony: Will the code be simpler if you just check if it is either true or false, then just return at this point. If it falls through, then it is an array.

makandre: Thanks for the suggestion. Updated the logic.

sghung:
@elsony thanks your review. Andrew has made the fix, can you re-review? Thanks!

PR #1462, opened 2019-12-10
Templates API test, fix failing ID check

james-wallis:
### Summary
@makandre pointed out that the Template API tests were failing in Jenkins, this appears to be because the `id` field of a Template repository is missing. 

When the `repository_list.json` file is first created it does not have `id` fields for any of the repositories. They are added when a POST request happens, likely when the state is reset after a test run. This means that the first test run would fail but subsequent ones would pass as the ID field is present.

The API docs show that only `url` is required for the GET API and so I have relaxed the tests to only test for the required parameters.

[Codewind API Doc](https://eclipse.github.io/codewind/#/paths/~1api~1v1~1templates~1repositories/get)

### Testing
* Ran the API tests


jcockbain:
LGTM

Issue #1337, opened 2019-12-02
Update Turbine Kube Tests for Registry Secrets API


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
When #665 is consumed by the IDE, Functional tests need to set the Registry Secrets via the PFE API rather than thru the Che dashboard.

**Steps to reproduce:**

**Workaround:**

malincoln:
@maysunfaisal pls verify and close if resolved. Thanks

PR #1470, opened 2019-12-10
Add in the registrysecrets API for Turbine Functional test on Kube

maysunfaisal:

Issue https://github.com/eclipse/codewind/issues/1337

### DIFF
--- a/src/pfe/file-watcher/server/test/scripts/exec.sh
+++ b/src/pfe/file-watcher/server/test/scripts/exec.sh
@@ -53,6 +53,18 @@ function downloadCwctl() {
     checkExitCode $? "Failed to give correct permission to run installer."
 }
 
+function setupRegistrySecret() {
+    echo -e "${BLUE}Setting up the Registry Secret ... ${RESET}"
+    ENCODED_CREDENTIALS=$( node ./scripts/utils.js encode $REGISTRY_SECRET_USERNAME $REGISTRY_SECRET_PASSWORD )
+    if [[ ($? -ne 0) ]]; then
+        echo -e "${RED}Test setup failed during Registry Secret's base64 credential encode. ${RESET}\n"
+        exit 1
+    fi
+
+    REGISTRY_SECRET_SETUP_CURL_API="curl -d '{\"address\": \"$REGISTRY_SECRET_ADDRESS\", \"credentials\": \"$ENCODED_CREDENTIALS\"}' -H \"Content-Type: application/json\" -X POST https://localhost:9191/api/v1/registrysecrets -k"
+    kubectl exec -i $CODEWIND_POD_ID -- bash -c "$REGISTRY_SECRET_SETUP_CURL_API"
+}
+
 function createProject() {
    ./$EXECUTABLE_NAME project create --url $1 $2
 }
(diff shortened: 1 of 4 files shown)


maysunfaisal:
@ssh24 @rajivnathan pls help review PR, thx a lot

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -53,6 +53,42 @@ function downloadCwctl() {
...
+
+    oc policy add-role-to-user system:image-builder system:serviceaccount:$NAMESPACE:$REGISTRY_SECRET_USERNAME
+
+    ENCODED_TOKEN=$(oc get secret $(oc describe sa $REGISTRY_SECRET_USERNAME | tail -n 2 | head -n 1 | awk '{$1=$1};1') -o json | jq ".data.token")
+    REGISTRY_SECRET_PASSWORD=$( node ./scripts/utils.js decode $ENCODED_TOKEN )
+}
+
+function setupRegistrySecret() {
+    if [[ ! -z $INTERNAL_REGISTRY ]]; then
+        setupInternalRegistryCredentials
```
ssh24: So if internal registry is passed, we will set the default registry as well the user registry? I think we should exit out of the function if internal registry was passed

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -139,9 +175,16 @@ function setup {
...
     done
+
+    if [ $TEST_TYPE == "kube" ]; then
+        # Set up registry secrets (docker config) in PFE container/pod
+        setupRegistrySecret
+    fi
 }
 
 function run {
+    TURBINE_EXEC_TEST_CMD="cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml IMAGE_PUSH_REGISTRY_ADDRESS=${REGISTRY_SECRET_ADDRESS} IMAGE_PUSH_REGISTRY_NAMESPACE=${IMAGE_PUSH_REGISTRY_NAMESPACE} npm run $TEST_SUITE:test:xml"
```
ssh24: Any reason it was moved from the original place?

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: we cant exit, `setupInternalRegistryCredentials`just  sets the variables for address, username and password with the logic from che setup script

The rest of the function `setupRegistrySecret` (diff function, look at the name) calls the curl in PFE. So basically `setupInternalRegistryCredentials` just overwrites the variables before calling curl

src/pfe/file-watcher/server/test/scripts/exec.sh
maysunfaisal: since we're changing the address variable `REGISTRY_SECRET_ADDRESS` in function `setupInternalRegistryCredentials` above, we need the `TURBINE_EXEC_TEST_CMD` var to be declared after that function call since `TURBINE_EXEC_TEST_CMD` uses the address variable `REGISTRY_SECRET_ADDRESS`.

I can explain it better in person if that didnt make sense

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: Oh, I see. Ok then its fine

src/pfe/file-watcher/server/test/scripts/exec.sh
ssh24: Nope, it makes sense now

src/pfe/file-watcher/server/test/scripts/utils.js
```diff
@@ -0,0 +1,50 @@
...
+if (process.argv.length < 3) {
+    printUsage();
+}
+
+const option = process.argv[2];
+
+if (option.includes("help") || option == "-h") {
+    printUsage();
+} else if (option == "encode") {
+    if (process.argv[3] == undefined || process.argv[4] == undefined || process.argv.length != 5) {
```
ssh24: can we simplify this logic by only checking if the number of args is less than 5?
```
if (process.argv.length < 5)
```

src/pfe/file-watcher/server/test/scripts/utils.js
```diff
@@ -0,0 +1,50 @@
...
+        printUsage();
+    }
+    encodeCredentials(process.argv[3], process.argv[4]);
+} else if (option == "decode") {
+    if (process.argv[3] == undefined || process.argv.length != 4) {
```
ssh24: same here, just `if (process.argv.length < 4)`

src/pfe/file-watcher/server/test/scripts/utils.js
```diff
@@ -0,0 +1,50 @@
...
+    const credentials = { username: username, password: password };
+    const credentialsJSON = JSON.stringify(credentials);
+    const encodedCredentials = Buffer.from(credentialsJSON).toString("base64");
+    console.log(encodedCredentials);
+}
+
+function decodeCredentials(encodedCredentials) {
+    const decodedCredentials = Buffer.from(encodedCredentials, "base64").toString();
+    console.log(decodedCredentials);
+}
```
ssh24: nitpick: new line

src/pfe/file-watcher/server/test/scripts/utils.js
maysunfaisal: updated, Idk why i did that 😅

src/pfe/file-watcher/server/test/scripts/utils.js
maysunfaisal: updated, Idk why i did that 😅

src/pfe/file-watcher/server/test/scripts/utils.js
maysunfaisal: updated

Issue #1486, opened 2019-12-11
PFE does not update pushregistry if pushregistry namespace is the empty string


tetchel:
Pushregistry namespace needs to be allowed to be `""` because not all registry providers use namespaces

POST /imagepushregistry
`{ operation: "set", address: "docker.io", namespace: "" }`
--> appears to succeed
GET /imagepushregistry
`{ imagePushRegistry: false }`

There is a condition in PFE which basically checks `if (namespace)` which fails if the namespace is the empty string, it needs to be `if (namespace != null)`

And if the namespace is missing, it should respond 400 since the request is missing required fields

jagraj:
SVT also into similar issue.

maysunfaisal:
Dont think its a stopship since its a special case but i have a PR up for it

malincoln:
fix is ready; will go ahead and merge and have @hhellyer review on Monday

malincoln:
@tetchel the fix has been merged into 0.7.0 branch.  Can you test and verify?  Thanks

tetchel:
❤️

Issue #1484, opened 2019-12-11
No error if specific invalid image registry namespace is used


tetchel:
From #1483 

The app fails to start, but no error message is shown in the UI or in the project's logs.

Likely there is an error in the PFE logs.

![image](https://user-images.githubusercontent.com/8888438/70641555-d01cfc80-1c0b-11ea-83dc-7304d682dafe.png)

@sujeilyfonseca

maysunfaisal:
namespace: `docker.io/maysunfaisal`
```
[11/12/19 17:10:59 gomaysun1] [DEBUG] Run buildah command: buildah push,--tls-verify=false,cw-gomaysun1-1cc1adb0-1c39-11ea-ba7d,docker.io/docker.io/maysunfaisal/cw-gomaysun1-1cc1adb0-1c39-11ea-ba7d
[11/12/19 17:10:59 gomaysun1] [ERROR] containerBuildAndRun : 1394 | Getting image source signatures

[11/12/19 17:11:00 gomaysun1] [ERROR] containerBuildAndRun : 1394 | Copying blob sha256:7f9bf938b0531c93cc0d860e72aac23093189889d89604ce24b95197435b76cf

[11/12/19 17:11:00 gomaysun1] [ERROR] containerBuildAndRun : 1394 | Copying blob sha256:423d63eb4a274765e80c67db008e5cb9318b717c3a57790b75b53d3440a9f3e0

...
```

namespace: `sdfdd`(rubbish)
```
[11/12/19 17:14:26 sdfsdf] [DEBUG] Run buildah command: buildah push,--tls-verify=false,cw-sdfsdf-a97656c0-1c39-11ea-ba7d,docker.io/sdfdd/cw-sdfsdf-a97656c0-1c39-11ea-ba7d
[11/12/19 17:14:26 sdfsdf] [ERROR] containerBuildAndRun : 1394 | Getting image source signatures

[...]

PR #1492, opened 2019-12-11
Handle empty namespace and invalid namespace

maysunfaisal:

This PR has the following issues:
- https://github.com/eclipse/codewind/issues/1484 If a user enters namespace as `address/namespace` in the UI, strip off address and write to workspace settings file
- https://github.com/eclipse/codewind/issues/1486 if a user enters an empty namespace `""`, handle the case in PFE as valid

### DIFF
--- a/src/pfe/file-watcher/server/src/utils/workspaceSettings.ts
+++ b/src/pfe/file-watcher/server/src/utils/workspaceSettings.ts
@@ -237,7 +237,13 @@ export async function getImagePushRegistry(): Promise<string> {
         return "";
     }
     logger.logInfo("WorkspaceSettingsInfo for project operations: " + JSON.stringify(workspaceSettingsInfo));
-    const imagePushRegistry = workspaceSettingsInfo.registryAddress.trim() + "/" + workspaceSettingsInfo.registryNamespace.trim();
+    let imagePushRegistry = workspaceSettingsInfo.registryAddress.trim();
+
+    // Only append if user entered a namespace
+    if (workspaceSettingsInfo.registryNamespace.length > 0) {
+        imagePushRegistry = imagePushRegistry + "/" + workspaceSettingsInfo.registryNamespace.trim();
+    }
+
     return imagePushRegistry;
 }
 
--- a/src/pfe/file-watcher/server/src/utils/workspaceSettings.ts
+++ b/src/pfe/file-watcher/server/src/utils/workspaceSettings.ts
@@ -98,11 +98,6 @@ export async function readWorkspaceSettings(): Promise<IWorkspaceSettingsSuccess
 }
 
 export async function writeWorkspaceSettings(address: string, namespace: string): Promise<IWorkspaceSettingsSuccess | IWorkspaceSettingsFailure> {
-    // Clean up namespace if user enters address/namespace in the UI
-    if (namespace.startsWith(address.replace(/\/\s*$/, "") + "/")) {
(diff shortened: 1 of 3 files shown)


maysunfaisal:
@hhellyer @rajivnathan can you pls help review? If the changes are good, I will open one PR for 0.7.0

src/pfe/file-watcher/server/src/utils/workspaceSettings.ts
```diff
@@ -98,6 +98,11 @@ export async function readWorkspaceSettings(): Promise<IWorkspaceSettingsSuccess
 }
 
 export async function writeWorkspaceSettings(address: string, namespace: string): Promise<IWorkspaceSettingsSuccess | IWorkspaceSettingsFailure> {
+    // Clean up namespace if user enters address/namespace in the UI
+    if (namespace.startsWith(address.replace(/\/\s*$/, "") + "/")) {
+        namespace = namespace.replace(address.replace(/\/\s*$/, "") + "/", "");
+    }
+
```
hhellyer: I don't think this is the right fix. The address and namespace parameters should be validated and sanitized when received in the REST API here:
https://github.com/eclipse/codewind/blob/2a860de3981d06b81177d7fa93d903b470ac55a1/src/pfe/portal/routes/imagePushRegistry.route.js#L51

Looking at that code the namespace is altered if the settings are used in `writeWorkspaceSettings` but not in the call to `testImagePushRegistry`. I think they should be validated and cleaned up early as part of the REST API before either of those calls are made. That way if there's a problem the REST API can return an appropriate error straight away without having to return it all the way back up the call stack and the same fixes and validation will apply to the parameters passed to `writeWorkspaceSettings` and `testImagePushRegistry`.

maysunfaisal: makes sense, let me see how i can clean it up

maysunfaisal:
@hhellyer I have pushed the changes, can you pls have a look, thx

elsony:
Based on the 0.7.0 checkpoint meeting discussion, we'll go ahead and merge this PR given that it should address Howard's original comment so that we can get it into 0.7.0 quickly.  We can address any further comment (in case there is any) from Howard in a separate PR.

src/pfe/portal/routes/imagePushRegistry.route.js
```diff
@@ -46,6 +46,18 @@ router.post('/api/v1/imagepushregistry', validateReq, async function (req, res)
     const namespace = req.sanitizeBody('namespace');
     const operation = req.sanitizeBody('operation');
 
+    // if user enters it as address/namespace in the UI, throw an error
+    if (namespace.startsWith(address.replace(/\/\s*$/, "") + "/")) {
```
hhellyer: I think this check is in the right place now but it's really checking for the one very specific error case where the user has entered the address in the namespace field.
Would it be simpler just to check for characters that are valid for a namespace? That would eliminate the `namespace` as `address/namespace` case *and* a whole load of other invalid cases. Otherwise we're adding an error check that's very specific to one particular problem.
(I'm not sure of the rules for valid `namespace` characters though I've a suspicion it's just the characters valid between `/`'s in a url.)

maysunfaisal: chatted with @hhellyer, will clean up feedback with issue https://github.com/eclipse/codewind/issues/1507 since this requires clean up anyways

hhellyer:
@maysunfaisal has added a note to #1507 to improve the validation of the parameters. This can be merged.

PR #1499, opened 2019-12-12
[0.7.0] Handle empty namespace and invalid namespace

maysunfaisal:

This PR has the following issues:
- https://github.com/eclipse/codewind/issues/1484 If a user enters namespace as `address/namespace` in the UI, strip off address and write to workspace settings file
- https://github.com/eclipse/codewind/issues/1486 if a user enters an empty namespace `""`, handle the case in PFE as valid


maysunfaisal:
@hhellyer can you pls review the exact changes for 0.7.0

elsony:
Based on the 0.7.0 checkpoint meeting discussion, we'll go ahead and merge this PR given that it should address Howard's original comment so that we can get it into 0.7.0 quickly.  We can address any further comment (in case there is any) from Howard in a separate PR.

elsony:
Note that the PR doesn't seems to get kicked off for some reason. I have discussed that on the checkpoint call and will post to Joseph for investigation.

Issue #1399, opened 2019-12-05
Project sync support for reference to files outside of project folder


makandre:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description of the enhancement:**

I would like to be able to create/define in my project, references to files that exist outside of the project folder (think of it like a symlink, but cross-platform).  Conceptually, something like this:

```
myProj/
    fileA --> /someAbsolutePath/fileX
    fileB
    fileC
```

When the contents of `fileX` changes, it should be synced into PFE as `fileA`

**Use Case:**

`fileX` could contain some (potentially sensitive) configuration information required for PFE to build and run the application, but I do not want to keep this file inside the project folder because it may get accidentally checked into a repo.   Also the configuration would be retained if I were to delete/re-clone my project etc.

**Proposed solution:** 

See comments

makandre:
I discussed with @jgwest on a possible solution.

The first part would involve adding a new optional property to the payload `filewatcherd` gets from the `/api/v1/projects/watchlist` API:

```
{
   "projects":[
      {
         "projectId": "real-project-id", /* project id of an existing project */
         "pathToMonitor":"/codewind-workspace/myProj", /* the actual real project path to monitor, since this is an existing project it will already have one */

         "syncIntoPfeContainer" : [ <-- new optional property
            {
               "from" : "/someAbsolutePath/fileX",  /* absolute path to file on local machine */
               "to" : "fileA",  /* relative to root of project, inside container */
            }
         ],

         "ignoredPaths":[
            "/target/*"
         ],
         "ignoredFilenames":[
            "*.class"
         ], 
     "projectWatchStateId" : "[...]",
      },
   ]
}
```

The `syncIntoPfeContainer` defines the additional referenced files to watch for the project; this could possibly be defined by user in `.cw-settings`

Then, whenever a file change occurs, `filewatcherd` would call:

```
cwctl project sync -p /codewind-workspace/myProj -i real-project-id -t timestamp -r "representation of the syncIntoPfeContainer JSON"
```

The new optional `-r` flag would instruct `cwctl` to sync the additional files based on the `from` and `to` of each entry.

makandre:
I have the first part, `/api/v1/projects/watchlist` API with the extra `refPaths` attribute prototyped in https://github.com/eclipse/codewind/pull/1535 (where `refPaths` is defined in `.cw-settings.)

[...]

PR #1535, opened 2019-12-13
New refPaths property in .cw-refpaths.json for filewatcherd

makandre:
For #1399

- introduce a `.cw-refpaths.json` file in which users can define `refPaths`
- include `refPaths` in response to `GET /api/v1/projects/watchlist` API
- include `refPaths` in `WebSocket` events sent to `filewatcherd` on project creation and file update


makandre:
PR updated after discussions summarized in https://github.com/eclipse/codewind/issues/1399#issuecomment-574228199

jgwest:
I have reviewed the filewatcher API additions which are made by this code, and given the code itself a once over, and looks good 👍 .

makandre:
@elsony and @tobespc could you review please? (@tobespc some changes since you last reviewed)

src/pfe/file-watcher/server/src/controllers/projectsController.ts
```diff
@@ -1055,6 +1075,11 @@ export async function updateProjectInfo(projectID: string, keyValuePair: UpdateP
         keyValuePair.saveIntoJsonFile = true;
     }
 
+    // remove case
+    if (!keyValuePair.value) {
```
rajivnathan: If `keyValuePair.value` is falsy could this lead to removing something form the projectInfo unintentionally?

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
```diff
@@ -453,6 +453,81 @@ const reconfigIgnoredFilesForDaemon = async function (ignoredPaths: string[], op
...
+                logger.logProjectInfo("The ref path list for file watching is unchanged", projectID);
+                return;
+        }
+    }
+    // if undefined and previously it was also undefined, nothing changed
+    else if (!projectInfo.refPaths) {
+        logger.logProjectInfo("The ref path list for file watching is unchanged", projectID);
+        return;
+    }
+    // otherwise the property was removed
```
rajivnathan: In `projectSpecificationHandler()` it will call this function with undefined. When `refPaths` is undefined, the event that is sent from this function will have `refPaths: undefined`? What is the contract with the file-watchers? Does it expect `undefined` or an empty array? cc @jgwest

src/pfe/file-watcher/server/src/controllers/projectsController.ts
makandre: good point here, I'll change it

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
jgwest: I'd prefer that it always be an empty array, but undefined (or not even specifying the field) is tolerated by the filewatchers.

makandre: This makes it always an array: https://github.com/eclipse/codewind/pull/1535/commits/52d2690cefcb32ca9e56f0d09a18eee24d612546

src/pfe/file-watcher/server/src/controllers/projectsController.ts
makandre: Fixed by https://github.com/eclipse/codewind/pull/1535/commits/2b2cd19689c4425d9144e44663ff603491608857

docs/openapi.yml
```diff
@@ -1899,6 +1899,11 @@ components:
           description: regex of paths to be ignored for changes
           type: array
           example: [ "/target", "/.m2", "*/.git/*" ]
+        refPaths:
```
yangcao77: Why we need another field for `refPaths`? 
I saw your code that the refpath is under `path.join(projectInfo.location, ".cw-refpaths.json");`, so it would be monitored & reported by filewatcherd, since `pathToMonitor`is the project parent folder.

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
```diff
@@ -453,6 +453,81 @@ const reconfigIgnoredFilesForDaemon = async function (ignoredPaths: string[], op
...
+ * @param args <Required | Any> - Required arguments for reconfiguration of ref paths.
+ *
+ * @returns Promise<any>
+ */
+const reconfigRefPathsForDaemon = async function (refPaths: RefPath[], operation: Operation): Promise<any> {
+    const projectInfo: ProjectInfo = operation.projectInfo;
+    const projectID = projectInfo.projectID;
+
+    // if defined, do some checks, sanitized the input etc
+    if (refPaths) {
```
yangcao77: Do we need this check?
If it's undefined, it would be filtered out in `projectSpecificationHandler`:
https://github.com/eclipse/codewind/blob/d0ec065f591c9396b08045416aacc9c00e82d03a/src/pfe/file-watcher/server/src/projects/projectSpecifications.ts#L972-L981

src/pfe/portal/modules/FileWatcher.js
```diff
@@ -429,7 +429,7 @@ module.exports = class FileWatcher {
       const projectID = fwProject.projectID;
       const project = this.user.projectList.retrieveProject(projectID);
 
-      if (fwProject.ignoredPaths || event == "newProjectAdded") {
+      if (fwProject.ignoredPaths || fwProject.hasOwnProperty('refPaths') || event == "newProjectAdded") {
         // Send all file watcher clients project related data when a new project is added or ignored paths has changed
         const ignoredPaths = fwProject.ignoredPaths;
```
yangcao77: you cannot append the check here. 
if the event is settingsChanged for `refPaths`, the `ignoredPaths` would be undefined. 
And the update would overwrite the `ignoredPaths` in filewatcherd record and project.inf file.

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
yangcao77: Nevermind for this one. Saw the check for `refPaths` removal.

docs/openapi.yml
makandre: This is for letting filewatcherd know about the external files (outside of the project) to monitor for changes.

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
makandre: The check is needed because the function could be passed an undefined `refPaths` for the removal case

https://github.com/eclipse/codewind/pull/1535/files#diff-2a30a2a4dccd9b8e1dccf2220b27057fR992

src/pfe/portal/modules/FileWatcher.js
makandre: The change here: https://github.com/eclipse/codewind/pull/1535/files#diff-e992ab6a23b4524d0a71d4ee49d71101R116 is meant to ensure that whenever `.cw-settings` or `.cw-refpaths.json` file changes, we would invoke the `projectSpecification` handler with both `ignorePaths` and `refPaths` data, so the `projectSettingsChanged` event emitted back to portal would contain both pieces of data.

docs/openapi.yml
yangcao77: Makes sense. Thanks for the explanation.

docs/openapi.yml
```diff
@@ -1899,6 +1899,11 @@ components:
...
+          description: additional referenced paths to monitor
```
rajivnathan: Should we update the description to mention that it is used for including files that are not within the project folder?

docs/openapi.yml
```diff
@@ -1945,6 +1962,7 @@ components:
           type: string
           example: "Codewind extension for knapp support"
         commands:
+          type: array
```
rajivnathan: Is this an intended change?

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
```diff
@@ -453,6 +453,81 @@ const reconfigIgnoredFilesForDaemon = async function (ignoredPaths: string[], op
...
+            projectInfo.refPaths.every((oldRef) => {
+                return refPaths.findIndex(newRef => newRef.from === oldRef.from && newRef.to === oldRef.to) != -1;
+            })) {
+                logger.logProjectInfo("The ref path list for file watching is unchanged", projectID);
+                return;
+        }
+    }
+    // if undefined and previously it was also undefined, nothing changed
+    else if (!projectInfo.refPaths) {
+        logger.logProjectInfo("The ref path list for file watching is unchanged", projectID);
```
rajivnathan: It's a bit confusing that the lack of the property or if the property is set to `false` or `undefined` or `null` then that means that it's unchanged. Should we do a comparison of the current projectInfo instead?

docs/openapi.yml
rajivnathan: Should we mention that referenced paths are for files that reside outside of the project folder?

docs/openapi.yml
makandre: yes, noticed it was missing

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
makandre: It is unchanged in this case because the old value (`projectInfo.refPaths`) and the new value we are setting (`refPaths`) are both not an array

docs/openapi.yml
makandre: Updated in https://github.com/eclipse/codewind/pull/1535/commits/63672f46ef89d3518536258b818436f19bb52927

src/pfe/portal/modules/FileWatcher.js
yangcao77: Rethink of this one:
The `projectSettingsChanged` does not work like that. Each setting is an independent setting, which means, if multiple setting has been changed, it would emit multiple `projectSettingsChanged` event. And it would only emit if the setting has been changed. 

Even with the change you made https://github.com/eclipse/codewind/pull/1535/files#diff-e992ab6a23b4524d0a71d4ee49d71101R116 , this check would overwrite `ignoredPaths`.

src/pfe/file-watcher/server/src/projects/projectSpecifications.ts
rajivnathan: Ah, fair enough. Think I glanced through it too quickly and thought we were checking `projectInfo.refPaths` in both checks.

src/pfe/portal/modules/FileWatcher.js
makandre: I think I see what you mean.  Will look into it in the morning.

makandre: @yangcao77 thanks for catching this.  I made 2 new commits 99f42bc and 8c51dab and tested it.  Now the `projectSettingsEvent` will really have both `ignoredPaths` and `refPaths` whenever either changes, and both will then be passed to filewatcherd.

FYI @jgwest

yangcao77: The changes look good to me.

Issue #1551, opened 2019-12-16
Project has to be removed and re-added after changing push registry


tetchel:
If you:
1. Have an image push registry
2. Add/Create a project
3. Change your image push registry
4. Build the project

The build fails with the following message:
`Build Failed - The project will not build with the new Image Push Registry. Please remove and re-add the project to deploy to the new Image Push Registry.`

We should support changing the image push registry without this extra step, which is a total pain if you have multiple projects.

maysunfaisal:
This is expected behaviour from the days before we had buildah

but with buildah:

```
[root@codewind-workspaceuxezffqxnzcs1ib9-7c5d8bd95d-8t4rr /]# buildah images
REPOSITORY                                                     TAG                    IMAGE ID       CREATED             SIZE
localhost/cw-turbinetestdefaultsprin-default-spring-15765104   latest                 b1adee3daaa7   38 minutes ago      419 MB
localhost/cw-maylib2-272ada20-2025-11ea-9c6f                   latest                 02766cfc2022   About an hour ago   913 MB
localhost/cw-maylib1-57dff9c0-2020-11ea-9c6f                   latest                 300aaf7e8637   2 hours ago         913 MB
localhost/cw-maygo1-18ab6730-2020-11ea-9c6f                    latest                 38c770833434   2 hours ago         831 MB
<none>                                                         <none>                 ff628fc5085c   2 hours ago         419 MB
localhost/liberty-app-cache                                    latest                 6388fe6a6380   3 hours ago         885 MB
localhost/spring-app-cache                                     latest                 f83ea61bf6b2   3 hours ago         419 MB
docker.io/library/golang                                       latest                 a1072a078890   10 days ago         823 MB
docker.io/library/ibmjava                                      8-sfj                  b27826a6b1cc   2 weeks ago         170 MB
docker.io/library/node                                         8-stretch              c0eda5b3a935   3 weeks ago         923 MB
docker.io/ibmcom/codewind-java-project-cache                   latest                 30c11c49e29d   3 months ago        232 MB
docker.io/library/websphere-liberty                            19.0.0.3-webProfile7   c6892fcebf3c   3 months ago        468 MB
docker.io/library/hello-world                                  latest                 fce289e99eb9   11 months ago       8.03 KB
[root@codewind-workspaceuxezffqxnzcs1ib9-7c5d8bd95d-8t4rr /]# 
```

[...]

PR #1554, opened 2019-12-16
Allow users to update the push registry without having to re-bind their project

johnmcollier:
Fixes https://github.com/eclipse/codewind/issues/1551

This PR removes the restriction that required users to re-bind their project after changing the image push registry in Codewind.

Because we're not tagging the image as @maysunfaisal pointed out, the concern around images with outdated tags sticking around no longer exists, so we can safely remove that restriction.


johnmcollier:
I was able to do the following:
- Set push registry to `us.icr.io/johncollier`

Issue #1245, opened 2019-11-22
Turbine functional tests need to detect iterative development performance regressions


rajivnathan:
**Description of the enhancement:**

The functional tests currently cover creation and update of each of the supported project types. However, it currently cannot detect if performance has degraded. As long a test's project builds and runs the test will pass.

In issue #1222 we saw that on Windows a bug changed the behaviour such that a mvn package was done instead of a mvn compile and this impacted performance severely.

**Proposed solution:** 

A few options:
1. Adjust timeouts to some pre-determined minimum bar for test steps if they're granular enough. This may not be viable depending on variation of test environments (eg. machines with SSD vs without)
2. Monitor the pfe container logs while the tests run and validate that specific messages that indicate the correct code path is being followed and result in a test error otherwise.
3. Write performance focused tests
4. Other options?

rajivnathan:
Discussed this in the Iterative-dev meeting. We'll go with option 1 and capture timings for steps within the existing functional tests. Once we have enough data points we can decide on timeout values on a case by case basis.

elsony:
The other thing that has mentioned is to keep the record of the performance data on each GA release, our benchmark on the test will be based on the data of the last shipped release.

ssh24:
We now have a full suite of Performance Tests working. Closing this as done

/close

PR #1568, opened 2019-12-17
Automated Turbine Performance Test

ssh24:
### Description

Add the ability to run automated performance tests on turbine layer. The performance test can be ran after every DCUT to check against previous release versus this release to see how the performance changed. Running the script will generate a report under `src/pfe/file-watcher/server/test/performance-test/data/<release_name>`

Related to https://github.com/eclipse/codewind/issues/1245


ssh24:
@maysunfaisal Could you PTAL?

PR #1580, opened 2019-12-18
Always tag images with eclipse/

james-wallis:
### Summary
* https://github.com/eclipse/codewind-installer/pull/294 stopped `./run.sh` running the built images and ran `eclipse/` ones instead.
  * Any changes showing in the container were a fluke.

### Testing
* Ran `./run.sh` to ensure it started correctly


james-wallis:
> We should move the contents of publish.sh into the Jenkinsfile as we shouldn't be calling a script from here and it's very small

I've moved the `docker push` into the Jenkinsfile and removed `publish.sh`

ssh24:
@tobespc @stalleyj Yes, there is at least one place that I am aware of that would require the change https://github.com/eclipse/codewind/blob/2a6dfc2378afe0c4f4340ce1c1bff19422d15f06/src/pfe/file-watcher/server/test/cronjob/cronjob-win.sh#L24

Issue #1560, opened 2019-12-17
Improve directory deletion when syncing changes


tobespc:
Currently, when syncing files to pfe, we don't do anything specfic for directories.  As such, if a directory has been removed, we individually delete each file rather than just `rm -fr` on the directory.

james-wallis:
For this piece of work I'll need to make changes in 3 different places:
* Change the `/api/v1/projects/:id/upload/end` API to take a `directoryList` parameter.
* Modify `cwctl` to send a directoryList on the above API call.
* Modify PFE to delete by directory.
  * Delete directories
  * Rescan file system to get updated file system
  * Delete by individual file

james-wallis:
PR raised for CLI https://github.com/eclipse/codewind-installer/pull/301/

james-wallis:
PR raised for PFE https://github.com/eclipse/codewind/pull/1582

james-wallis:
CLI PR merged.

PR #1582, opened 2019-12-18
1560 delete by directory

james-wallis:
PFE part of https://github.com/eclipse/codewind/issues/1560
### Summary
* Added functionality in `upload/end` to delete by directory where possible before re-scanning the file system and deleting the files as we currently do.
  * Required added functionality to ensure only the highest level directory is deleted rather than attempting to delete its sub directories as well.
* Updated the API OpenApi docs to reflect the addition of a `directoryList` in `upload/end`
* Renamed `getFilesToDelete` to `getPathsToDelete` as its now used for files and directories.

### Testing
* Created unit tests for newly introduced functions
* Added directory tests to `path` functions (previously `file` functions)

### DIFF
--- a/src/pfe/portal/routes/projects/remoteBind.route.js
+++ b/src/pfe/portal/routes/projects/remoteBind.route.js
@@ -220,7 +220,7 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
       } else {
         const pathToProj = project.projectPath();
 
-        // // Delete by directory
+        // Delete by directory
         const currentDirectoryList = await recusivelyListDirectories(pathToProj);
         const directoriesToDelete = await getPathsToDelete(currentDirectoryList, keepDirList);
         if (directoriesToDelete.length > 0) {
@@ -228,7 +228,7 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
           const topLevelDirectories = getTopLevelDirectories(directoriesToDelete);
           log.info(`Removing locally deleted directories from project: ${project.name}, ID: ${project.projectID} - ` +
           `${topLevelDirectories.join(', ')}`);
-          await deleteFilesInArray(pathToProj, topLevelDirectories);
+          await deletePathsInArray(pathToProj, topLevelDirectories);
         }
 
         // Delete by file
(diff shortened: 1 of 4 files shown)


src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -260,13 +274,32 @@ router.post('/api/v1/projects/:id/upload/end', async (req, res) => {
...
+      // if there are subdirectories of "dir" already in the topLevelDirArray remove them
+      topLevelDirArray = topLevelDirArray.filter(finalDir => !compareDirs(finalDir, dir));
+      topLevelDirArray.push(dir);
+    }
+  });
+  return topLevelDirArray;
+}
+
+function compareDirs(string, prefix) {
+  string = path.join('/', string, '/');
```
hhellyer: I think this is trying to do something that you should probably do with `path.resolve`. Also could you call the function `isSubDir()` or similar as `compareDirs` doesn't really tell you what your comparing? (The name, the contents?)

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -328,6 +361,23 @@ async function syncToBuildContainer(project, filesToDelete, modifiedList, timeSt
   }
 }
 
+// List all the directories under a given directory
+async function recusivelyListDirectories(absolutePath, relativePath = '') {
```
hhellyer: I think `recusivelyListDirectories` and `listFilesInDirectory` should probably be the same function and should definitely use the same naming scheme. (Probably `recusivelyListDirectories` and `recusivelyListFiles` otherwise we'll have `listDirectoriesInDirectory`.)

hhellyer: Also it's recursively not recusively. ;-)

james-wallis: I've renamed this function and removed the listFilesInDirectory one.

src/pfe/portal/routes/projects/remoteBind.route.js
james-wallis: Path.resolve didn't achieve the same effect. I checked other Path methods and couldn't find a better method than what is already implemented, I've left it as is. Have renamed the function though.

Issue #1508, opened 2019-12-12
SVT:[OCP 4.2]:Performance tests are not available in history to compare results


jagraj:
**Codewind version:** 0.7.0
**OS:** CoreOS

**Che version:** 0.7.0
**IDE extension version:** 0.7.0
**IDE version:**  Theia 7.3.1
**Kubernetes cluster:**
OKD 4.2 cluster

**Description:**
Performance tests results are not stored after running tests for Node.js default project. 
History section does not show any of the test results.

**Steps to reproduce:**
1. Install OKD 4.2 cluster.
2. Install Eclipse che.
3. Create codewind 0.7 workspace
4. Create Node.js project from default templates.
5. Run performance tests.
6. Test results history is blank.

![image](https://user-images.githubusercontent.com/713849/70731462-babfd500-1cd4-11ea-8bdf-a6a4b9fb9d96.png)

![image](https://user-images.githubusercontent.com/713849/70731361-93690800-1cd4-11ea-9021-3a1505a8094c.png)

PFE and Performance logs..

[performance.txt](https://github.com/eclipse/codewind/files/3956904/performance.txt)
[pfe.txt](https://github.com/eclipse/codewind/files/3956905/pfe.txt)

**Workaround:**

jagraj:
I see same problem with appsody projects as well.

markcor11:
I've tried testing the 0.7.0 branch of PFE/Performance/Theia in Che and was unable to reproduce this.  

From the performance container logs, I can see the 3 x 20sec loadtest ran.  All 3 completed but none of the requests were serviced by the project each time:   

**Performance pod reports :** 

```
  totalRequests: 1958, totalErrors: 1958,
  totalRequests: 1966, totalErrors: 1966,
  totalRequests: 1947,  totalErrors: 1947,
```

**PFE pod reports :** 

```
[91m[12/12/19 16:33:35 LoadRunner.js] [ERROR][39m createCollection: Error in function createCollection
[91m[12/12/19 16:33:35 LoadRunner.js] [ERROR][39m { Error: connect ECONNREFUSED 9.30.204.147:30377
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1106:14)
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '9.30.204.147',
  port: 30377 }
```

Was the project responding to browser requests ok between load runs ? 

[...]

PR #1588, opened 2019-12-18
 When in Kube, Loadrunner should use project service to determine endpoint

markcor11:
## Problem: 

Performance dashboard does not work in hybrid as documented in Issue https://github.com/eclipse/codewind/issues/1508. Caused by PFE being unable to access the running project container using http://master:nodeport

## Solution

This PR changes how Codewind communicates with a project container.  Now PFE when running in Kube (including CHE) will : 

Discover project endpoint by querying the project service
Make LoadRunner apply load pressure to project directly rather than through ingress
Make metrics controller gather metrics from project directly rather than ingress

**And in performance dashboard :** 
When PFE is secured, send authentication token to access the protected UISocket

### DIFF
--- a/src/pfe/portal/modules/LoadRunner.js
+++ b/src/pfe/portal/modules/LoadRunner.js
@@ -51,7 +51,6 @@ module.exports = class LoadRunner {
    */
   async createCollection() {
     const metricsContextRoot = this.project.getMetricsContextRoot();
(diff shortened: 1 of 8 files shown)


src/pfe/portal/modules/LoadRunner.js
```diff
@@ -219,12 +237,22 @@ module.exports = class LoadRunner {
       await cwUtils.timeout(5000);
       try {
         // eslint-disable-next-line no-await-in-loop
```
tobespc: can this eslint comment be removed ?

Issue #1576, opened 2019-12-17
SVT: VSCode: Linux: Default Java Microprofile and Java Spring projects failing to build on Ubuntu.


DavidG1011:
**Codewind version:** 0.7.0
**OS:** Ubuntu 18.04.3 LTS

**Description:** Java Microprofile and Java Spring projects are failing to build in VSCode on Ubuntu.

**Steps to reproduce:**
1. Build a default Microprofile or Spring project:
<img width="403" alt="Screen Shot 2019-12-17 at 3 55 44 PM" src="https://user-images.githubusercontent.com/30476587/71033644-bdac3280-20e5-11ea-88f6-dbc24c27e395.png">

Log:
[pfe.txt](https://github.com/eclipse/codewind/files/3975472/pfe.txt)

FYI: @jagraj

ssh24:
Looks like the failure is on:
```
[91m[17/12/19 20:31:48 javamicro] [ERROR][39m executeBuildScript : 399 | docker: Error response from daemon: error while creating mount source path '/codewind-workspace/.logs/javamicro-fb1fe5f0-210b-11ea-b1a1-7f531288df6f': mkdir /codewind-workspace: read-only file system.
```

ssh24:
@DavidG1011 Did you make sure to start VSCode as a root user in Ubuntu?

DavidG1011:
Running as root as far as I know

jagraj:
I am trying on my linux system now with eclipse plugin. Will post the results here.

jagraj:
Looks good on my RHEL 7.2 machine, both default Java and Spring projects built fine. I think this might be related to mount issue with the permissions.

![Screen Shot 2019-12-17 at 4 46 08 PM](https://user-images.githubusercontent.com/713849/71037181-306cdc00-20ed-11ea-9311-b70c5c56d0a6.png)

DavidG1011:
I had a leftover `/codewind-workspace` directory, so I tried removing it. Still running into the same issue:
[pfe_2.txt](https://github.com/eclipse/codewind/files/3975690/pfe_2.txt)

ssh24:
@DavidG1011 Could you please verify this on master?

DavidG1011:
Sorry for the delay. This is fixed on master.

PR #1593, opened 2019-12-18
Remove log directory mounting for liberty and spring projects

ssh24:
### Description

We should not be mounting the log directories for liberty and spring projects anymore because we now support log streaming directly based on workspace or containers. This PR removes the log directory mounting for the two project types. Also note, Node.js or Swift didn't have the log directory mounting and thus was not causing issues.

Related to https://github.com/eclipse/codewind/issues/1576

After the fix, both liberty and spring projects build and start correctly and the logs streaming also works
Screen Shot 2019-12-18 at 1 22 48 PM

### DIFF
--- a/src/pfe/file-watcher/scripts/springScripts/new_entrypoint.sh
+++ b/src/pfe/file-watcher/scripts/springScripts/new_entrypoint.sh
@@ -42,8 +42,6 @@ if [[ "$IN_K8" == "true" ]]; then
     ln -s /codewind-workspace/$PROJECT_NAME /root/app
 fi
 
-# Create symlinks to /root/logs and /root/app on the Liberty app container / pod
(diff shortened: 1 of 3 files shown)


src/pfe/file-watcher/scripts/springScripts/new_entrypoint.sh
```diff
@@ -39,12 +39,13 @@ if [[ "$IN_K8" == "true" ]]; then
         exit 1;
     fi
 
-    # Create symlinks to /root/logs and /root/app on the Liberty app container / pod
-    # Note that $PROJECT_NAME is an env-var defined and set in the liberty app's deployment.yaml
-    mkdir -p /root/logs
     ln -s /codewind-workspace/$PROJECT_NAME /root/app
 fi
 
+# Create symlinks to /root/logs and /root/app on the Liberty app container / pod
```
rajivnathan: This comment needs to be updated

Issue #1530, opened 2019-12-13
Turbine test autoamtion should test if dangling kube resoureces exist


GeekArthur:
Currently Turbine test automation doesn't have capabilities of testing if dangling kube resources (eg. all kube resources deployed by Turbine like deploy, svc, pvc, etc) still exist after project deletion, we should add the capabilities when testing project deletion.

jgwest:
@ssh24: Unable to locate pipeline with name: `in-progress`

PR #1596, opened 2019-12-19
Test for kube/docker resources on create and delete functional tests

ssh24:
Related to https://github.com/eclipse/codewind/issues/1530

Screen Shot 2019-12-19 at 10 02 22 AM


ssh24:
@maysunfaisal Could you PTAL?

src/pfe/file-watcher/server/test/scripts/exec.sh
```diff
@@ -188,7 +188,7 @@ function setup {
 }
 
 function run {
-    TURBINE_EXEC_TEST_CMD="cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml IMAGE_PUSH_REGISTRY_ADDRESS=${REGISTRY_SECRET_ADDRESS} IMAGE_PUSH_REGISTRY_NAMESPACE=${IMAGE_PUSH_REGISTRY_NAMESPACE} npm run $TEST_SUITE:test:xml"
+    TURBINE_EXEC_TEST_CMD="cd /file-watcher/server; JUNIT_REPORT_PATH=/test_output.xml IMAGE_PUSH_REGISTRY_ADDRESS=${REGISTRY_SECRET_ADDRESS} IMAGE_PUSH_REGISTRY_NAMESPACE=${IMAGE_PUSH_REGISTRY_NAMESPACE} CHE_NS=${NAMESPACE} npm run $TEST_SUITE:test:xml"
```
maysunfaisal: Do we need to have a new env `CHE_NS`? Why not use `NAMESPACE`, just dont want there to be a long list of env just for test...

ssh24: `NAMESPACE` is the variable we call in our test scripts. It doesn't really matter what we set it to be called inside our test because we just set the variable to the value of `NAMESPACE` anyways. If you want I can change it `CHE_NS` -> `NAMESPACE`

PR #1608, opened 2019-12-20
Shuffle project types before running functional tests

ssh24:
### Description

Shuffle project types before running functional tests allows us to not be biased on the order of the list of project types we use to run the test. This will run the tests in shuffled order of projects and let's us know that if something wrong with PFE is happening is not specific to the project type.


ssh24:
@maysunfaisal Could you PTALL?

Issue #1613, opened 2019-12-23
Add a background while loop to keep PFE up when running functional test in OKD4


ssh24:
**Description of the enhancement:**
The issue happens in OKD4 where basically kube exec timeout expires while running the test and it just cancels it. We need a way to keep a background process running to keep the exec command running.

/kind test
/area iterative-dev

Issue #1612, opened 2019-12-23
Investigate test failures on kube deployment


ssh24:
**Description:**
The kube deployment checks are failing on functional tests. This is might be because the `includes` function in mocha could be different in checking substring compared to the `includes` function within a string. https://github.com/eclipse/codewind/blob/1fd00c42040c40b7121e8c70d18f2ecd8588274d/src/pfe/file-watcher/server/test/functional-test/lib/utils.ts#L325

/kind test

PR #1618, opened 2019-12-28
Fix kube turbine tests

ssh24:
### Description

This PR fixes the kube deployment resource failures and also adds a background loop to keep PFE alive. The background loop is necessary because kubectl exec has a timeout which exceeds on OKD4.

Closes https://github.com/eclipse/codewind/issues/1612
Closes https://github.com/eclipse/codewind/issues/1613

Screen Shot 2019-12-27 at 7 32 58 PM


ssh24:
@jingfuwang Could you PTAL?

Issue #1563, opened 2019-12-17
Ensure build queue applies to all builds, not just restart of Codewind


rajivnathan:
**Description of the enhancement:**

Related to #1383. There can be significant load when a user creates or updates many projects in a short period of time.

**Proposed solution:** 

The build queue should limit the number of builds that can run at any time. Any type of build (create/update)

ssh24:
Not sure I understand the enhancement properly. What are you proposing?

rajivnathan:
As the proposed solution says "Any type of build (create/update)". The build queue was original implemented for restart scenarios. It currently does not handle update builds. That may be related to the problems seen in #1383

ssh24:
Ah, I see. So currently the build queue logic is for create/recreate scenarios. You are proposing for it to be extended to update scenarios as well

PR #1620, opened 2019-12-31
Allow build queue to apply on update scenarios

ssh24:
### Description

Before we only allowed queuing of builds on turbine restart or project creation scenarios. This PR enhances the build queue by also applying to update scenario. Similar to the create case, update now also runs MAX_BUILDS (i.e 3 builds) concurrently while putting the rest of the projects in the queue.

The screenshot below depicts build queue updating queue ranks as projects are rebuilt
Screen Shot 2019-12-31 at 12 57 33 PM

Closes https://github.com/eclipse/codewind/issues/1563

### DIFF
--- a/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
+++ b/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
@@ -23,6 +23,7 @@ import * as projectSpecifications  from "../projects/projectSpecifications";
 import AsyncLock from "async-lock";
 import * as workspaceSettings from "../utils/workspaceSettings";
 import * as utils from "../utils/utils";
+import { changedFilesMap, IFileChangeEvent } from "../utils/fileChanges";
 const lock = new AsyncLock();
 
 const fileStatAsync = promisify(fs.stat);
@@ -34,7 +35,6 @@ const fileStatAsync = promisify(fs.stat);
  * value: timer
  */
 export const timerMap: Map<string, NodeJS.Timer> = new Map<string, NodeJS.Timer>();
-export const changedFilesMap: Map<string, IFileChangeEvent[]> = new Map<string, IFileChangeEvent[]>();
 export const chunkRemainingMap: Map <string, ChunkRemainingMapValue[]> = new Map<string, ChunkRemainingMapValue[]>();
 
 const workspaceSettingsInfo =  workspaceSettings.workspaceSettingsInfoCache ? JSON.parse(workspaceSettings.workspaceSettingsInfoCache) : undefined;
@@ -304,13 +304,7 @@ export interface IUpdateProjectFailure {
     error: { msg: string };
 }
 
-export interface IFileChangeEvent {
-    path: string;
-    timestamp: number;
-    type: string;
-    directory: boolean;
-    content?: string;
-}
+
 export interface ChunkRemainingMapValue {
     timestamp: number;
     chunkRemaining: number;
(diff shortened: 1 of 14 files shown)


ssh24:
@rajivnathan Could you please review and merge this after you are back from vacation? I will be away starting Jan 6th by the time you will be back, so feel free to land it

josephca:
@ssh24 are we planning to merge this PR sooner or later? it's been sitting for a while and using some Jenkins resources (50 builds as of today)

ssh24:
It should be going in soon. Testing out some changes and after review, it should be good to land

ssh24:
@rajivnathan The PR is ready to be reviewed now. Could you PTAL?

src/pfe/file-watcher/server/src/controllers/projectsController.ts
```diff
@@ -27,6 +27,7 @@ import * as projectUtil from "../projects/projectUtil";
 import * as logHelper from "../projects/logHelper";
 import * as logger from "../utils/logger";
 import * as statusController from "./projectStatusController";
+import * as projectEventsController from "./projectEventsController";
```
rajivnathan: This introduces a circular dependency since projectEventsController imports projectsController as well. Maybe we should move the handling of `changedFilesMap` to a separate module that can then be imported by both.

src/pfe/file-watcher/server/src/controllers/projectEventsController.ts
```diff
@@ -212,7 +212,13 @@ export async function updateProjectForNewChange(projectID: string, timestamp: nu
                     if (projectInfo.autoBuildEnabled) {
                         if (!statusController.isBuildInProgressOrQueued(projectID)) {
                             const operation = new projectOperation.Operation("update", projectInfo);
-                            projectHandler.update(operation, changedFilesMap.get(projectInfo.projectID));
+                            const project: projectsController.BuildQueueType = {
+                                operation: operation,
+                                handler: projectHandler
+                            };
+                            // for update operation
+                            await statusController.updateProjectStatus(statusController.STATE_TYPES.buildState, projectID, statusController.BuildState.inProgress, "action.calculateDifference");
```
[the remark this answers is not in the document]
ssh24: Sure, I am fine with an empty message

src/pfe/file-watcher/server/src/controllers/projectsController.ts
ssh24: I don't think there is a "real" circulation here. Yes, both modules call each other but circular dependency is not just two module calling one another but also their initialization, calls and etc. I was able to verify that by kicking off our test suite which verifies and catches circular dependencies during runtime of the filewatcher module

rajivnathan: While the tests may not have caught any circular dependencies I'm not sure that means there aren't any problems since our test coverage is not 100%. Also, if we keep it this way then in the future it would be easier for others to introduce code that does cause problems without realizing it since they could just use the already imported module. The best way to prevent future problems would be to avoid modules importing each other.

ssh24: I agree with you. I have no problem with moving that into a separate module and that is the correct way to do it to avoid future issues. But just wanted to point out there are other places in the code where we do have circular module calls like this and it works because the circular calls are not full, meaning it does not have calls to function during initialization etc. If they do have real circular calls, they would be caught by the test not because of the coverage but because of how Node works - all modules are preloaded during runtime and if Node detects any circular dependencies on initialization, it will blow up.

Issue #1611, opened 2019-12-20
Jenkins build does not run Portal Integration Tests


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
I have noticed on Dec 20 that Jenkins PR builds do not run Portal's integration tests.

I am not sure how long this is broken, it means we've been checking in PRs without running it against the integration tests.

From one of the PRs

```
13:58:49  +++   TURBINE UNIT TESTS COMPLETED SUCCESSFULLY   +++
[Pipeline] }
[Pipeline] // withDockerRegistry
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // timeout
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Run Codewind test suite)
[Pipeline] timeout
13:58:49  Timeout set to expire in 2 hr 0 min
[Pipeline] {
[Pipeline] withEnv
[Pipeline] {
[Pipeline] withDockerRegistry
13:58:50  $ docker login -u codewindbot -p ******** https://index.docker.io/v1/
13:58:50  WARNING! Using --password via the CLI is insecure. Use --password-stdin.
13:58:51  WARNING! Your password will be stored unencrypted in /home/root/workspace/Codewind_codewind_PR-1588@tmp/13de9e5f-fb04-4501-bf42-4c0821a89533/config.json.
13:58:51  Configure a credential helper to remove this warning. See
13:58:51  https://docs.docker.com/engine/reference/commandline/login/#credentials-store
13:58:51  
13:58:51  Login Succeeded
[Pipeline] {
[Pipeline] sh
13:58:53  Starting tests for Eclipse Codewind ...
13:58:53  
13:58:53  
13:58:53  Platform: x86_64 
13:58:53    % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
13:58:53                                   Dload  Upload   Total   Spent    Left  Speed
13:58:53  
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 13226  100 13226    0     0   516k      0 --:--:-- --:--:-- --:--:--  516k
13:58:53  => nvm is already installed in /root/.nvm, trying to update using git
13:58:53  
=> => Compressing and cleaning up git repository
13:58:53  
13:58:53  => nvm source string already in /root/.bashrc
[...]

PR #1632, opened 2020-01-06
Remove docker-compose logic

tobespc:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -159,7 +159,8 @@ pipeline {
                         sh '''#!/usr/bin/env bash
                         echo "Starting tests for Eclipse Codewind ..."
                         export PATH=$PATH:/home/jenkins/.jenkins/tools/jenkins.plugins.nodejs.tools.NodeJSInstallation/node_js/bin/
-
+                        mkdir -p $HOME/dc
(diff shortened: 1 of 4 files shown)


james-wallis:
Fix for https://github.com/eclipse/codewind/issues/1611
### Summary
* Remove the `docker-compose` logic for starting PFE and replace it with `start.sh` (now use `cwctl` to start PFE in Jenkins.
  * Had to move the `docker-compose` install location to `/usr/bin` so it is on the path. I've added cleanup at the bottom of the Jenkinsfile.
  * Also added a check to ensure that the image PFE is using for the tests is the one that is built in Jenkins (ensuring `cwctl` does not pull down a new image and use that).
* Added exit code checking to the linting tests and unit tests.
* Split the Portal unit tests and API tests up so the unit tests can run before PFE is started (We'll be able to get unit tests failures quicker to improve work efficiency.
* Added an exit code to `start.sh` if a container exits, Jenkinsfile uses this to see if Codewind has started with no errors.
* Removed tests from `remoteBind.test.js` as the function `fileIsProtected` had been deleted in a previous PR.

### Testing
* Ensured green build in Jenkins with latest master.

PR #1636, opened 2020-01-06
Move Turbine unit test logic to main cron script

maysunfaisal:
This PR moves Turbine unit test logic to our main cron script. The test suites had their own cron since the pivot to Codewind and was a lot of dup code and maintenance.


maysunfaisal:
@rajivnathan pls help review when you have a chance.. I just cleaned up some of our tests as there was some duplication and separate cron scripts for unit tests. Now both functional and unit run from the same cron script with this PR.

I have nightly tests running against this branch for now.

src/pfe/file-watcher/server/test/test.sh
```diff
@@ -216,8 +226,12 @@ if [[ $TEST_TYPE == "kube" ]]; then
...
-else
-    run
+if [ $TEST_SUITE == "functional" ]; then
+    if [[ ($CLEAN_RUN == "y") ]]; then
+        cleanRun
+    else
+        run
+    fi
+elif [ $TEST_SUITE == "unit" ]; then
+    executeTests
```
rajivnathan: What's the difference between `executeTests` and `run`? Do we need both?

maysunfaisal: updated

maysunfaisal:
@rajivnathan can you pls help review the update? I cleaned up the script further based on your comment.

Issue #1055, opened 2019-11-08
No response body if error occurs in bind start/end


makandre:
**Codewind version:** 0.6.0
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**

First spotted as part of:
https://github.com/eclipse/codewind/issues/1036#issuecomment-551326911

In the `bind/start` and `bind/end` APIs, if an error occurs portal returns status `500` but it doesn't look like it sends a response body.

This is possibly masking the cause of the error and the reason we saw the `EOF` in #1036

**Steps to reproduce:**

**Workaround:**

makandre:
Looks gd

PR #1638, opened 2020-01-07
Return an error response if project bind start or end fails

markcor11:
## Problem 

If Project Bind Start or End fails portal will return an HTTP 500 response but without any meaningful data. As reported in issue : #1055 

## Solution

Continue to set the HTTP status code to 500 but in addition include an error in the body making these two functions consistent with the other project bind routes. 

### DIFF
--- a/src/pfe/portal/routes/projects/remoteBind.route.js
+++ b/src/pfe/portal/routes/projects/remoteBind.route.js
@@ -116,7 +116,11 @@ async function bindStart(req, res) {
     if (err.code === 'INVALID_PROJECT_NAME'){
       res.status(400).send(err.info);
     } else {
-      res.sendStatus(500);
+      let errorMessage = ""
+      if (err) {
+        errorMessage = err.info || err
+      }
+      res.status(500).send(`Project bind failed ${errorMessage}`);
     }
     log.error(err);
     return;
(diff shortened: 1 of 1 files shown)


src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -116,7 +116,11 @@ async function bindStart(req, res) {
     if (err.code === 'INVALID_PROJECT_NAME'){
       res.status(400).send(err.info);
     } else {
-      res.sendStatus(500);
+      let errorMessage = ""
+      if (err) {
+        errorMessage = err.info || err
+      }
+      res.status(500).send(`Project bind failed ${errorMessage}`);
```
rwalle61: nit: I think this is equivalent to:
```suggestion
      res.status(500).send(`Project bind failed ${err.info || err}`);
```

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -142,6 +146,7 @@ async function bindStart(req, res) {
       error: err.info || err
     }
     user.uiSocket.emit('projectBind', data);
+    res.status(500).send(`Project creation failed for project ${data.name}. Error: ${data.error}`);
```
hhellyer: This is called after we will have finished with the `res` object and sent `202`. Here the response can only be reported by the socket.

Issue #242, opened 2019-08-21
Improve the Jenkinsfile as per comments in PR 230


tobespc:
- [ ] Is there an environment variable for the codewind directory that we can use to change directory instead of relative paths? If not we can create one before cd src/pfe/portal. It'd help make the Jenkinsfile more robust in case other changes are added in between in the future.

- [ ] Check uses of -eq instead of -gt?

- [ ] Think we need the -a flag to find exited containers and in the other uses of docker ps below?

PR: https://github.com/eclipse/codewind/pull/230

james-wallis:
/assign

I'm working on the Jenkinsfile anyway so I'll pick these up.

james-wallis:
I've completed all the items in this issue, @tobespc can you close please.

PR #1658, opened 2020-01-08
Run eslint and unit tests before Docker build in Jenkins, remove directory traversal

hhellyer:
Save the base directory to ensure we can return to it in the Jenkinsfile.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,7 @@ pipeline {
     }
 
     stages {
-        stage('Run Portal eslint and Unit Test Suite') {
+        stage('Run Portal eslint and unit tests') {
             options {
                 timeout(time: 30, unit: 'MINUTES') 
             }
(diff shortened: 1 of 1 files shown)


james-wallis:
### Summary
* Move portal and turbine eslint and unit tests to above the Docker build. (exposed https://github.com/eclipse/codewind/issues/1673)
  * Means the PR build will fail sooner if eslint or a unit test fails so we can get faster feedback
  * Additionally we will stop running `docker build`s for Jenkins builds which are destined to fail anyway
* Ensure that all directory traversal commands in the Jenkins built use a `$DIR` variable to stop errors in the future
  * Fix for the final bullet point in https://github.com/eclipse/codewind/issues/242

### Testing
* Ensured the Jenkins build passes

Jenkinsfile
```diff
@@ -15,6 +15,112 @@ pipeline {
...
+                        if [ $? -ne 0 ]; then
+                            exit 1
+                        fi
+
+                        npm run eslint
+                        if [ $? -ne 0 ]; then
+                            exit 1
+                        fi
+
+                        cd $DIR
```
jcockbain: is this used now?

james-wallis: 🥇

james-wallis: Removed

PR #1671, opened 2020-01-09
Revert "Return error response if unbind fails (#1630)"

markcor11:
Reverts this change since a response may have already been dispatched. If an error occurs after the response is sent it would be too late to send back anything other than the socket message.


tobespc:
Force merging as the PR is raised as a revert and original committer has signed the agreement

Issue #1491, opened 2019-12-11
Initial 'templates repos get' is very slow on kube


tetchel:
```
[INFO: 17:22:56.171 CLIWrapper.js:129]: Running CLI command: cwctl --json --insecure templates repos list --conid K41SXS5Z
<a bunch of other stuff happens>
[DBUG: 17:23:18.907 CLIWrapper.js:192]: Successfully ran CLI command cwctl --json --insecure templates repos list --conid K41SXS5Z, Output was:
[
	{
		"description": "Codewind project templates help you create containerized projects for various runtimes.",
		"url": "https://raw.githubusercontent.com/codewind-resources/codewind-templates/master/devfiles/index.json",
		"name": "Default templates",
		"id": "",
		"enabled": true,
		"protected": true,
		"projectStyles": [
			"Codewind"
		]
	},
	{
		"description": "The set of templates for new OpenShift projects in Codewind.",
		"url": "file:///codewind-workspace/.extensions/codewind-odo-extension/templates/index.json",
		"name": "OpenShift templates",
		"id": "",
		"enabled": true,
		"protected": true,
		"projectStyles": [
			"OpenShift"
		]
	},
	{
		"description": "Use Appsody in Codewind to develop applications with sharable technology stacks.",
		"url": "https://github.com/appsody/stacks/releases/latest/download/incubator-index.json",
		"name": "Appsody Stacks - incubator",
		"id": "incubator",
		"enabled": true,
		"protected": true,
		"projectStyles": [
			"Appsody"
		]
	}
]
```

You can see the initial templates sources GET took 22 seconds!

tetchel:
Applies to che too

tobespc:
The initial pull is always slow, but is that a bug, a lot of git queries are happening at that point?  I think we should use this to investigate the reasons why it's slow.   It does seem a lot of the time that local infrastructure has a huge impact on this first pull.  Lots of things are slow on the first startup of codewind ;-)

Moving this to investigate task

tetchel:
it looks like a bug to a user, maybe it could be done as soon as the server starts instead of the first time we GET /templates/repositories

jopit:
We can debate whether this is a bug or not, but it is a bad user experience.  @tobespc can you assign someone to investigate how we can improve it?

tobespc:
[...]

PR #1705, opened 2020-01-10
Cache template repositories and templates at startup.

hhellyer:
This PR fixes the caching of template repositories at start up. Previously we were doing it when the user object was created but before we loaded extensions which then invalidated the cached repository.
The caching has been moved after the extensions are loaded, we also now cache the list of templates from those repositories.

This should improve performance for https://github.com/eclipse/codewind/issues/1491 as the slowness will be moved to startup. We don't add the routes to access the REST API until after the User object that populates the cache on initialisation is ready so the repositories and templates can't be queried before we have cached them. (There may still be a delay if the list of repositories is changed.)

Additionally there was a bug where the flag to getTemplates to say whether to show enabled or disabled templates was only honoured if we didn't have a cached list of templates. If a list was cached we always returned the list that was generated on the first call, even if the value of the flag was different. This PR fixes that and adds test cases to verify that the flag is honoured.

### DIFF
--- a/src/pfe/portal/modules/User.js
+++ b/src/pfe/portal/modules/User.js
@@ -108,7 +108,7 @@ module.exports = class User {
       await this.templates.initializeRepositoryList();
       log.info(`Initialising template list`);
       await this.templates.getTemplates(false);
-      log.info(`Default templates cached.`);
+      log.info(`Default templates cached`);
 
       // Connect up the UI Socket Authentication handler
       if (process.env.CODEWIND_AUTH_HOST) {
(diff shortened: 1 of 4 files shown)


src/pfe/portal/modules/User.js
```diff
@@ -102,6 +101,15 @@ module.exports = class User {
...
       }
 
+      // Initialise the list after we have created extensions as they may
+      // add template repositories. The log messages allow us to confirm
+      // how long this takes.
+      log.info(`Initialising template repository list`);
+      await this.templates.initializeRepositoryList();
+      log.info(`Initialising template list`);
+      await this.templates.getTemplates(false);
+      log.info(`Default templates cached.`);
```
james-wallis: Massive nit: `Default templates cached.` has a full stop when the other `log`s don't.

PR #1725, opened 2020-01-13
fixed dependency vulnerabilities

NikCanvin:
ran 'npm audit fix --force' (used 'force' because several dependencies required major version updates to get the fixes)..

then rebuilt and ran Codewind with the changes, I saw ExpressNode and OpenLiberty Java applications both continue to work, the application metrics still display on the monitor..


NikCanvin:
just checked the legals of the changed packages, all looks fine!

Issue #1522, opened 2019-12-13
[Remote/Kube] Can't have projects which share the same first three letters


james-wallis:
**Codewind version:** Master
**OS:** Using VSCode extension on Mac OS connecting to a remote deployment in OpenShift.

**Description:**
If I have two projects which share the same first 3 characters in their name ("app" for example) I can only reach the first application using the ingress address.

I have created two applications using remote mode into an open shift cluster:
1. `appsody-node-express-mongrel`
2. `appsody-scaffold-mongrel`
They both build and launch fine.
The Ingress url given to me for each is:
1. `http://cw-app-jamesw0001.apps.exact-mongrel-icp-mst.9.20.195.90.nip.io/`
2. `http://cw-app-jamesw0001.apps.exact-mongrel-icp-mst.9.20.195.90.nip.io/`

So they're exactly the same and when I go to the URL I can only get to the first project.
I am not sure on the details but it looks like we're using `cw-app` as the only identifier for the ingress.

*This might also be present in a Che environment but I have not tested for that.*

My OpenShift dashboard shows the second project's URL as invalid as the route it is trying to use already exists (the black invalid url text):
![image](https://user-images.githubusercontent.com/17385115/70805648-56267800-1db1-11ea-9d6f-0ffd7ef24bcc.png)

**Steps to reproduce:**
1. Connect to remote Codewind
2. Create two projects with the same first 3 characters in their name (I did `appsody-*`)
    * I created Appsody and Node.js
3. Let them build and start
4. Check their given URLs and they will be the same
5. Check the OpenShift dashboard to see the behaviour shown in the screenshot above.

**Workaround:** 

If the project has an exposed port AND its a NodePort you can take any URL for that OpenShift cluster and add its port onto the end.
So: 
  * `http://cw-app-jamesw0001.apps.exact-mongrel-icp-mst.9.20.195.90.nip.io: 32688` got me to my first Node.js project
  * `http://cw-app-jamesw0001.apps.exact-mongrel-icp-mst.9.20.195.90.nip.io: 31802` got me to my second Node.js project

johnmcollier:
@james-wallis I know why this is happening, it's cause we're trimming the ingress hostname we create in a bad way.

On OKD 3.x clusters, it requires that the host name for the ingress/route be < 63 characters. The problem here is that we have no way to know what the maximum allowed ingress host name length is, as it heavily varies by Kube distribution (even OCP 4.x doesn't have this issue). So what we do is:
[...]

PR #1736, opened 2020-01-13
Prevent collision of ingresses

johnmcollier:
Fixes https://github.com/eclipse/codewind/issues/1522

This PR updates the ingress creation logic to prevent collisions from occurring when the ingress needs to be trimmed due to character limits:
- New ingress URL format for projects: `<4-random-alphanumeric-characters>-<projectname>-<ingressdomain>` 
- We now trim the ingress if: the ingress domain is < 62 characters **and** the resultant project ingress is over that limit
     - This is a slight change as was proposed in earlier meetings (append random characters to the end of the project name, and trim right to left). 
     - This is because we allow non-alphanumeric characters in our project names and ingress URLs **must** start with an alphanumeric character
     
Since we're using the project name in the ingress, I noticed that the `projectName` variable for generic docker projects was `undefined` and it's because `projectInfo.projectName` wasn't set. So to fix it, I used `projectLocation.split("/").pop()`, which is how we get it elsewhere in Turbine.

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/projectUtil.ts
+++ b/src/pfe/file-watcher/server/src/projects/projectUtil.ts
@@ -425,7 +425,7 @@ async function executeBuildScript(operation: Operation, script: string, args: Ar
                         const servicePort = parseInt(containerInfo.internalPort, 10);
 
                         // Expose an ingress for the project
-                        const baseURL = await kubeutil.exposeOverIngress(projectID, operation.projectInfo.isHttps, servicePort);
+                        const baseURL = await kubeutil.exposeOverIngress(projectID, projectName, operation.projectInfo.isHttps, servicePort);
 
                         // Set the appBaseURL to the ingress we exposed earlier
                         projectInfo.appBaseURL = baseURL;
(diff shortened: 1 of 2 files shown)


johnmcollier:
@rajivnathan @elsony Please review when you get a chance, thanks!

johnmcollier:
Verified trimming now works. Two projects `my-go-project` and `my-go-test` each with separate ingresses:
- http://6g1o-my-go--codewind.codewind.apps.codewind-ocp.os.fyre.ibm.com/
- http://jsj2-my-go--codewind.codewind.apps.codewind-ocp.os.fyre.ibm.com/

src/pfe/file-watcher/server/src/projects/projectUtil.ts
```diff
@@ -1857,7 +1857,7 @@ async function getPODInfoAndSendToPortal(operation: Operation, event: string = "
     const projectInfo = operation.projectInfo;
     const projectLocation = projectInfo.location;
     const projectID = projectInfo.projectID;
-    const projectName = projectInfo.projectName;
+    const projectName = projectLocation.split("/").pop();
```
rajivnathan: Do we know why projectInfo.projectName was undefined here? Wondering whether projectInfo was not updated somewhere before this step.

johnmcollier: Not sure why, to be honest.

rajivnathan: As discussed in the meeting we'll go forward with this change since projectName is mainly used as an identifier for users that do a `kubectl get po` etc. but for Codewind it makes little difference since labels are used for querying resources.

PR #1775, opened 2020-01-16
Made changes to make it work on ppc64le systems

vrushaliinamdar:
Made following changes to make it work on ppc64le systems:

1. Created ppc64le specific Dockerfiles for pfe and performance components
2. Changed build.sh to disable codewind-appsody-extension for ppc64le

Signed-off-by: Vrushali Inamdar vrushali.inamdar@ibm.com

### DIFF
--- a/script/build.sh
+++ b/script/build.sh
@@ -61,9 +61,20 @@ cp -r $DIR/docs ${SRC_DIR}/pfe/portal/
 echo -e "\n+++   DOWNLOADING EXTENSIONS   +++\n";
 mkdir -p ${SRC_DIR}/pfe/extensions
 rm -f ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-*.zip
+# In case of ARCH = ppc64le, do not download the zip file.
+if [ "$ARCH" == "ppc64le" ]; then
+   echo -e "\n+++++   DISABLING APPSODY EXTENSION FOR PPC64LE ARCHITECTURE +++++\n";
+else
+   curl -Lo ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-9.9.9999.zip http://download.eclipse.org/codewind/codewind-appsody-extension/master/latest/codewind-appsody-extension-9.9.9999.zip
+fi
+
 rm -f ${SRC_DIR}/pfe/extensions/codewind-odo-extension-*.zip
-curl -Lo ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-9.9.9999.zip http://download.eclipse.org/codewind/codewind-appsody-extension/master/latest/codewind-appsody-extension-9.9.9999.zip
 curl -Lo ${SRC_DIR}/pfe/extensions/codewind-odo-extension-9.9.9999.zip http://download.eclipse.org/codewind/codewind-odo-extension/master/latest/codewind-odo-extension-9.9.9999.zip
+
+# In case of ppc64le, download 'appsody' using ppc64le specific url and build 'appsody-controller' from source
+if [ "$ARCH" == "ppc64le" ]; then
+  $DIR/script/build-appsody-controller.sh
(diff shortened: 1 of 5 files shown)


src/keycloak/Dockerfile
```diff
@@ -9,7 +9,8 @@
 #     IBM Corporation - initial API and implementation
 ################################################################################
 
-FROM jboss/keycloak:8.0.0
+FROM jboss/keycloak:latest
```
makandre: I don't think we always want to build with latest keycloak, we had problems in the past where we picked up some unwanted breaking changes.

Is there a newer version of keycloak that you require?

Also, should this change be in a `ppc64le` version of the dockerfile?

You may want to discuss this part with @markcor11

src/gatekeeper/Dockerfile
```diff
@@ -9,7 +9,7 @@
 #     IBM Corporation - initial API and implementation
 ################################################################################
 
-FROM node:12-alpine
+FROM ppc64le/node:12-alpine
```
makandre: This is changing the main dockerfile that would affect all platforms

script/build.sh
```diff
@@ -61,10 +61,21 @@ cp -r $DIR/docs ${SRC_DIR}/pfe/portal/
...
+   curl -Lo ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-9.9.9999.zip http://download.eclipse.org/codewind/codewind-appsody-extension/master/latest/codewind-appsody-extension-9.9.9999.zip
+fi
+
 rm -f ${SRC_DIR}/pfe/extensions/codewind-odo-extension-*.zip
-curl -Lo ${SRC_DIR}/pfe/extensions/codewind-appsody-extension-9.9.9999.zip http://download.eclipse.org/codewind/codewind-appsody-extension/master/latest/codewind-appsody-extension-9.9.9999.zip
 curl -Lo ${SRC_DIR}/pfe/extensions/codewind-odo-extension-9.9.9999.zip http://download.eclipse.org/codewind/codewind-odo-extension/master/latest/codewind-odo-extension-9.9.9999.zip
 
+# In case of ppc64le, download 'appsody' using ppc64le specific url and build 'appsody-controller' from source
+if [ "$ARCH" == "ppc64le" ]; then
+  $DIR/script/build-appsody-controller.sh
```
makandre: What does this do?  This PR does not include this script.

src/keycloak/Dockerfile
vrushaliinamdar: The script picks up Dockerfiles for pfe and performance specific to architecture. For gatekeeper and keycloak, I dont find any such logic, hence changed the same Docker file.

script/build.sh
vrushaliinamdar: removed.

src/keycloak/Dockerfile
vrushaliinamdar: Changed tag for keycloak image.

src/gatekeeper/Dockerfile
hhellyer: The fix is probably to rename this file `Dockerfile_x86_64` and create a new `Dockerfile_ppc64le` with this change in.
The `src/gatekeeper/build` script already gets passed an argument with the dockerfile name:
https://github.com/eclipse/codewind/blob/6a7590a49dd5979599a46c271c7b811557e8d1d7/script/build.sh#L77
it's just ignoring it. It should use it as the --file argument like pfe:
https://github.com/eclipse/codewind/blob/ae2e9051f23e3b0270c2cacbc0966e7da9cb173e/src/pfe/build#L33

(The same should probably be done for the keycloak container.)

vrushaliinamdar:
Another PR https://github.com/eclipse/codewind/pull/1983 is raised, hence closing this.

PR #1799, opened 2020-01-17
Fix Microrprofile, Spring Cache and Microprofile entrypoint

maysunfaisal:

Issue #1062 

This PR fixes caching for Local Microprofile. Since we moved to the Hybrid architecture, local Microprofile has been intermittently failing. Turbine tests have been catching it but not users.. Instead of saving the cache in the project root dir, we have moved them to `/tmp`

The same has been done for Spring on local

### DIFF
--- a/src/pfe/file-watcher/scripts/spring-container.sh
+++ b/src/pfe/file-watcher/scripts/spring-container.sh
@@ -18,7 +18,6 @@ CONTAINER_NAME=$5
 logName=$7
 START_MODE=$8
 DEBUG_PORT=$9
-MAVEN_M2_CACHE=.m2/repository
 FOLDER_NAME=${11}
 IMAGE_PUSH_REGISTRY=${12}
 MAVEN_SETTINGS=${13}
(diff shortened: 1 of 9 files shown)


maysunfaisal:
@elsony can you pls help review this PR, I've updated the Spring like Microprofile for Master

src/pfe/file-watcher/scripts/spring-container.sh
```diff
@@ -75,8 +74,8 @@ function create() {
 		echo "Stop watching flag found. Doing nothing.";
 	else
 		# If the maven m2 cache doesn't exist then pull it from dockerhub
-		if [ ! -d $MAVEN_M2_CACHE ]; then
-			if [ "$IN_K8" == "true" ]; then
+		if [ "$IN_K8" == "true" ]; then
```
elsony: Is there a reason why we remove the code of checking if the .m2 repo already exist?  If the m2 repo cache already there, is there a reason why we want to pull the cache again?

maysunfaisal: `spring-container.sh` will be executed in PFE and `mvn` is going to run on the spring app container with `.m2` dir in the app container. Since mount is removed from local but is there only for Kube, I thought it made sense to do the checking the other way `if [ ! -f localm2cache.zip ]; then` which is what `liberty-container.sh` is doing right now. What do you think?

elsony: Do we always clean up the zip file after expand?  Will we run into cases where we constantly unzip that file even after the cache has already be created?

maysunfaisal: yes in Spring we delete after unzipping https://github.com/eclipse/codewind/pull/1799/files#diff-5d6e8453c454a38f03af492cf1c98808R29

in liberty too https://github.com/eclipse/codewind/pull/1799/files#diff-ba1eb71263eb360db01ec49381ae4283R38 and https://github.com/eclipse/codewind/pull/1799/files#diff-b031023b67c621fd8c6925c0affbf38eR33

Issue #1808, opened 2020-01-20
Enhance Portal Code Coverage 


james-wallis:
**Description of the enhancement:**

Currently in Portal we have the ability to get the code coverage for our unit tests only. This gives us the appearance that our coverage is very low whereas we touch a lot of the other code though our API tests.

**Proposed solution:** 

I've created a development branch which gives us the ability to:
* Run PFE using `nyc` to get information about the files that are touched during the API Tests.
* Collect the generated coverage data out of a Docker container.
* Combine this file with a coverage file generated from the Unit tests (or local run tests).
* Display the coverage as HTML (or any other specified reporter).
Branch: https://github.com/eclipse/codewind/compare/master...james-wallis:get-coverage-from-container?expand=1

This branch needs tidying and testing as it was only designed to be an MVP.
The thought for this is in the future it'll be able to be shown on Jenkins (https://wiki.jenkins.io/display/JENKINS/HTML+Publisher+Plugin).

Design decisions:
* Whether the coverage should be apart of the `test.sh` or separate?

Note:
* One this feature has been completed it would be possible to combine it with the Turbine files to show the code coverage there (and report to Jenkins).

james-wallis:
This has been delivered.

PR #1816, opened 2020-01-20
Get code coverage from container (api tests) and merge with local coverage (unit tests)

james-wallis:
Implementation for https://github.com/eclipse/codewind/issues/1808
### Summary
* Provide a script which can stop PFE and get the code coverage from the API tests as long as it is run with the environment variable `CODE_COVERAGE_ENABLED` so that PFE starts using `npx nyc`  meaning `nyc` can track the files that are hit during PFE's running.
* Made it so that the script is called during `test.sh`.
  * can call the script manually but `codewind-pfe` needs to be running. 
* Added a new way to set the `NODE_ENV` and `CODE_COVERAGE_ENABLED` env vars for the PFE container by setting `PFE_DEV_NODE_ENV` and `PFE_CODE_COVERAGE` variables respectively.
  * This is an alternative to the `devbuild.env` and will take priority over it. Means we can set the `NODE_ENV` for PFE in scripts.
* Changed the `docker logs` in `test.sh` to be outputted to a file so that the coverage can actually been seen and so `docker logs` doesn't take up huge amounts of the terminal.

### Testing
* Ran the script multiple times

### Packages
* Added our log package to the `test` `package.json`
* Updated `nyc` to version 15.0.0.
* Added nyc@15.0.0 into the `npm-start.sh` script.


src/pfe/Dockerfile_x86_64
```diff
@@ -148,8 +148,10 @@ COPY /portal/NOTICE.md /
 WORKDIR /portal
 ARG NODE_ENV
 ARG IMAGE_BUILD_TIME
+ARG ENABLE_CODE_COVERAGE
```
hhellyer: Would it clear up some of the code below in `src/pfe/build` if we used:
```
ARG ENABLE_CODE_COVERAGE=false
```
to set a default value?

test.sh
```diff
@@ -83,9 +87,17 @@ npm run test
 rc=$?;
 cd ..
 
+# Run coverage
```
hhellyer: It's a total nit but we already ran with coverage. It would be nice if this comment explained we are merging the coverage from the unit tests and the REST tests.

src/pfe/Dockerfile_x86_64
james-wallis: Have to keep it as is due to the `--build-arg` being passed in overwriting the defaults in the Dockerfile. I've added the default anyway so in the future if we build the Dockerfile without the `--build-arg` flag we won't error out.

PR #1827, opened 2020-01-21
Move installBuiltInExtensions out of the ExtensionList Class

james-wallis:
### Summary
* Move `installBuiltInExtensions` (and the functions only it uses) out of `ExtensionList.js`.
  * Removes eslint error as it does not interact with the `this` variable.
  * Also due to it not doing anything with `Extension` objects.
  * Its moved into the `utils` function so that it can be called like a script.
* Changed some functionality so that we don't use an `exec('grep` and instead load the yaml into a JS object and get the version that way (couldn't test as the `grep` command wouldn't work on my mac).
  * Did testing in the Docker container to ensure (to the best of my knowledge) that the implementation is the same as before.
* Added tests for the moved functions.

### Testing
* Ran the unit tests
* Ensured that Codewind started and initialised the extensions on my disk (with an empty Docker volume so it doesn't use a previous extension version).


james-wallis:
The build for this PR is failing as Jenkins does not have the `zip` and `unzip` packages installed. 
I'll either: 
1. Get the packages installed on Jenkins
2. Skip the tests which make it to the `unzip` stage of the function on Jenkins.

james-wallis:
@josephca has raised a bug with Eclipse to get `zip` and `unzip` installed on our build machines: https://bugs.eclipse.org/bugs/show_bug.cgi?id=559419.

For now I'm going to programatically skip the tests in Jenkins using an environment variable.

Issue #1779, opened 2020-01-16
SVT: [Che] Error deleting project


sujeilyfonseca:
**Codewind version:** 0.8.0
**OS:** Red Hat Enterprise Linux 7 (64 bit)

**Che version:** 7.5.1
**Kubernetes cluster:** OKD/Openshift 

**Description:** An error message "Error deleting project" appears in Eclipse Che, but then the project is removed. I have seen this error with multiple projects.

![Screen Shot 2020-01-16 at 11 54 16 AM](https://user-images.githubusercontent.com/49531960/72546666-30c9d580-3859-11ea-8f4e-fb18901eefb7.png)
![Screen Shot 2020-01-16 at 11 53 57 AM](https://user-images.githubusercontent.com/49531960/72546667-30c9d580-3859-11ea-8169-86e9959e05a5.png)
![Screen Shot 2020-01-16 at 11 36 11 AM](https://user-images.githubusercontent.com/49531960/72546668-30c9d580-3859-11ea-9700-911fae6f6ce4.png)

**Logs:**
[codewind-workspacel3sjxgfbqx5n3tp3-d77fdfd96-72kbj.txt](https://github.com/eclipse/codewind/files/4072612/codewind-workspacel3sjxgfbqx5n3tp3-d77fdfd96-72kbj.txt)

@jagraj

**There are two workarounds for this issue**
1. User can upgrade to use NFS v4.1 or above
2. User can shell into the PFE container to manually delete corresponding project files/directories (`location: /codewind-workspace`) and project log files/directories (`location: /codewind-workspace/.logs`)

sujeilyfonseca:
/priority stopship
Marking this as "stopship" since it is a regression.

tetchel:
can you post the theia container logs?

tetchel:
```
[16/01/20 16:53:57 User.js] [INFO] Removing project directory at /codewind-workspace/cw-appsody-loopback4
[16/01/20 16:53:57 Turbine] [INFO] No dangling images were found.
[16/01/20 16:54:01 cw-appsody-node-red] [INFO] operationId: 3fc649e9f484f964a9364d52b530c6ea
[16/01/20 16:54:01 cw-appsody-node-red] [INFO] containerDelete: Kill running processes and remove container... 
[16/01/20 16:54:01 cw-appsody-node-red] [INFO] Project ID:        39676c30-3880-11ea-a260-657370f2a7ff
[...]

PR #1843, opened 2020-01-22
Fix project deletion issue

GeekArthur:
Issue: https://github.com/eclipse/codewind/issues/1779

This PR is for fixing project deletion issue. Our own implementation (combine `fs.rmdir` and `fs.unlink`) to delete directory recursively can't handle `EBUSY` and `ENOTEMPTY` properly, especially if user uses NFS v4.1 or below (NFS generates `.nfs<XXXXX>` for "delete after close application"), so that we can hit the issue of project log directory fails to delete when we delete project.

In order to fix this, we can use `rimraf` npm package (https://www.npmjs.com/package/rimraf), I do some research and find it’s the best implement to handle `EBUSY` and `ENOTEMPTY` cases and all other corner cases for recursive delete, it has 24,430,729 weekly download and ISC license.

### DIFF
--- a/src/pfe/file-watcher/server/src/controllers/projectsController.ts
+++ b/src/pfe/file-watcher/server/src/controllers/projectsController.ts
@@ -14,6 +14,7 @@ import crypto from "crypto";
 import fs from "fs";
 import path from "path";
 import AsyncLock from "async-lock";
+import rimraf from "rimraf";
 // local imports
 import { actionMap } from "../projects/actions";
 import * as projectSpecifications from "../projects/projectSpecifications";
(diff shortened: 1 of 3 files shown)


GeekArthur:
@elsony @rajivnathan Can you please review when you have time? Thanks.

src/pfe/file-watcher/server/src/controllers/projectsController.ts
```diff
@@ -936,7 +937,7 @@ export function deleteFolder(dir: string): Promise<void> {
                 Promise.all(files.map((file) => {
                     return deleteFile(dir, file);
                 })).then(() => {
-                    fs.rmdir(dir, (err) => {
+                    rimraf(dir, (err) => {
```
rajivnathan: If we're using rimraf do we still need the individual file deletions beforehand?

GeekArthur: The `.nfs<XXXXX>` is generated in the log dir by NFS after we use the individual delete function to delete the log files in the log dir, so it's still necessary to keep the individual delete functio beforehand.

rajivnathan: Ok, I suppose it can't hurt to keep it

Issue #1723, opened 2020-01-13
Error thrown by mkdir /codewind-workspace/cw-temp/<project>


makandre:
**Codewind version:** 0.8.0
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**

I see this error in Codewind log if I create a project using the same name as I project that had already been removed from Codewind.

```
[13/01/20 16:12:06 /portal/routes/projects/remoteBind.route.js] [ERROR] Project creation failed for project apjava01. Error: { [Error: EEXIST: file already exists, mkdir '/codewind-workspace/cw-temp/apjava01']
  errno: -17,
  code: 'EEXIST',
  syscall: 'mkdir',
  path: '/codewind-workspace/cw-temp/apjava01' }
```

I believe it is thrown here: https://github.com/eclipse/codewind/blob/master/src/pfe/portal/routes/projects/remoteBind.route.js#L134

Probably related to https://github.com/eclipse/codewind/pull/1629 since we do not remove the folder in `cw-temp` anymore

**Steps to reproduce:**

**Workaround:**

makandre:
@tobespc FYI

stalleyj:
The behaviour of syncing has changed slightly so we are no longer cleaning up the temp directory. We should add the directory delete to our unbind functionality.

jgwest:
@markcor11: Unable to locate pipeline with name: `Validate`

makandre:
This fixes Codewind projects, but for Appsody it still a problem because `project.workspace` is `/mounted-workspace` for Appsody projects.   The temp dir path needs to be constructed using the `global.codewind.CODEWIND_WORKSPACE` variable.

/reopen

jgwest:
Moving back to In Progress per above.

stalleyj:
PR here https://github.com/eclipse/codewind/pull/2036

malincoln:
@makandre pls verify and close. Thanks

makandre:
Fix #2036 is not in 0.9.0

https://github.com/eclipse/codewind/blob/0.9.0/src/pfe/portal/modules/User.js#L568

stalleyj:
@makandre Issue is not stop-ship so the fix went into master.

PR #1864, opened 2020-01-24
Delete project files from cw-temp on Project unbind

markcor11:
## Problem

Error seen in logs : Error: EEXIST: file already exists thrown by mkdir /codewind-workspace/cw-temp/<project>   reported in issue #1723

## Solution 

PFE was not removing the project files from /codewind-workspace/cw-temp/<project> when unbinding. 

This PR removes the files to avoid a future collision. 

### DIFF
--- a/src/pfe/portal/modules/User.js
+++ b/src/pfe/portal/modules/User.js
@@ -469,6 +469,15 @@ module.exports = class User {
       projectID: projectID,
       status: 'success'
     };
+
+    // Delete files from PFE's /codewind-workspace/cw-temp/<project> location
+    const cwTempPath = path.join(project.workspace, "cw-temp", project.directory);
+    try {
+      await cwUtils.forceRemove(cwTempPath);
+    } catch (err) {
+      log.info(`Unable to delete temp files from ${cwTempPath}`)
+    }
+
     // Stop streaming the logs files.
     project.stopStreamingAllLogs();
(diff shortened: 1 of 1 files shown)


src/pfe/portal/modules/User.js
```diff
@@ -469,6 +469,15 @@ module.exports = class User {
...
     };
+
+    // Delete files from PFE's /codewind-workspace/cw-temp/<project> location
+    const cwTempPath = path.join(project.workspace, "cw-temp", project.directory);
+    try {
+      await cwUtils.forceRemove(cwTempPath);
+    } catch (err) {
+      log.info(`Unable to delete temp files from ${cwTempPath}`)
+    }
+
```
hhellyer: I think we ought to add this to deleteProjectFiles.  That does the rest of the file cleanup and removes the project from the project list so it can no longer be queried or have files uploaded to it. Deleting the temporary space before we do that sounds like we could end up with a hard to figure out bug later on.

markcor11: I thought about that, but it seems there was a chance exceptions could get thrown before deleteProjectFiles gets called. That would leave the temp files lingering wouldn't it ?

markcor11: Moved code to deleteProjectFiles to ensure filewatcher is done with the files before deleting temp

PR #1868, opened 2020-01-24
Install cwctl version based on passed argument instead of always master

ssh24:
### Description

We should install cwctl for our tests based on the passed argument instead of always master. This was seen when doing performance test on 0.8.0 and the cwctl project create command had a different syntax to compared to what master has and the performance test broke.


ssh24:
@maysunfaisal Could you PTAL?

ssh24:
This PR needs to be back ported to both 0.8.0 and 0.7.0 in order to get performance metrics automatically.

PR #1878, opened 2020-01-24
Install cwctl version based on passed argument instead of always master

ssh24:
Cherry pick of https://github.com/eclipse/codewind/pull/1868 to 0.8.0


tobespc:
why push this to 0.8.0 ?   We've shipped that

Issue #1862, opened 2020-01-23
Allow handling of app statuses for serverless templates 


ssh24:
**Description of the enhancement:**
Related to https://github.com/eclipse/codewind/issues/1269

**Proposed solution:** 
Add function to Turbine so that if the port on cw-settings is set to a special value, e.g. -1, it indicates that the app does not have a port for ping so it will not cause error on PFE on failing to ping.

/area iterative-dev
/assign
/pipeline in progress

PR #1880, opened 2020-01-24
Allow handling of app status for server less applications

ssh24:
Closes https://github.com/eclipse/codewind/issues/1862

### Description
Add function to Turbine so that if the port on cw-settings is set to a special value, e.g. -1, it indicates that the app does not have a port for ping so it will not cause error on PFE on failing to ping.

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/constants.ts
+++ b/src/pfe/file-watcher/server/src/projects/constants.ts
@@ -36,3 +36,5 @@ export enum StartModes { run = "run", debug = "debug", debugNoInit = "debugNoIni
 export enum ControlCommands { stop = "stop", start = "start", restart = "restart" }
 
 export enum MavenFlags { profile = "-P ", properties = "-D " }
+
+export const serverLessAppPort = "-1";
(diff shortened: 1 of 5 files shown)


src/pfe/file-watcher/server/src/projects/constants.ts
```diff
@@ -36,3 +36,5 @@ export enum StartModes { run = "run", debug = "debug", debugNoInit = "debugNoIni
 export enum ControlCommands { stop = "stop", start = "start", restart = "restart" }
 
 export enum MavenFlags { profile = "-P ", properties = "-D " }
+
+export const serverLessAppPort = "-1";
```
rajivnathan: Suggest renaming this variable to something like `DisablePingPort` since "serverless" can be confused with serverless computing.

src/pfe/file-watcher/server/src/utils/locales/en/translation.json
```diff
@@ -75,6 +75,7 @@
   },
   "projectUtil": {
     "projectInfoError": "could not retrieve project info",
+    "serverLessApp": "server less application detected. app status set to stopped",
```
rajivnathan: Can we update the key to something similar to the variable? eg. "pingDisabled": "The port is set to -1, pinging is disabled and that app status is set to Stopped."

PR #1606, opened 2019-12-20
Add Code Coverage for PFE into Jenkins

tobespc:
Signed-off-by: Toby Corbin <corbint@uk.ibm.com>

PR #1882, opened 2020-01-27
Add Code Coverage for PFE into Jenkins

james-wallis:
Proof that the code coverage is the same as the Toby/Liam branch: https://github.com/tobespc/codewind/compare/liam...james-wallis:add-code-coverage-jenkins?expand=1

Use the test passing and failing from the #1606 branch as these will not run with the updated Jenkinsfile.

### Summary
* Added functionality to generate the code coverage for PFE unit and API tests in Jenkins, also optionally the file-watcher unit test coverage.
  * Had to enhance the `generate_complete_coverage.js` file to report the file-watcher code optionally.
  * Uses images with the tag `:test` as the PFE needs a change to run code coverage, doing it this way means that the original images `:latest` will be pushed without changes made for testing.
* Made the running of code coverage optional
  * Stops it being run when developing if wanted
  * Fixed bug in Jenkins where the `apitest` run would overwrite the previous `unittest` run. In the future this fix might need to be more complex (i.e. different `.nyc_output` directories) but for now this is fine. - We don't care about the `apitest` local `.nyc_output` as the code coverage is inside the container.

### Testing
* Ran the Jenkins test
* Ran the Portal tests locally to ensure nothing changed there and the generate coverage script too.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -74,7 +74,7 @@ pipeline {
 
         stage('Run Turbine unit test suite') {
             options {
-                timeout(time: 30, unit: 'MINUTES') 
+                timeout(time: 30, unit: 'MINUTES')
             }
             steps {
                 withEnv(["PATH=$PATH:~/.local/bin;NOBUILD=true"]) {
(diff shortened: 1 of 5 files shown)


james-wallis:
@rwalle61 has reviewed this PR on the tobes/liam branch: https://github.com/eclipse/codewind/pull/1606#pullrequestreview-348128984

Jenkinsfile
```diff
@@ -195,6 +195,22 @@ pipeline {
...
+
+                                echo -e "\n+++   RETAGGING $IMAGE_NAME TO HAVE TAG:':test' FOR TEST ONLY MODIFICATIONS +++\n";
+                                # Allows us to modify the Docker images for test only modifications
+                                # Such as telling PFE to run with Code Coverage generation enabled 
+                                # The original image (latest) will be the one that is pushed to Dockerhub
+
+                                if [ "$image" = "$PFE" ]; then
+                                    # Now overwrite the PFE image ':test' tag with new changes
+                                    echo -e "FROM eclipse/codewind-pfe-amd64:latest\nENV ENABLE_CODE_COVERAGE=true" >> pfe-test-dockerfile
+                                    cat pfe-test-dockerfile
```
hhellyer: I think you left a debug statement here.

hhellyer:
Merged. (No point waiting for the build as Jenkins won't use the Jenkinsfile from this branch. Changes were built under PR #1606 )

Issue #1875, opened 2020-01-24
TypeError: Converting circular structure to JSON


tetchel:
Codewind version: master `f60f0cff5c83e9084d55426987879f97851f6272`

I see this throughout my PFE log:
```
[24/01/20 17:54:43 /portal/modules/Project.js] [ERROR] TypeError: Converting circular structure to JSON
    at JSON.stringify (<anonymous>)
    at stringify (/portal/node_modules/jsonfile/index.js:86:18)
    at Object.writeFile (/portal/node_modules/jsonfile/index.js:101:11)
    at Promise (/portal/node_modules/universalify/index.js:13:12)
    at new Promise (<anonymous>)
    at Object.writeFile (/portal/node_modules/universalify/index.js:7:14)
    at Project.writeInformationFile (/portal/modules/Project.js:266:16)
```

Looks like something in the structure being written to the inf has a circular reference and can't be stringified. This probably means the inf update fails, so the data stored may be out-of-date.

mattcolegate:
Recreation: take any running project in the IDE and click 'Show all logs'. Subsequent changes that cause the project.inf file to be written to (e.g. a small code change) will have the OP error in the PFE logs.

mattcolegate:
I've raised https://github.com/eclipse/codewind/pull/1925 which fixes the issue by ensuring the offending object never gets written out to project.inf - I've stated my reasoning as to why that was the original intention in the PR. I'd like IDE people to comment on how they're using that part of the project object to make doubly sure that not having it in the project.inf won't break anything.

tetchel:
We are not using `logStreams` directly but this may have an effect on the /logs api which we do use

mattcolegate:
FYI `logStreams` will still exist in the in-memory Project object, it's just not being written out to project.inf and so it's values won't be preserved over PFE restarts.

tetchel:
makes sense to me, afaik all the log streams are turned off when pfe restarts anyway

stalleyj:
Will be fixed when https://github.com/eclipse/codewind/pull/1941 lands.

mattcolegate:
@tetchel PR has landed so this issue is ready for verification please

PR #1925, opened 2020-01-29
Remove logStreams from .inf storage

mattcolegate:

This fixes https://github.com/eclipse/codewind/issues/1875 by not attempting to write `logStreams` out to the project.inf file. The throw error means it never made it there anyway, and https://github.com/eclipse/codewind/blob/master/src/pfe/portal/modules/Project.js#L153 seems to be making it clear that it wasn't supposed to by design; nevertheless I would like reviews from IDE people and also Howard who wrote the streaming section(?)


hhellyer:
Closing this as we need to move the other fields that are stripped out in toJSON() too.

PR #1932, opened 2020-01-29
Fix exit code of run.sh when build script fails

ssh24:
### Description

We were always exiting out of run.sh with exit code = 0 when build.sh fails because of the exit code of the previous command of printf `$?`.

Changed this to save the actual exit code of build.sh and then exit with that exit code if the build failed.


ssh24:
@maysunfaisal Could you PTAL?

PR #1933, opened 2020-01-29
Fix K8 condition on workspace test after hook

ssh24:
### Description

The after hook on workspace test needs to be encapsulated on K8 condition because setting and testing image push registry only works in Kube.


ssh24:
@maysunfaisal Could you PTAL?

Issue #1812, opened 2020-01-20
Provide users with a link to docs for help with project status problems


rajivnathan:
**Description of the enhancement:**

As part of https://github.com/eclipse/codewind/issues/1732 we discussed creating a troubleshooting doc for project status issues (eg. Why is my app stuck in starting state?). It would also be good to include a link to this doc from the IDEs in some cases. (eg. when the project status timeout has exceeded)

May need to consolidate with existing docs. For example, I found this:
https://www.eclipse.org/codewind/mdteclipsetroubleshooting.html#application-stuck-in-starting-state

Issue for the new doc: https://github.com/eclipse/codewind/issues/2117

**Proposed code changes**

This is currently what the timeout dialog looks like in the VSCode extension:
![image](https://user-images.githubusercontent.com/20015929/72941541-39ca1380-3d3f-11ea-95e1-24ff67a31f42.png)

We could have a general property that indicates to the IDEs that they should display a link. The new property could be included in the timeout event or other events.
eg.
```
"linkLabel": "Troubleshooting status"
"link": "https://www.eclipse.org/codewind/troubleshooting.html#checking-the-application-and-build-statuses",
```

The current event is:
```
message: projectStatusChanged
 data: {
  "projectID": "80d46e00-3d68-11ea-a131-75340c3788ad",
  "appStatus": "starting",
  "detailedAppStatus": {
    "severity": "WARN",
    "message": "Timeout: Pinging http://172.22.0.4:8080/health and http://172.22.0.4:8080/. The application is taking longer than expected to start. Please check the application logs for problems and ensure the project's 'healthCheck' setting is correct. Last error message received: Pinging http://172.22.0.4:8080/health and http://172.22.0.4:8080/",
    "notify": true
  }
}
```

An updated event with the `linkLabel` and `link` properties included would look like:
```
message: projectStatusChanged
 data: {
  "projectID": "80d46e00-3d68-11ea-a131-75340c3788ad",
  "appStatus": "starting",
  "detailedAppStatus": {
    "severity": "WARN",
    "message": "Timeout: Pinging http://172.22.0.4:8080/health and http://172.22.0.4:8080/. The application is taking longer than expected to start. Please check the application logs for problems and ensure the project's 'healthCheck' setting is correct. Last error message received: Pinging http://172.22.0.4:8080/health and http://172.22.0.4:8080/",
    "notify": true,
    "linkLabel": "Troubleshoot status",
[...]

PR #1935, opened 2020-01-29
Update project status event with troubleshoot info

lentheivendra:
For https://github.com/eclipse/codewind/issues/1812

Updated project status event to include troubleshooting info as discussed, and tested changes by creating Appsody node.js project which recreates scenario of app stuck in starting state.

Signed-off-by: Len Theivendra <theivend@ca.ibm.com>

### DIFF
--- a/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts
+++ b/src/pfe/file-watcher/server/src/controllers/projectStatusController.ts
@@ -51,6 +51,8 @@ const pingInterval = 10000;
 const inTransitPingInterval = 2000;
 export const pingCountMap = new Map();
 
+const troubleShootLink = "https://www.eclipse.org/codewind/troubleshooting.html#checking-the-application-and-build-statuses";
+
 // Keep track of project status.  The type parameter is used to determine the type of the status
 // such as application state or build status.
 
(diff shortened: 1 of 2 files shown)


rajivnathan:
@jopit @elsony Do you think we should add another optional property to `detailedAppStatus` like `statusId` or something so that IDEs can differentiate between status events in case we want to do something like fine-grained filtering of these events if needed?

jopit:
@rajivnathan would the status id identify the type of detailedStatus, like "ping timed out" or would it identify a set of related detailedStatus messages -- i.e. an unbroken sequence of "ping timed out" errors would get one id, then if ping worked for a while and then broke again, there'd be a new status id for the next unbroken sequence of "ping timed out" messages?

rajivnathan:
@jopit I was thinking the former, so just a general type of detailedStatus eg. "ping timed out". But after talking it through with Elson I think we can put that in if/when we decide we need something like that.

src/pfe/file-watcher/server/src/controllers/projectStatusController.ts
```diff
@@ -530,10 +532,13 @@ function pingInTransitApplications(): void {
                                         data.detailedAppStatus = detailedAppStatus;
                                         pingCountMap.delete(projectID);
                                     } else if (newState === AppState.starting && pingCount == pingCountLimit) {
+                                        const troubleShootLinkLabel = await locale.getTranslation("projectStatusController.linkLabel");
                                         detailedAppStatus = {
                                             severity: "WARN",
                                             message: newMsg,
-                                            notify: true
+                                            notify: true,
+                                            linkLabel: troubleShootLinkLabel,
```
rajivnathan: Do we need updates to the unit/functional tests to validate that these events have the `linkLabel` and `link` properties when `notify` is `true`?

rajivnathan: We can handle this in the subsequent PR

PR #1950, opened 2020-01-30
ensure ignoredPaths and refPaths always defined

makandre:

Address a bug where if you added `refPaths` but you did not have `ignoredPaths` defined, then we miss sending out event to Filewatcherd


jgwest:
For https://github.com/eclipse/codewind/issues/1399

Issue #1973, opened 2020-01-31
Project Create Failing on Helm Install


maysunfaisal:
**Codewind version:** latest Jan 31
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:** OKD 3.11

**Description:**
```
[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:43c67172d1d182ca5460fc962f8f053f33028e0a3a1d423e05d91b532429e73d

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:f55aa0bd26b801374773c103bed4479865d0e37435b848cb39d164ccb2c3ba51

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:e993ebdaa176d2f4c726067d33b836eeec5de7d492fb94cbe23bd44cd21a121e

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:1d0dfb259f6a31f95efcba61f0a3afa318448890610c7d9a64dc4e95f9add843

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:dceebd4a88d8325a8509045dc5623bce5882d4e05c60c0102374ff03f44f060d

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:21ec61b65b20ec53a1b7f069fd04df5acb0e75434bd3603c88467c8bfc80d9c6

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:67cc3254f33d92e8593b343b430a73a329b3498ac231dccd8c1aeedc73f2848b

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:3bc9bc90fae672068ee00733474d7a2539e441a621017953aa9ce7491cb79bc4

[31/01/20 19:46:47 frispr1] [ERROR] executeBuildScript : 295 | Copying blob sha256:81219c22a87f596e5fc65b9468203b2e7c50d9947e23ffb4d64f9ecf954bae40

[...]

PR #1976, opened 2020-01-31
Revert to yq 2.4.1

johnmcollier:
Fixes https://github.com/eclipse/codewind/issues/1973

`yq` 3.0 was recently released and it changed the syntax of a number of commands, causing the commands in https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/scripts/kubeScripts/modify-helm-chart.sh to fail. This was causing the Helm chart to fail to deploy with the following error:
```
 Error: unable to build kubernetes objects from release manifest: error validating "": error validating
data: [ValidationError(Service.metadata.ownerReferences[0]): missing required field "kind" in
io.k8s.apimachinery.pkg.apis.meta.v1.OwnerReference,
ValidationError(Service.metadata.ownerReferences[0]): missing required field "name" in
io.k8s.apimachinery.pkg.apis.meta.v1.OwnerReference,
ValidationError(Service.metadata.ownerReferences[0]): missing required field "uid" in
io.k8s.apimachinery.pkg.apis.meta.v1.OwnerReference,
ValidationError(Service.metadata.ownerReferences[1]): missing required field "apiVersion" in
io.k8s.apimachinery.pkg.apis.meta.v1.OwnerReference]
```

Since our yq scripts are heavily dependent on the yq 2.0 syntax, this PR is reverting the yq we're using back to 2.4.1, so that the error doesn't occur.

### DIFF
--- a/src/pfe/Dockerfile_x86_64
+++ b/src/pfe/Dockerfile_x86_64
@@ -51,8 +51,8 @@ RUN curl -f -o buildah.rpm $BUILDAH_RPM \
     && cp -r /tmp/$OPENJ9_JRE /opt/java/jre \
     && rm -rf /tmp/$OPENJ9_JRE /tmp/OpenJDK8U-jre_x64_linux_openj9_8u222b10_openj9-0.15.1.tar.gz \
 # Install the latest version of yq
-    && VERSION=2.4.1 \
-    && curl -sL -O https://github.com/mikefarah/yq/releases/download/$VERSION/yq_linux_amd64 \
+    && YQ_VERSION=2.4.1 \
+    && curl -sL -O https://github.com/mikefarah/yq/releases/download/$YQ_VERSION/yq_linux_amd64 \
     && echo '754c6e6a7ef92b00ef73b8b0bb1d76d651e04d26aa6c6625e272201afa889f8b  yq_linux_amd64' | sha256sum -c - \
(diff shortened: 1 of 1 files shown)


src/pfe/Dockerfile_x86_64
```diff
@@ -51,9 +51,9 @@ RUN curl -f -o buildah.rpm $BUILDAH_RPM \
     && cp -r /tmp/$OPENJ9_JRE /opt/java/jre \
     && rm -rf /tmp/$OPENJ9_JRE /tmp/OpenJDK8U-jre_x64_linux_openj9_8u222b10_openj9-0.15.1.tar.gz \
 # Install the latest version of yq
-    && LATEST_RELEASE=$(curl -L -s -H 'Accept: application/json' https://github.com/mikefarah/yq/releases/latest) \
-    && LATEST_VERSION=$(echo $LATEST_RELEASE | sed -e 's/.*"tag_name":"\([^"]*\)".*/\1/') \
-    && curl -sL -O https://github.com/mikefarah/yq/releases/download/$LATEST_VERSION/yq_linux_amd64 \
+    && VERSION=2.4.1 \
```
maysunfaisal: maybe rename this to `YQ_VERSION`, other than that LGTM

johnmcollier: 👍

Issue #1972, opened 2020-01-31
Turbine Local Functional Failures on create & modify


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
### During modify, PFE emits socket events but tests dont catch it:

```
[32m[31/01/20 05:34:05 turbinetest-default-swift] [INFO][39m Emitting event 
 message: projectChanged
 data: {
  "operationId": "b9664a32b6f3bfb45a1d856e15a56dfc",
  "projectID": "default-swift-1580440079540",
  "ignoredPaths": [
    "*/node_modules*",
    "*/.git/*",
    "*/.DS_Store",
    "*/.dockerignore",
    "*/.gitignore"
  ],
  "contextRoot": "/",
  "isHttps": false,
  "statusPingTimeout": "30",
  "status": "success",
  "host": "172.18.0.4",
  "ports": {
    "exposedPort": "32787",
    "internalPort": "8080"
  },
  "containerId": "33dff1cd9bbf6b5c6f2286ac4f02d9da86d25577771d31f61113cd5261e49106",
  "logs": {
    "build": [
      {
        "origin": "workspace",
        "files": [
          "/codewind-workspace/.logs/turbinetest-default-swift-default-swift-1580440079540/docker.app.log",
          "/codewind-workspace/.logs/turbinetest-default-swift-default-swift-1580440079540/app.compile.log",
          "/codewind-workspace/.logs/turbinetest-default-swift-default-swift-1580440079540/docker.build.log"
        ]
      }
    ],
    "app": [
      {
        "origin": "workspace",
        "files": [
          "/codewind-workspace/.logs/turbinetest-default-swift-default-swift-1580440079540/app.log"
        ]
      }
    ]
  }
}
[...]

PR #1977, opened 2020-01-31
Fix functional test on update build queue

ssh24:
### Description

This PR fixes the functional test on update cases with the new build queue on update scenario. It also fixes the kubectl copy on bulk copy of the project directory.

Screen Shot 2020-01-31 at 4 40 00 PM

Closes https://github.com/eclipse/codewind/issues/1972


ssh24:
@maysunfaisal Could you PTAL?

PR #1983, opened 2020-02-03
Made changes to support ppc64le architecture

vrushaliinamdar:
Made following changes to support ppc64le architecture:

1. Suffixed architecture name to gatekeeper dockerfile
2. Changed gatekeeper build file to accept filename as an argument
3. Created ppc64le specific Dockerfiles for pfe, performance and gatekeeper components
4. Changed build.sh to disable codewind-appsody-extension for ppc64le

Signed-off-by: Vrushali Inamdar <vrushali.inamdar@ibm.com>

### DIFF
--- a/src/gatekeeper/Dockerfile_x86_64
+++ b/src/gatekeeper/Dockerfile_x86_64
@@ -1,0 +1,40 @@
+################################################################################
+# Copyright (c) 2019 IBM Corporation and others.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v2.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v20.html
+#
+# Contributors:
+#     IBM Corporation - initial API and implementation
+################################################################################
+FROM centos:8.1.1911
+RUN yum -y install nodejs \
+    && yum clean all \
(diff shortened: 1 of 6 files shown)


src/gatekeeper/Dockerfile_x86_64
```diff
@@ -0,0 +1,40 @@
+################################################################################
```
hhellyer: This is the right thing to do - the only thing missing here is a delete or rename of the original Dockerfile to Dockerfile_x86_64. (If I check out this branch both  `src/gatekeeper/Dockerfile_x86_64` and `src/gatekeeper/Dockerfile` exist.  src/gatekeeper/Dockerfile should be deleted now we've split this by platform.)

hhellyer:
This looks good, the changes don't affect the x86_64 builds so this can be merged safely. (I can't verify they work on ppc64le as I don't have access to that build env but if @vrushaliinamdar is happy with that there's no reason not to merge.)

src/gatekeeper/Dockerfile_x86_64
hhellyer: (I'm fixing this in a follow up PR to enable us to merge the ppc64le changes.)

Issue #1222, opened 2019-11-20
Performance difference in large Java application between 0.6.0 and 0.5.0 


jgwest:
There appears to be a significant performance difference between 0.6.0 and 0.5.0, when tested against a largish Java application (the Rogue Cloud client). 

#### Windows Laptop, local, on Eclipse (Eclipse and Codewind version matching):
```
0.6: consistently about ~40 seconds to build complete, after making a source change (likewise it seems to takes an additional ~5 seconds before the build actually starts, which is not the case on 0.5)
0.5: consistently about ~15  seconds to build complete, after making a source change
```

#### Linux Desktop,  local (desktop machine, but VM on non-SSD) on Eclipse (Eclipse and Codewind version matching):
```
0.6: 40 seconds
0.5: 16 seconds
```

#### Windows Laptop, local, on VSCode (versions matching)
```
VScode on Windows on 0.6: 46 seconds
VSCode on Windows on 0.5: 11* seconds
(* auto build didn't work for unknown reasons, so this is me hitting the 'Build' button, which should be appx the same.)
```

### Steps to run

Run the following (or use the UI in the IDE of your choice):
```
git clone https://github.com/microclimate-dev2ops/rogue-cloud-client-codewind

# WARNING: You MUST rename the project before importing into Codewind, to avoid https://github.com/eclipse/codewind/issues/1217
mv rogue-cloud-client-codewind  gameclient
```

Replace the `gameclient/.cw-settings` file with this:
```
{
  "contextRoot": "",
  "internalPort": "",
  "healthCheck": "",
  "internalDebugPort": "",
  "isHttps": false,
  "ignoredPaths": [
    ""
  ],
  "mavenProfiles": [
    ""
  ],
  "mavenProperties": [
    ""
  ]
}
```
WARNING: If you don't do this, you will get a nasty error message from Eclipse indicating that `watchedFiles` could not be found (this due to https://github.com/eclipse/codewind/issues/133)

Import the project into your workspace, and add it to Codewind. Wait for the project to build and start.

Open StartAgentServlet.java (CTRL-Shift-R to find it in Eclipse, other key combs in other IDEs). Make a change to it. Hit Save.

Immediately start a stopwatch.
Stop the stopwatch when IDE has 'build successful' status on the project (do not use stopping/stopped/starting/started as the indicator to stop the stopwatch.)

### Methodology

This is a holistic approach meant to capture the user's experience of build time:
[...]

PR #1997, opened 2020-02-03
Set chunk total to be 1 on update project function for test call

ssh24:
### Description

Related to https://github.com/eclipse/codewind/issues/1222

We need to set chunk total equal to 1 for the update to not get stuck in a 20s timer. That is why we would see a 20s extra on update cases on our performance tests.

The issue is that in our product code, we expect the function to be passed with `chunk_total` value of 1. See https://github.com/eclipse/codewind/blob/b0f16f0a76be2f543d501fc718ddeaa727d75d1d/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts#L167-L169

So if chunk_total value is not passed 1, in our previous case, we just used a random number, this if statement https://github.com/eclipse/codewind/blob/b0f16f0a76be2f543d501fc718ddeaa727d75d1d/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts#L217 was never entered because the value of `newChunkRemaining` was not 0 from missing the previous if statement.

As a result, we would get stuck in this `setTimeout` function https://github.com/eclipse/codewind/blob/b0f16f0a76be2f543d501fc718ddeaa727d75d1d/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts#L262 for 20s as defined here https://github.com/eclipse/codewind/blob/b0f16f0a76be2f543d501fc718ddeaa727d75d1d/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts#L42

Not sure why we have that 20s timer in the first place - looks like most of the code inside that timer is duplicated on the if statement provided above, or why the `updateProjectForNewChange` function only takes `chunk_total` value of 1 is not documented properly. But for that I have opened another issue https://github.com/eclipse/codewind/issues/1998

### DIFF
--- a/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
+++ b/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
@@ -30,11 +30,10 @@ export function projectEventTest(socket: SocketIO, projData: projectsController.
             socket.clearEvents();
         });
 
-        const chunk = Math.random() * 1000;
         const data: any = {
             "projectID": projData.projectID,
             "timestamp": Date.now(),
-            "chunk": chunk,
+            "chunk": 1,
             "chunk_total": 1,
             "eventArray": []
         };
--- a/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
+++ b/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
@@ -35,7 +35,7 @@ export function projectEventTest(socket: SocketIO, projData: projectsController.
             "projectID": projData.projectID,
             "timestamp": Date.now(),
             "chunk": chunk,
-            "chunk_total": chunk + Math.random() * 100,
+            "chunk_total": 1,
             "eventArray": []
         };
 


ssh24:
@maysunfaisal Could you PTAL?

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
```diff
@@ -35,7 +35,7 @@ export function projectEventTest(socket: SocketIO, projData: projectsController.
             "projectID": projData.projectID,
             "timestamp": Date.now(),
             "chunk": chunk,
```
yangcao77: The `chunkNum` indicates number of chunks in `chunk_total`. It should be <= `chunk_total`

Since you only send 1 chunk in the update test, setting `chunk_total=1` is correct. But don't forget to also set `chunk=1`. Your original value is a random number.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
```diff
@@ -35,7 +35,7 @@ export function projectEventTest(socket: SocketIO, projData: projectsController.
...
-            "chunk_total": chunk + Math.random() * 100,
+            "chunk_total": 1,
             "eventArray": []
```
yangcao77: If your `eventArray` is an empty array. It would not triggers the update. Since no file is considered `changed`, no build would be triggerred.

yangcao77: The old value would triggers the build because Turbine thinks there are more chunks/changedFiles, but the chunk is lost due to network issue. 
So it would trigger build after the timer's timeout reaches.

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
ssh24: So how does is 20s timer not hit when I have `chunk` set to a random number and `chunk_total` is set to 1?

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
ssh24: `eventArray` is not empty. We update it dynamically https://github.com/eclipse/codewind/blob/ad1e09bdb0bc2210d5e5ba952f1a35a16a382675/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts#L103 and updates do happen

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
yangcao77: Since when `chunk_total=1`, we would proceed the build directly. 
https://github.com/eclipse/codewind/blob/b0f16f0a76be2f543d501fc718ddeaa727d75d1d/src/pfe/file-watcher/server/src/controllers/projectEventsController.ts#L167-L168

But since this is the functional test, the `chunkNum` should be set to an appropriate value. 
Random number does not make any sense in the functional level.

ssh24: Alright, set `chunk` to 1 as well.

yangcao77:
Changes now LGTM.

PR #2006, opened 2020-02-04
Move DockerJava=OpenLiberty into utility function

### LINKED ISSUE #2002: Updating pom.xml removes javametrics from an open liberty project.
**Codewind version:**
latest

**Description:**
[04/02/20 11:41:18 /portal/routes/projects/remoteBind.route.js] [WARN] Error: Injection of metrics collector is not supported for projects of type 'docker'
    at Object.injectMetricsCollectorIntoProject (/portal/modules/metricsService/index.js:41:11)
    at uploadEnd (/portal/routes/projects/remoteBind.route.js:274:32)

**Steps to reproduce:**
- Enable metrics injection for an open liberty project.
- Confirm javametrics has been injected into the remote pom.xml file.
- Edit the local pom.xml file and save to sync a new version.
- Examine the remote pom.xml file again, javametrics will have been removed. 

**Workaround:** 
Disabling and enabling metrics injection will restore it after a pom.xml change.

mattcolegate:

Fixes #2002 by moving the identification of Docker-Java projects as OpenLiberty projects into MetricsService, so that it can be called from both metrics.controller and remote.bind.

### DIFF
--- a/src/pfe/portal/modules/metricsService/index.js
+++ b/src/pfe/portal/modules/metricsService/index.js
@@ -36,20 +36,30 @@ const metricsCollectorRemovalFunctions = {
   spring: removeMetricsCollectorFromSpringProject,
 }
 
-async function injectMetricsCollectorIntoProject(projectType, projectDir) {
-  if (!metricsCollectorInjectionFunctions.hasOwnProperty(projectType)) {
-    throw new Error(`Injection of metrics collector is not supported for projects of type '${projectType}'`);
+async function identifyProject(projectType, projectLanguage) {
+  if (projectType === 'docker' && projectLanguage === 'java') {
+    return 'openLiberty';
+    // this is the best we can do at the moment
   }
-  await metricsCollectorInjectionFunctions[projectType](projectDir);
-  log.debug(`Successfully injected metrics collector into ${projectType} project`);
+  return projectType
 }
 
-async function removeMetricsCollectorFromProject(projectType, projectDir) {
-  if (!metricsCollectorRemovalFunctions.hasOwnProperty(projectType)) {
-    throw new Error(`Injection of metrics collector is not supported for projects of type '${projectType}'`);
+async function injectMetricsCollectorIntoProject(projectType, projectLanguage, projectDir) {
+  let projType = await identifyProject(projectType, projectLanguage);
+  if (!metricsCollectorInjectionFunctions.hasOwnProperty(projType)) {
+    throw new Error(`Injection of metrics collector is not supported for projects of type '${projType}'`);
   }
-  await metricsCollectorRemovalFunctions[projectType](projectDir);
-  log.debug(`Successfully removed metrics collector from ${projectType} project`);
+  await metricsCollectorInjectionFunctions[projType](projectDir);
+  log.debug(`Successfully injected metrics collector into ${projType} project`);
+}
+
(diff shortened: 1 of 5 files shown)


hhellyer:
I don't think this is working, not simply due to the changes in this PR.
This line:
https://github.com/eclipse/codewind/blob/6fa349a077ae8d0e5b9dafcc1a97eb62ee621a7a/src/pfe/portal/modules/metricsService/index.js#L144-L145
means that we won't inject metrics into a project that has had metrics injected once before.

- If a pom file is updated while metrics are injected we will end up overwriting pom.xml with the new one, without metrics injected.
- When we disable injection we restore the pom file to whatever it was *before* metrics were injected, even if that's out of sync with the filesystem.

This line should probably just be deleted so we always re-inject.

mattcolegate:
The presence of a backup pom.xml acts as a flag to prevent re-injection of a project that has already been injected. Updating the pom.xml means that we should back up this new pom.xml and then inject it - essentially treating it like a new project from an injection perspective. I've exposed a method to delete the backup pom.xml to enable this. Tested and the problem looks fixed.
@hhellyer please re-test

mattcolegate:
@hhellyer can you give me an outlook for when the new changes can be reviewed please?

src/pfe/portal/routes/projects/remoteBind.route.js
```diff
@@ -270,8 +270,13 @@ async function uploadEnd(req, res) {
       await cwUtils.copyProject(pathToTempProj, path.join(project.workspace, project.directory), getMode(project));
 
       if (project.injectMetrics) {
+        const projectDir = path.join(project.workspace, project.directory);
+        if (modifiedList.includes("pom.xml")) {
```
hhellyer: This checks for a pom.xml but most of the other functions are project type neutral. This issue probably affects NodeJS as well.
Either we need a function on metricsService that can return the name (or names) of the files that trigger reinjection so we can decide correctly here or we just always re-inject. We would only need to restore from the backup when injection was disabled.

src/pfe/portal/modules/metricsService/index.js
```diff
@@ -87,6 +97,13 @@ async function removeMetricsCollectorFromJvmOptions(projectDir) {
   }
 }
 
+async function deleteBackupPomXml(projectDir) {
+  const pathToBackupPomXml = getPathToBackupPomXml(projectDir);
+  if (await fs.exists(pathToBackupPomXml)) {
+    await fs.remove(pathToBackupPomXml);
```
hhellyer: No real need for the `fs.exists` check here. The `fs.remove()` function doesn't error if the file doesn't exist - https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/remove.md

test/src/unit/controllers/metrics.controller.test.js
```diff
@@ -17,7 +17,7 @@ const { suppressLogOutput } = require('../../../modules/log.service');
 
 chai.should();
 
-describe('metrics.controller.js', () => {
+describe.only('metrics.controller.js', () => {
```
hhellyer: Need to remove `only`. (I think the linter will catch this too.)

hhellyer:
I pushed a new PR https://github.com/eclipse/codewind/pull/2132 with Matts changes, some of mine and that had been rebased on the latest master, I'll close this PR for now as the same changes are under a more recent one.

Issue #1993, opened 2020-02-03
Can't add new NPM dependency through Codewind


james-wallis:
**Codewind version:** Master
**OS:** Mac
**Deployment:** Docker but may happen in K8s - can test tomorrow if needed to.

**Description:**
AFAIK I should be able to modify the dependency object in a Node.js `package.json` and it should pull in the new dependency. Once I make the code change the code isn't being rebuilt. I've tried pressing the `build` button (doesn't do anything) and disabling and enabling the application which appears to make the image rebuild but then it hangs on `Building - Build started` (from the VSCode IDE).

This has happened in three applications:
1. Source code downloaded from a Node.js template over a month ago, looks like a generic express app to me.
2. The Node.js simple express Kabanero template.
3. Appsody Node.js simple express application - couldn't even see the `dependencies` section but I assume this is an Appsody thing (architect vs dev). 
    * On a side note I'm not sure what this brings us within Codewind as I couldn't see a way to pull in the dependencies? Is there additional setup for Appsody with Codewind?

**Steps to reproduce:**
1. Create a Node.js application
2. Wait for it to build and verify its started in the browser
3. Add a new dependency to the `package.json` (I used `"axios": "^0.19.2"` - had to drop out of Codewind and look up the version) and save it.
4. Add some code into your Node application which uses the new package and see if works as you'd expected to.

**Workaround:** 

disable and enable the application.

johnmcollier:
Wonder if this is related to Nodemon? I know we're using it on local for "Codewind-style" projects, and I believe the Appsody node stack also uses Nodemon to check for updates.

tetchel:
I tried it just now but ran `npm install` from the project directory on my disk - instead of editing package.json directly - and it worked

tetchel:
this is probably a cross-squad problem so I will tag both

james-wallis:
Just checked out master again and retried this. Added Axios to the `package.json` `"axios": "^0.19.2"` directly after the two changes described in #2001.
Made my application crash:
```
> node@1.0.0 start /app
> node $npm_package_config_entrypoint

[Tue Feb  4 10:54:37 2020] com.ibm.diagnostics.healthcenter.loader INFO: Node Application Metrics 5.1.1.202002041039 (Agent Core 4.0.5)
[Tue Feb  4 10:54:38 2020] com.ibm.diagnostics.healthcenter.mqtt INFO: Connecting to broker localhost:1883
[...]

PR #2012, opened 2020-02-04
Must check for noderun exit code and update build status on nodemon restart

ssh24:
### Description

Previously we would not check if the `noderun` start command was successful or not and the build status would be stuck at `building - build started`. This PR fixes this by checking the exit code status of the nodemon start command, and then sets the build status to success.

Fixes https://github.com/eclipse/codewind/issues/1993

### DIFF
--- a/src/pfe/file-watcher/scripts/nodejs-container.sh
+++ b/src/pfe/file-watcher/scripts/nodejs-container.sh
@@ -423,8 +423,8 @@ elif [ "$COMMAND" == "update" ]; then
 				echo "Build succeeded, setting build status to success"
 				$util updateBuildState $PROJECT_ID $BUILD_STATE_SUCCESS " "
 			else
-				echo "Noderun failed to start, setting build status to failed"
-				$util updateBuildState $PROJECT_ID $BUILD_STATE_FAILED "Noderun failed to start"
+				echo "Nodemon has failed to start, setting build status to failed"
+				$util updateBuildState $PROJECT_ID $BUILD_STATE_FAILED "buildscripts.nodemonFailedStart"
 			fi
 			$util updateAppState $PROJECT_ID $APP_STATE_STARTING
 		fi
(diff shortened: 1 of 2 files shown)


src/pfe/file-watcher/scripts/nodejs-container.sh
```diff
@@ -419,6 +419,13 @@ elif [ "$COMMAND" == "update" ]; then
 			$util updateAppState $PROJECT_ID $APP_STATE_STOPPING
 
 			$IMAGE_COMMAND exec $project /scripts/noderun.sh start $AUTO_BUILD_ENABLED $START_MODE $HOST_OS
+			if [[ $? -eq 0 ]]; then
+				echo "Build succeeded, setting build status to success"
+				$util updateBuildState $PROJECT_ID $BUILD_STATE_SUCCESS " "
+			else
+				echo "Build failed, setting build status to failed"
+				$util updateBuildState $PROJECT_ID $BUILD_STATE_FAILED " "
```
rajivnathan: We should have a message for the failure case

src/pfe/file-watcher/scripts/nodejs-container.sh
```diff
@@ -432,6 +439,13 @@ elif [ "$COMMAND" == "update" ]; then
 			$IMAGE_COMMAND exec $project /scripts/noderun.sh stop
 			$util updateAppState $PROJECT_ID $APP_STATE_STOPPING
 			$IMAGE_COMMAND exec $project /scripts/noderun.sh start $AUTO_BUILD_ENABLED $START_MODE $HOST_OS
+			if [[ $? -eq 0 ]]; then
+				echo "Build succeeded, setting build status to success"
+				$util updateBuildState $PROJECT_ID $BUILD_STATE_SUCCESS " "
+			else
+				echo "Build failed, setting build status to failed"
+				$util updateBuildState $PROJECT_ID $BUILD_STATE_FAILED " "
```
rajivnathan: We should have a message for the failure case

ssh24:
@rajivnathan This fix would solve #1993 completely. Portal does not have any role to play here and it is something Turbine needs to handle

PR #2016, opened 2020-02-04
Fix path for creating a new file in lagom project test

ssh24:
### Description

The performance time for creating a new file for lagom in our test is significantly low. This is because the file is created in a folder that is not taken into account for docker builds. We need to have the file under `hello-impl` for it to be taken into consideration.

Reran the test after this change and saw the full dockerbuild take place which takes ~3.5mins


src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts
```diff
@@ -22,6 +22,7 @@ import * as utils from "../../../lib/utils";
 import * as eventConfigs from "../../../configs/event.config";
 import * as project_configs from "../../../configs/project.config";
 import * as timeoutConfigs from "../../../configs/timeout.config";
+import { projecLanguges } from "../../../configs/app.config";
```
rajivnathan: typo `projecLanguges`

PR #2029, opened 2020-02-05
Profiling API uses ReadStream

ebuckle:
- Update the profiling API to return a readStream for both Node and Java, simplifying the API to have a consistent return
- Enables the return of profiling to work on both local and remote as files can be retrieved from PFE without the need for docker commands
- Updated tests to reflect the new output
- Inclusion of a counter in the retry loop to check for .hcd file saving to prevent spinning infinitely
- Consistency of emitted socket messages from LoadRunner

### DIFF
--- a/src/pfe/portal/modules/LoadRunner.js
+++ b/src/pfe/portal/modules/LoadRunner.js
@@ -199,7 +199,7 @@ module.exports = class LoadRunner {
         }
       }
 
-      this.user.uiSocket.emit('runloadStatusChanged', { projectID: this.project.projectID,  status: 'preparing' });
+      this.user.uiSocket.emit('runloadStatusChanged', { projectID: this.project.projectID,  status: 'preparing', timestamp: this.metricsFolder });
 
       // start profiling if supported by current language
       if (this.project.language == 'nodejs' || this.project.language === 'javascript') {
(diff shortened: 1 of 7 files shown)


src/pfe/portal/modules/Project.js
```diff
@@ -718,7 +723,7 @@ async function isHcdSaved(pathToLoadTestDir) {
   const files = await fs.readdir(pathToLoadTestDir);
```
hhellyer: This function should probably be renamed `findHcd` or `getHcdPath` to reflect what it does now.

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -309,18 +313,22 @@ module.exports = class LoadRunner {
       method: 'GET',
     };
     
-    setTimeout(() => this.getJavaHealthCenterData(options, projectID, this.metricsFolder), duration * 1000);
+    setTimeout(() => this.getJavaHealthCenterData(options, projectID, this.metricsFolder, 1), duration * 1000);
   }
 
-  async getJavaHealthCenterData(options, projectID, timestamp) {
+  async getJavaHealthCenterData(options, projectID, timestamp, counter) {
+    if (counter > 20) {
```
hhellyer: 20 should probably be replaced by a constant like `RETRY_COUNT`.

tobespc:
tested PR and can see the data coming back from the containers and appearing in the local workspace

Issue #2025, opened 2020-02-05
Add functional test for verifying -1 set on internalPort


ssh24:
**Description of the enhancement:**
We need to add functional test that sets internalPort to -1 on `.cw-settings` file that makes the app stopped. It is targetted for server less applications, but even applications that have server can set this value on the .cw-settings which will allow them to block the server from being pinged and the app will be shown stopped.

/area iterative-dev
/kind test

PR #2048, opened 2020-02-05
Add test on the ability to specify -1 on internalPort setting

ssh24:
### Description

Closes https://github.com/eclipse/codewind/issues/2025

This PR for adding functional tests on testing out the ability to specify -1 on internalPort settings. Normally it would be done for serverless applications to stop them from being pinged and eventually timedout. But it can work with apps with servers as well if you want to block the "server" functionality of that app and just have it stopped.

This PR sets -1 on internalPort settings and check if the app gets stopped. And resets it back to its original port and check if the app is now started.

### DIFF
--- a/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
+++ b/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
@@ -26,11 +26,14 @@ import { SocketIO } from "../../../lib/socket-io";
 import { fail } from "assert";
 import { Operation } from "../../../../../src/projects/operation";
 import * as projectUtil from "../../../../../src/projects/projectUtil";
+import * as constants from "../../../../../src/projects/constants";
 
 import * as utils from "../../../lib/utils";
 
 export function projectSpecificationTest(socket: SocketIO, projData: projectsController.ICreateProjectParams, projectTemplate: string, projectLang: string): void {
     describe("projectSpecification function", () => {
+        const cwSettingsPath = path.join(projData.location, ".cw-settings");
+
         const data: any = {
             "projectID": projData.projectID
         };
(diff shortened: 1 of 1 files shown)


ssh24:
@maysunfaisal Could you PTAL?

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
```diff
@@ -322,6 +325,61 @@ export function projectSpecificationTest(socket: SocketIO, projData: projectsCon
...
+                "projectID": projData.projectID,
+                "status": "success",
+                "ports": {
+                    "internalPort": `${constants.disablePingPort}`,
+                }
+            }, {"projectID": projData.projectID, "appStatus": "stopped"}];
+            await reconfigProjectSettings("internalPort", constants.disablePingPort, events, eventDatas);
+
+            const projectInfo = await projectUtil.getProjectInfo(projData.projectID);
+            await projectUtil.getContainerInfo(projectInfo, true);
```
maysunfaisal: is this required here? Not using it ..

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
```diff
@@ -322,6 +325,61 @@ export function projectSpecificationTest(socket: SocketIO, projData: projectsCon
...
+
+            events = [eventConfigs.events.settingsChanged, eventConfigs.events.statusChanged];
+            eventDatas = [{
+                "projectID": projData.projectID,
+                "status": "success",
+                "ports": {
+                    "internalPort": `${constants.disablePingPort}`,
+                }
+            }, {"projectID": projData.projectID, "appStatus": "stopped"}];
+            await reconfigProjectSettings("internalPort", constants.disablePingPort, events, eventDatas);
```
maysunfaisal: do we need to add `.timeout(timeoutConfigs.defaultTimeout);` on the func call like it is done below?

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
ssh24: We do this to update the latest container info....notice how we pass the `true` in the second arg which means force refresh and have the latest container info. We do similar things [here](https://github.com/eclipse/codewind/blob/74c6f4f4f7d6608a3ad1c018dd239bd8ca517852/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts#L458)

src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-specification.ts
ssh24: The function call does not require timeout. It will all fall under one timeout for the `it` case which is set by `.timeout(timeoutConfigs.defaultTimeout);`

Issue #2044, opened 2020-02-05
Turbine Swift Local Test Build Queue stuck at 3


maysunfaisal:
**Codewind version:**
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
Create file test is failing because build queue is full when we try to create a file for Swift project type:
```
[32m[05/02/20 04:43:13 turbinetest-default-swift] [INFO][39m Emitting event 
 message: projectStatusChanged
 data: {
  "projectID": "default-swift-1580872223504",
  "buildStatus": "queued",
  "detailedBuildStatus": "Rank 1/1"
}
Keeping PFE pod/container alive every 30s ...
[36m[05/02/20 04:43:18 Turbine] [DEBUG][39m Found at least one build in the queue
[34m[05/02/20 04:43:18 Turbine] [TRACE][39m Emitting project ranks
[34m[05/02/20 04:43:18 Turbine] [TRACE][39m Key translated. Returning message: Rank 1/1
[36m[05/02/20 04:43:18 Turbine] [DEBUG][39m Running Builds queue: ["default-swift-1580872223504","default-swift-1580872223504","default-swift-1580872223504"]
[36m[05/02/20 04:43:18 Turbine] [DEBUG][39m Builds in progress: 3
[34m[05/02/20 04:43:18 turbinetest-default-swift] [TRACE][39m Setting rank for projectID default-swift-1580872223504: 1/1
[32m[05/02/20 04:43:18 turbinetest-default-swift] [INFO][39m Build state changed for project: default-swift-1580872223504 from: queued, to: queued, with message: projectStatusController.buildRank
[32m[05/02/20 04:43:18 turbinetest-default-swift] [INFO][39m Emitting event 
 message: projectStatusChanged
 data: {
  "projectID": "default-swift-1580872223504",
  "buildStatus": "queued",
  "detailedBuildStatus": "Rank 1/1"
}
[36m[05/02/20 04:43:23 Turbine] [DEBUG][39m Found at least one build in the queue
[34m[05/02/20 04:43:23 Turbine] [TRACE][39m Emitting project ranks
[34m[05/02/20 04:43:23 Turbine] [TRACE][39m Key translated. Returning message: Rank 1/1
[36m[05/02/20 04:43:23 Turbine] [DEBUG][39m Running Builds queue: ["default-swift-1580872223504","default-swift-1580872223504","default-swift-1580872223504"]
[...]

PR #2050, opened 2020-02-05
Add a delay on setting build status so that the running build queue can be cleared

ssh24:
### Description

Closes https://github.com/eclipse/codewind/issues/2044

Have a slight delay when setting the build status which triggers the build from being removed from the running build queue. This is needed because some tests are so fast, that if we reset the build status before without waiting, the next test will overwrite it by adding it to the queue again. A next build starts before even the next 5s interval check happens so the status is now build in progress and the condition will to remove the build from running queue fails and the queue keeps getting clogged.


ssh24:
@maysunfaisal Could you PTAL?

Issue #2053, opened 2020-02-06
Lagom project modify files path broken on Functional Test


ssh24:
**Description:**
The path for modifications of file in `lagom` is broken at this [link](https://github.com/eclipse/codewind/blob/a9fc706101ca8ad85ea0bc76b5ba1eefde2222ed/src/pfe/file-watcher/server/test/functional-test/suites/project/tests/project-event.ts#L81)

This needs to be fixed only for the `creation` of a new file only where the file needs to be created under `hello-impl` folder for the entire Dockerfile build to happen.

```
  2) PFE - functional test
       Project Specific Suite: docker-lagom
         project test
           updateProjectForNewChange function
             update file with type = modify
               modify file: Dockerfile:
     Error: ENOENT: no such file or directory, open '/codewind-workspace/turbinetest-docker-lagom/hello-impl/Dockerfile'
      at Object.openSync (fs.js:443:3)
      at Object.readFileSync (fs.js:343:35)
      at Context.it (test/functional-test/suites/project/tests/project-event.ts:84:61)

  3) PFE - functional test
       Project Specific Suite: docker-lagom
         project test
           updateProjectForNewChange function
             update file with type = modify
               modify file: hello-api/src/main/java/com/example/rp/test/lagomendpoints/api/HelloService.java:
     Error: ENOENT: no such file or directory, open '/codewind-workspace/turbinetest-docker-lagom/hello-impl/hello-api/src/main/java/com/example/rp/test/lagomendpoints/api/HelloService.java'
      at Object.openSync (fs.js:443:3)
      at Object.writeFileSync (fs.js:1194:35)
      at Object.appendFileSync (fs.js:1240:6)
      at Context.it (test/functional-test/suites/project/tests/project-event.ts:93:34)

  4) PFE - functional test
       Project Specific Suite: docker-lagom
         project test
           updateProjectForNewChange function
             update file with type = modify
               modify file: hello-impl/src/main/java/com/example/rp/test/lagomendpoints/impl/HelloModule.java:
     Error: ENOENT: no such file or directory, open '/codewind-workspace/turbinetest-docker-lagom/hello-impl/hello-impl/src/main/java/com/example/rp/test/lagomendpoints/impl/HelloModule.java'
      at Object.openSync (fs.js:443:3)
      at Object.writeFileSync (fs.js:1194:35)
      at Object.appendFileSync (fs.js:1240:6)
      at Context.it (test/functional-test/suites/project/tests/project-event.ts:93:34)

[...]

PR #2055, opened 2020-02-06
Fix file path condition on new file create for lagom test

ssh24:
### Description

Closes https://github.com/eclipse/codewind/issues/2053

The filepath for creating a new file in lagom should be tailored to the `create` operation only. This PR fixes it with the new condition.


ssh24:
@maysunfaisal Could you PTAL?

PR #2069, opened 2020-02-06
Change load test defaults.

hhellyer:
Change the load test defaults to a set of lower numbers that won't swamp applications with more requests than they can handle but will run for long enough to generate profiling data.


hhellyer:
Merging now the PR has built.

PR #2123, opened 2020-02-11
Remote Profiling

ebuckle:
Ability to fetch profiling data from remote projects:
- Changing all docker exec commands used within LoadRunner.js to use cwUtils, calling either docker or kube commands depending on deployment
- Kube and Docker versions of a CopyFileFromContainer helper function in cwUtils

To be used with [457](https://github.com/eclipse/codewind-vscode/pull/457) to enable saving of remote profiling data to local workspace.

Tested with node and java projects, local and remote and received profiling data back from both.

### DIFF
--- a/src/pfe/portal/modules/LoadRunner.js
+++ b/src/pfe/portal/modules/LoadRunner.js
@@ -341,48 +341,46 @@ module.exports = class LoadRunner {
     // For Java Liberty restart the server in preparation for profiling
     if (this.project.language == 'java' && this.project.projectType == 'liberty') {
       log.info(`beginJavaProfiling: Stopping liberty server`); 
-      await docker.exec(this.project, ['bash', '-c', `source $HOME/artifacts/envvars.sh; $HOME/artifacts/server_setup.sh; cd $WLP_USER_DIR;  /opt/ibm/wlp/bin/server stop; `]);
+      const stopCommand = ['bash', '-c', `source $HOME/artifacts/envvars.sh; $HOME/artifacts/server_setup.sh; cd $WLP_USER_DIR;  /opt/ibm/wlp/bin/server stop; `];
+      await cwUtils.spawnContainerProcess(this.project, stopCommand);
       await this.waitForHealth(false);
       log.info(`beginJavaProfiling: Starting liberty server`); 
-      await docker.exec(this.project, ['bash', '-c', `source $HOME/artifacts/envvars.sh; $HOME/artifacts/server_setup.sh; cd $WLP_USER_DIR;  /opt/ibm/wlp/bin/server start;`]);
+      const startCommand = ['bash', '-c', `source $HOME/artifacts/envvars.sh; $HOME/artifacts/server_setup.sh; cd $WLP_USER_DIR;  /opt/ibm/wlp/bin/server start; `];
+      await cwUtils.spawnContainerProcess(this.project, startCommand);
       await this.waitForHealth(true);
     }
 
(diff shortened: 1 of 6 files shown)


src/pfe/portal/modules/LoadRunner.js
```diff
@@ -289,48 +289,45 @@ module.exports = class LoadRunner {
...
-      method: 'GET',
-    };
-    
-    setTimeout(() => this.getJavaHealthCenterData(options, projectID, this.metricsFolder, 1), duration * 1000);
+    const project = this.project;
+    setTimeout(() => this.getJavaHealthCenterData(project, projectID, this.metricsFolder, 1), duration * 1000);
   }
 
-  async getJavaHealthCenterData(options, projectID, timestamp, counter) {
+  async getJavaHealthCenterData(project, projectID, timestamp, counter) {
```
hhellyer: I'm a bit worried about the confusing use of `project` and `this.project` in the same function. We should pick one.
(Also we don't really need to pass `projectID` now we pass the project object, or if it should be the `projectID` for `this.project`.)

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -523,22 +520,22 @@ module.exports = class LoadRunner {
...
+      log.info(`profiling.json saved for project ${this.project.name}`);
       const data = { projectID: this.project.projectID,  status: 'profilingReady' , timestamp: this.metricsFolder }
       this.user.uiSocket.emit('runloadStatusChanged', data);
       this.profilingSocket.emit('disableprofiling');
       this.profilingSocket.disconnect();
       this.profilingSocket = null;
       this.profilingSamples = null;
+      this.project = null;
     }
-    this.project = null;
```
hhellyer: We don't null the project if there wasn't a socket? Can we run load against projects we don't create sockets for? (It looks like we only create sockets for Node projects.)

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -289,48 +289,45 @@ module.exports = class LoadRunner {
...
-
-    let options = {
-      host: "localhost",
-      port: 9090,
-      path: `/api/v1/projects/${projectID}/profiling/${this.metricsFolder}`,
-      method: 'GET',
-    };
-    
-    setTimeout(() => this.getJavaHealthCenterData(options, projectID, this.metricsFolder, 1), duration * 1000);
+    const project = this.project;
```
hhellyer: I'm not sure what this line achieves. ;-)

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -341,48 +341,42 @@ module.exports = class LoadRunner {
...
+      await this.project.getProfilingByTime(this.metricsFolder);
+    } catch (error) {
+      log.error(error);
       log.info(`.hcd file not found, trying again. Attempt ${counter}/20`);
-      setTimeout(() => this.getJavaHealthCenterData(apiOptions, projectID, timestamp, counter + 1), 3000);
-    } else {
-      log.info(".hcd copied to PFE");
-      const data = { projectID: projectID,  status: 'hcdReady', timestamp: timestamp};
-      this.user.uiSocket.emit('runloadStatusChanged', data);
+      setTimeout(() => this.getJavaHealthCenterData(counter + 1), 3000);
```
hhellyer: Could you add a null check for project here? Then we won't keep retrying 20 times if someone cancels the run and we null out the this.project object.

ebuckle: Change is in

PR #2132, opened 2020-02-11
Always re-inject metrics

hhellyer:
This PR just ensures that we always re-inject metrics on a project change.
This is necessary because we always copy all the files from our temporary upload area to the real project directory here:
https://github.com/eclipse/codewind/compare/master...hhellyer:fixOpenLibIdentify4Inject?expand=1#diff-b73b1946195409164e2ccc8a3b7fe79cR272
so the pom.xml or package.json is *always* replaced by the real one without metrics injected.
Always re-injecting actually makes the logic quite a lot simpler.

### DIFF
--- a/test/src/unit/modules/metricsService.test.js
+++ b/test/src/unit/modules/metricsService.test.js
@@ -104,7 +104,7 @@ describe('metricsService/index.js', () => {
         // console.log(util.inspect(expectedPomXmlForLiberty, { showHidden: false, depth: null }));
     });
 
-    describe('injectMetricsCollectorIntoProject(projectType, projectDir)', () => {
+    describe('injectMetricsCollectorIntoProject(projectType, project.language, projectDir)', () => {
         afterEach(() => {
             fs.removeSync(projectDir);
         });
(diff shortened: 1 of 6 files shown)


test/src/unit/modules/metricsService.test.js
```diff
@@ -206,18 +204,18 @@ describe('metricsService/index.js', () => {
...
-                    (await readXml(pathToTestPomXml))
-                        .should.deep.equal(originalPomXmlForOpenLiberty);
-                    (await readXml(pathToTestPomXml))
-                        .should.deep.equal(originalPomXmlForOpenLiberty);
+                it(`should re-inject without error`, async() => {
+                    await metricsService.injectMetricsCollectorIntoProject('docker', 'java', projectDir);
+
+                    const outputJvmOptions = fs.readFileSync(pathToJvmOptions, 'utf8');
+                    outputJvmOptions.should.equal(expectedNewJvmOptions);
+                    const outputBackupJvmOptions = fs.pathExistsSync(pathToBackupJvmOptions);
```
rwalle61: nit: `outputJvmOptions` is the contents of a jvm.options file, yet `outputBackupJvmOptions` is a boolean. Maybe rename the latter

Issue #1105, opened 2019-11-13
Project-types API sometimes fails to include AppsodyExtension


james-wallis:
**Codewind version:** Master
**OS:** MacOS

**Description:**
Project types API test is intermittently failing for unknown reasons and when it does it makes all the PR builds to fail.
I made attempts to find the issue but did not get anywhere.

**Steps to reproduce:**
Sorry, no idea. Couldn't reproduce locally.
Failing Jenkins log: https://ci.eclipse.org/codewind/job/Codewind/job/codewind/job/PR-1088/6/consoleText

**Current "fix"**
I've made a PR which removes the "appsodyExtension" check from the test. I'd like it to be put back but it as it stands when it fails it is holding up PR's from going in to Codewind even when they do not touch the templates/extensions.
PR: https://github.com/eclipse/codewind/pull/1103

tetchel:
This is the expected behaviour if there are no template sources/repositories enabled which have Appsody style projects in them.

I think it is bad behaviour, though.

makandre:
The recent failure may have something to do with us updating the version after branching off for `0.6.0`

But now the code will dynamically set the version based on the git branch, and hopefully that will mean less disruption going forward.

https://github.com/eclipse/codewind/pull/1090

tobespc:
@james-wallis can resolve this one please and test if it is better based on the comments above

james-wallis:
@makandre is this fix stable enough for the `AppsodyExtension` to be re-added to the test?

makandre:
I think we can add it back

james-wallis:
The PR I opened for adding the AppsodyExtension is still failing the test in Jenkins, locally it passes.
https://github.com/eclipse/codewind/pull/1174

james-wallis:
Still failing: https://github.com/eclipse/codewind/pull/1644

tobespc:
@james-wallis please retest and close/fix

james-wallis:
https://github.com/eclipse/codewind/pull/2066

james-wallis:
This is still failing, @hhellyer is making a change to the way that template repositories are working so I am going to keep this issue open and re-test once that change is in.

james-wallis:
The cause of this bug was that the extension download was silently failing in Jenkins and so the extension was not in the PFE container for the tests (PR build only). 
PR: https://github.com/eclipse/codewind/pull/2133

james-wallis:
[...]

PR #2133, opened 2020-02-11
Install extensions on Jenkins in PR builds, fail build if extension cannot be downloaded

tobespc:


james-wallis:
@tobespc please change the title to: `Install extensions on Jenkins in PR builds, fail build if extension cannot be downloaded`

Fix for https://github.com/eclipse/codewind/issues/1105 and https://github.com/eclipse/codewind/issues/2145
### Summary
* Add an `if` statement to check the exit code of the `curl` commands that download the Appsody and ODO extension, build will fail if the extension cannot be downloaded.
* Added functionality to use the pull request target branch when downloading the extension as currently a PR would not be able to pull a valid extension version.
* Made Jenkins build fail if extensions cannot be downloaded.
* Added the Appsody test back into the `project-types` API test.
* Added some `trace` log messages so that its easier to debug what extensions should be installed in the future.

### Testing
* Jenkins PR build passes

Issue #2140, opened 2020-02-11
capabilitiesReady is no longer provided by GET /api/v1/projects


tetchel:
0.9 and master

`capabilitiesReady` is used by the IDEs to determine if a project has been received by turbine so the IDE can request the project's capabilities. Since this field is now always undefined in GET /projects, the capabilities are never fetched for a project that was created before the current IDE session (they are still fetched for new projects since the IDE receives a `projectChanged` event that has this field set).

While we're at it, `detailedAppStatus` should be provided by this API too. https://github.com/eclipse/codewind/issues/1856#issuecomment-583505870

https://github.com/eclipse/codewind/blob/master/src/pfe/portal/modules/Project.js#L339

I expect this was broken by some combination of https://github.com/eclipse/codewind/pull/1981 and https://github.com/eclipse/codewind/pull/1941 so pinging @hhellyer and @stalleyj. 
It looks like the fields not saved into the inf are now the same as the fields not provided by this API at all, while `capabilitiesReady` and `detailedAppStatus` should still be provided, but not saved.

Marking as stopship because this disables restart actions for all existing projects after restarting the IDE.

tetchel:
😎

PR #2147, opened 2020-02-12
[0.9.0] Include non-persisted fields in JSONified Project object.

hhellyer:
We were stripping out fields we didn't want to persist to the project .inf files the same way as we were stripping out fields we couldn't convert to JSON.
This was simple but meant the non-persistent fields were never returned via /api/v1/projects as that just converts the Project object to JSON.
This PR takes advantage of the replacer() function you can pass when writing a JSON file to remove the non-persistent fields.

This is for issue https://github.com/eclipse/codewind/issues/2140 and ports the fix to 0.9.0

### DIFF
--- a/src/pfe/portal/modules/Project.js
+++ b/src/pfe/portal/modules/Project.js
@@ -322,8 +322,10 @@ module.exports = class Project {
       throw new ProjectError('LOCK_FAILURE', this.name);
     }
     try {
-      // Fields that shouldn't be persisted are stripped out by toJSON() below.
-      await fs.writeJson(infFile, this, { spaces: '  ' });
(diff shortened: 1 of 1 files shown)


hhellyer:
This PR is linking to the ECA check https://accounts.eclipse.org/legal/eca/validation/26284 which is actually for PR https://github.com/eclipse/codewind/pull/2146 - That PR existed briefly but I accidentally targetted it at master instead of the 0.9.0 branch. Somehow the ECA check for that PR is targetting this one.

The actual ECA check for this PR is:
https://accounts.eclipse.org/legal/eca/validation/26285
and doesn't quite seem to be working.

Issue #2139, opened 2020-02-11
Convert the build queue logic to have queues that are set instead of arrays


ssh24:
**Description of the enhancement:**
We have been trying to do a lot of checks to validate the current implementation of queues are unique or not. I believe if we used `set` datastructure instead of `arrays`, this issue would not exist. Opening this enhancement to convert our existing queue datastructure from arrays to queues.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

/area iterative-dev

PR #2164, opened 2020-02-12
Transform build queues from arrays to set datastructure

ssh24:
### Description

Closes https://github.com/eclipse/codewind/issues/2139

This PR converts the existing datastructure for both the type of build queues - running and waiting, to set datastructure. This allows us to have a unique set of builds in each queue and removes the overhead to check if the queues are always unique.

### DIFF
--- a/src/pfe/file-watcher/server/src/controllers/projectsController.ts
+++ b/src/pfe/file-watcher/server/src/controllers/projectsController.ts
@@ -49,7 +49,7 @@ const fileStatAsync = promisify(fs.stat);
 
 const projectInfoCache = {} as ProjectInfoCache;
 
-let buildQueue: Array<BuildQueueType> = [];
+const buildQueue: Set<BuildQueueType> = new Set();
 let runningBuilds: Array<BuildQueueType> = [];
 const MAX_BUILDS = parseInt(process.env.MC_MAX_BUILDS) || 3;
 const BUILD_KEY = "projectStatusController.buildRank";
@@ -434,14 +434,12 @@ export async function addProjectToBuildQueue(project: BuildQueueType): Promise<v
 
     await lock.acquire("buildQueueLock", done => {
         // Check if project is already in the build queue, if it is then no need to add it
-        target = buildQueue.find((item) => {
-            return item.operation.projectInfo.projectID === project.operation.projectInfo.projectID;
-        });
-        if (target) {
+        if (buildQueue.has(project)) {
+            target = project;
             logger.logProjectInfo("The project is already in the build queue so it won't be added again", project.operation.projectInfo.projectID);
         } else {
             logger.logProjectInfo("Pushing project to build queue", project.operation.projectInfo.projectID);
-            buildQueue.push(project);
+            buildQueue.add(project);
         }
         done();
     }, async () => {
@@ -466,7 +464,7 @@ async function checkBuildQueue(): Promise<void> {
         // assert the builds in progress is always between 0 to MAX_BUILDS inclusive
         logger.assert(runningBuilds.length >= 0 && runningBuilds.length <= MAX_BUILDS, "Builds in progress must be between [0, MAX_BUILDS]");
         // if we have builds in the queue
-        buildQueueLength = buildQueue.length;
+        buildQueueLength = buildQueue.size;
         if (buildQueueLength > 0) {
             logger.logDebug("Found at least one build in the queue");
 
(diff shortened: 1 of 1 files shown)


src/pfe/file-watcher/server/src/controllers/projectsController.ts
```diff
@@ -585,32 +579,24 @@ async function triggerBuild(project: BuildQueueType, changedFiles?: IFileChangeE
...
         });
 
-        if (runningBuilds.length > 0) {
+        if (runningBuilds.size > 0) {
             // Only log the runningBuilds details if there is a project building to avoid spamming the logs
             const currentBuilds: Array<String> = [];
-            for (let i = 0; i < runningBuilds.length; i++) {
-                currentBuilds.push(runningBuilds[i].operation.projectInfo.projectID);
-            }
+            runningBuilds.forEach((project: BuildQueueType) => {
```
rajivnathan: Do we need `currentBuilds`? Looks like it's only used for the log message below. Can we use `runningBuilds` directly for the message instead?

src/pfe/file-watcher/server/src/controllers/projectsController.ts
```diff
@@ -627,13 +613,13 @@ function checkInProgressBuilds(): void {
...
     logger.logTrace("Emitting project ranks");
     await lock.acquire("buildQueueLock", async done => {
-        buildQueue = buildQueue.filter(temp => {
-            return temp != undefined;
-        });
-        buildQueue.forEach( async (project, index) => {
+        // convert the build queue set into an inplace array for project rank emition that relies on the element index
+        // sets are not great for storing elements in any order, so this is required for the project ranks which relies on position
+        const buildQueueArray = [...buildQueue];
+        buildQueueArray.forEach( async (project, index) => {
```
rajivnathan: I thought we discovered that sets in JS stay in insertion order which is why we can use it as a queue with unique values. Perhaps the actual problem is the use of `forEach` here? If we iterate over the array with a standard for loop instead and use `buildQueue.values().next()` in each iteration it should work? Then we don't have the overhead of converting from set to array.

src/pfe/file-watcher/server/src/controllers/projectsController.ts
```diff
@@ -713,18 +698,16 @@ export async function deleteProject(projectID: string): Promise<IDeleteProjectSu
...
-                logger.assert((initialLength - buildQueue.length) === 1, "If project was deleted, the difference in build queue length should be exactly one");
+                logger.assert((initialLength - buildQueue.size) === 1, "If project was deleted, the difference in build queue length should be exactly one");
             }
         }, {});
 
         await lock.acquire("runningBuildsLock", async done => {
             // remove the project from in-progress queue only if the project deleted was in the in-progress build
-            runningBuilds = runningBuilds.filter((project: BuildQueueType) => {
-                if (project.operation.projectInfo.projectID === projectID) {
+            runningBuilds.forEach((buildQueueItem: BuildQueueType) => {
```
rajivnathan: Should we change `buildQueueItem` to `runningQueueItem` since this is looping through the running queue?

src/pfe/file-watcher/server/src/controllers/projectsController.ts
ssh24: The reason we use `currentBuilds` is because when we log the running builds queue, we only show the `projectID` instead of the entire build object which is of type BuildQueueType. `runningBuilds` is also of type BuildQueueType so we just take the projectID from each project and print them

src/pfe/file-watcher/server/src/controllers/projectsController.ts
ssh24: The insertion do stay in order however there is no indexing in sets. So we can't just do `buildQueue[0]` to get the first element. Since for ranking we need to know the exact position (number) of them, we need to transform it into an array.

The way you described `forEach` don't return the index on sets. It will only give me the element. And sets don't have an easy way to get the "index" of an element

src/pfe/file-watcher/server/src/controllers/projectsController.ts
ssh24: Sure, I can rename that. I probably copied it from the `buildQueue` for each that is why it has the same name

src/pfe/file-watcher/server/src/controllers/projectsController.ts
rajivnathan: This is what I was suggesting:
```
const set1 = new Set();
set1.add('projectA');
set1.add('projectB');

const iterator1 = set1.values();

for(let i=1; i<(set1.size+1); i++) {
  console.log(iterator1.next().value + ' ' + i + '/' + set1.size);  
}
```

Output:
```
> "projectA 1/2"
> "projectB 2/2"
```
It's more efficient than copying to an array and then iterating.

src/pfe/file-watcher/server/src/controllers/projectsController.ts
rajivnathan: The extra information could be beneficial and it's for debug logs so won't pollute the logs by default. But I'm okay if you'd rather have only the projectID.

src/pfe/file-watcher/server/src/controllers/projectsController.ts
ssh24: Applied those changes and tested it out. PTAL
Screen Shot 2020-02-13 at 12 26 32 PM

PR #2210, opened 2020-02-14
[0.9.0] Perf Dashboard - Fix Inline loadtest description

markcor11:
## Problem : 

Fixes a problem where when editing a load run description in the history table, the change was not saved after a page refresh.  Issue : https://github.com/eclipse/codewind/issues/2211

## Solution :

Used the snapshot ID rather than the start time to indent the correct record in the collection. 


markcor11:
Too late for 0.9.0, closing PR - fix went into master

PR #2222, opened 2020-02-14
Increase combo timeout for getWorkspaceSettingsInfo function unit test

ssh24:
### Description

Related to https://github.com/eclipse/codewind/issues/1928

Signed-off-by: ssh24 <sakib@ibm.com>

PR #2242, opened 2020-02-18
Adjust timeouts for failing (and similar) tests.

hhellyer:
We were seeing intermittent timeouts on different tests in the Turbine test suite on Jenkins:
https://ci.eclipse.org/codewind/job/Codewind/job/codewind/job/master/

This PR just raises the timeout on the tests we have seen fail due to timeouts and similar tests in the same files using the pattern @ssh24 already used in a previous PR: https://github.com/eclipse/codewind/pull/2222/files


hhellyer:
This was replaced by https://github.com/eclipse/codewind/pull/2256

PR #2249, opened 2020-02-18
Refactor noderun.sh script to use sh rather than bash

james-wallis:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
* Changes the `noderun.sh` script that turbine uses to start Node.js containers to use `sh` rather than `bash` to support containers that do not have bash, such as `alpine`.

### Summary
* Use `awk` rather than a for loop and bash variable stuff to determine the cached `PID`.
* Use the `/proc` directory in linux to determine what processes are running and gather data around them.
* Change bash variables

### Testing
* Created a default Node.js application that uses ubuntu to verify that it is not broken by this change.
* Imported a Node.js project running on `node:alpine-8`
* Ran file-watcher server unit and functional tests (took 3.5hrs)
* Ran PFE test suite.

afaik no as we don't say the prereqs for running Node as it is.

@rajivnathan fyi

### DIFF
--- a/src/pfe/file-watcher/scripts/nodejs-container.sh
+++ b/src/pfe/file-watcher/scripts/nodejs-container.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 #*******************************************************************************
 # Copyright (c) 2019 IBM Corporation and others.
 # All rights reserved. This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


james-wallis:
@rajivnathan would you be able to review or assign someone from the Turbine team to review it please?

Issue #2254, opened 2020-02-18
Have a script that can compare two or more csv file for performance data


ssh24:
**Description of the enhancement:**
As part of https://github.com/eclipse/codewind/issues/2091 we are setting up the cronjob. We want to have a scrip that can easily compare two or more csv for the performance data as a baseline approach to see if the performance deteriorated or improved.

/area iterative-dev
/kind test

Issue #2228, opened 2020-02-14
Post cleanup argument doesn't work when pre clean up argument is set to "n" on functional test


ssh24:
**Description:**
If I want only post clean up of my tests but already have a codewind up and want to run without any pre clean up, the post clean up won't work. This is because we only do post clean up when there is a pre clean up according to the code:

https://github.com/eclipse/codewind/blob/16e5fd9e21d711a72422e7c435144f95997f16c4/src/pfe/file-watcher/server/test/scripts/runner.sh#L98-L129

https://github.com/eclipse/codewind/blob/16e5fd9e21d711a72422e7c435144f95997f16c4/src/pfe/file-watcher/server/test/scripts/runner.sh#L204-L205

We should support post clean up even when pre cleanup is set to "n".

/area iterative-dev
/kind test

Issue #2175, opened 2020-02-13
Skip run of generic suites for turbine performance test


ssh24:
**Description of the enhancement:**
We should skip the run of generic suites for our performance test run. This can be done by using the performance test env variable and passing in false value to the generic test run https://github.com/eclipse/codewind/blob/f9373feabbc56e39a6807575fa2c66135df6ef2d/src/pfe/file-watcher/server/test/functional-test/functional.test.ts#L78

/area iterative-dev
/kind test

Issue #2173, opened 2020-02-13
Allow ability bypass of TLS verify on oc login for functional tests


ssh24:
**Description of the enhancement:**
We don't allow the ability to bypass of TLS verification on `oc login` for functional tests so for secured clusters, the test gets stuck on here:
```
>> Creating project turbinetest-odo-perl from https://github.com/sclorg/dancer-ex in /root/codewind/src/pfe/file-watcher/server/test/performance-test/performance-test-codewind/src/pfe/file-watcher/server/test/projects/turbinetest-odo-perl 
{“status”:“success”,“projectPath”:“/root/codewind/src/pfe/file-watcher/server/test/performance-test/performance-test-codewind/src/pfe/file-watcher/server/test/projects/turbinetest-odo-perl”,“result”:{“language”:“unknown”,“projectType”:“docker”}}
✔ Done. 

✔ Done. 

/root/codewind/src/pfe/file-watcher/server/test/performance-test/performance-test-codewind/src/pfe/file-watcher/server/test
The server uses a certificate signed by an unknown authority.
You can bypass the certificate check,
```
We can just set ` --insecure-skip-tls-verify=true` by default but that my not be the best solution. We could add a flag to the test script which the user can enable to skip the TLS verification on the oc login dynamically.

/area iterative-dev
/kind test

PR #2298, opened 2020-02-24
Turbine performance test cronjob

ssh24:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?

This PR sets up the performance test cronjob for turbine module. It also adds an utility that generates a report on the performance run. In addition to that, it also fixes some additional test bugs listed below.

#2091 
#2254 
#2175 
#2173 
#2228


ssh24:
@maysunfaisal Could you PTAL?

Issue #2056, opened 2020-02-06
javametrics-dash plotting data outside of axes for garbage collection


tobespc:
running a java app under load show this graph

![Screenshot 2020-02-06 at 09 00 26](https://user-images.githubusercontent.com/13151286/73921595-8ecd6400-48bf-11ea-9b36-0d485d7f3937.png)

gc graph needs normalizing

tobespc:
@mattcolegate please pick this up

mattcolegate:
@tobespc says this graph came from a performance run, 1 user making 1 request a second for 180 seconds of /v1/example, with the following addition to the code:
```java
 @GET
    @Produces(MediaType.TEXT_PLAIN)
    public Response example() {
        List<String> list = new ArrayList<>();
        //return a simple list of strings
        for (int i = 0; i < 100000; i++) {
            String blah[] = new String[i];
            for (int z=0; z < i; z++) {
                blah[z] = "tobes";
            }
        }
        list.add("Congratulations, your application is up and running");
        return Response.ok(list.toString()).build();
    }
```

mattcolegate:
Valid issue.
/pipeline portal backlog

mattcolegate:
This isn't a case of a graph not scaling - the data being displayed has entries > 100, which for a stat representing percentages isn't great! Continuing investigation.

mattcolegate:
Highest data point seen: 1689.6551724137933

mattcolegate:
Change agreed was to update the graph to plot summary data, as individual sample data was proving unreliable under heavy load. PRs above achieve that for injected and vanilla java projects.
/pipeline Review/QA

mattcolegate:
/pipeline in progress
After PR review, there is more to do here - now that we know that the GCTime metric can be dodgy, the decision has been taken to remove the generation of that metric altogether so that we know our customers are not affected by erroneous data. This will include:

- commenting out the system calls that create the data item
- changing the various files that export that data (mostly via JSON)
- changing the files that receive that data and make use of the GCTime metric
- updating any GCTime graphs to use GCSummaryTime instead
- checking the documentation accurately reflects this change.

mattcolegate:
/pipeline Review/QA

The code has now all been put into PRs: here's how to merge it - 

1. Review and merge https://github.com/RuntimeTools/graphmetrics/pull/59
1. Release graphmetrics (or at least, update javametrics to the new commit hash).
[...]

PR #2306, opened 2020-02-25
{WIP} Update GC Graph to summarize gc time since app start

mattcolegate:

## What type of PR is this ? 

- [X] Bug fix
- [ ] Enhancement

## What does this PR do ?
It updates the GC graph on injected metrics to display the percentage of time spent in GC since the start of the application, rather than every sampling period

https://github.com/eclipse/codewind/issues/2056
No

Remember to inject to see this working


mattcolegate:
This PR is no longer required due to solving the underlying problem a different way

Issue #2027, opened 2020-02-05
cpu graph plotting data outside of range


tobespc:
I'm monitoring an open liberty application after injecting metrics and can see that the cpu graph is plotting outside the range of the axis
![Screenshot 2020-02-05 at 09 35 53](https://user-images.githubusercontent.com/13151286/73829559-14d3a700-47fb-11ea-806f-905fc1fb04e8.png)

tobespc:
@mattcolegate please pick this up

mattcolegate:
Valid issue.
/pipeline portal backlog

stalleyj:
We also see the GC graph plotting outside the axes.

mattcolegate:
@stalleyj that issue covered separately under https://github.com/eclipse/codewind/issues/2056

mattcolegate:
This seems to be restricted to injected projects only - can't recreate this with normal javametrics

mattcolegate:
The cause of this is down to the data collectors for injected projects using the value of base_cpu_systemLoadAverage from /metrics as input for the system load graph.
```
# HELP base_cpu_systemLoadAverage Displays the system load average for the last minute. The system load average is the sum of the number of runnable entities queued to the available processors and the number of runnable entities running on the available processors averaged over a period of time. The way in which the load average is calculated is operating system specific but is typically a damped time-dependent average.
```
This is why high numbers are registered, depending on the number of entities. It's not a percentage indication of system load.
Will raise PRs to stop the use of that metric for system load.

PR #2310, opened 2020-02-26
Stop using base cpu system load average for CPU graph

mattcolegate:

## What type of PR is this ? 

- [X] Bug fix
- [ ] Enhancement

## What does this PR do ?
Stops using the metric `base_cpu_systemloadAverage` as input to the CPU graph on the Metrics dashboard

Fixes https://github.com/eclipse/codewind/issues/2027
https://github.com/eclipse/codewind/issues/2027
No

No

### DIFF
--- a/src/performance/monitor/public/java-metrics.html
+++ b/src/performance/monitor/public/java-metrics.html
@@ -222,7 +222,7 @@
         }
         if (['base:cpu_process_cpu_load_percent', 'base_cpu_processCpuLoad_percent', 'process_cpu_used_ratio '].includes(metric.name)) {
           cpuData.process = numMetricValue;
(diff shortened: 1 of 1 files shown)


mattcolegate:
The cause of this is down to the data collectors for injected projects using the value of base_cpu_systemLoadAverage from /metrics as input for the system load graph.

# HELP base_cpu_systemLoadAverage Displays the system load average for the last minute. The system load average is the sum of the number of runnable entities queued to the available processors and the number of runnable entities running on the available processors averaged over a period of time. The way in which the load average is calculated is operating system specific but is typically a damped time-dependent average.

This is why high numbers are registered, depending on the number of entities. It's not a percentage indication of system load.

The CPU process metrics are, and have been verified to be, a percentage measure of CPU used by the process and are therefore correct.

PR #2268, opened 2020-02-19
[WIP] Slightly refactor Template API test

james-wallis:
## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
* Attempts to remove the intermittent bug from the Template API tests by resetting the template repositories before each second level `describe` block.
* Also changes the `POST` `/api/v1/templates/repositories` (add a repository) to use the test one by deleting it first rather than another one which had the potential to cause errors + further upkeep.

## Does this PR require a documentation change ?
No.

PR #2346, opened 2020-02-28
Templates API - 423 Locked

james-wallis:
## What type of PR is this ? 

- [x] Bug fix - Tests
- [x] Enhancement - Lock

## What does this PR do ?
Changes the implementation of `Template.js` to update the repository list and template list in the `add repository` function rather than on a `get repositories` or `get templates` to resolve the race conditions that can appear when doing multiple `GET` requests. 
Uses a `lock` and returns `423 Locked` if the templates file is `locked`.

#### Templates
* Added `lock` to `Templates.js`
* Move the functions which fetch the templates and update the repository list into the `add`, `delete` functions rather than the `get`
  * `add` and `delete` will take longer as they do more (`POST`, `DELETE)
  * `get` should be a lot quicker (`GET`)
  * removed the `projectListNeedsUpdating` and the `templatesNeedUpdating` variables.
* Refactor functions which didn't need to be in the Class into local functions.
  * This will further lock down the use of them.
  * All functions within the class have the locking mechanism.
* Refactor `batchUpdate` to not fetch a new template list but to `enable` or `disable` based on the templates respective repository.
  * Added a `sourceURL` field to the template object due to this.

#### ExtensionList
* Changed the `addExtensionsToTemplates` function to be a `for` loop to resolve problems with the the Templates being locked when an `async` function was used. 
* As this is only called when Codewind is intialised it currently makes more sense to make this change rather than refactor much of the logic to cope with the lock -> In the future if enhance PFE to add extensions while its running (API?) then we'll need a more complex lock.

### Testing
* Added tests for new functions in Template.js
* Added `setupReposAndTemplatesForTesting` function which when included will set the `repositoryList` to only contain the testing Codewind repository.
  * The Appsody repository will be still show as it cannot be deleted.
* Enhanced the `addProvider` tests now that more logic has been added.

Takes some testing logic from https://github.com/eclipse/codewind/pull/2268/files which this PR supersedes.

#### Appsody test fix for `repository.yaml` in `.appsody` directory that does not include the `incubator` repository
* Removed the Appsody checks from the `project-types` and `templates styles` API tests.
* Added an Appsody check into the `extension` API test.

Only API docs -> I have made that change in the `openapi.yml`

This PR supersedes the previous on on the `409-conflict` branch, diff here: https://github.com/james-wallis/codewind/compare/409-conflict...james-wallis:423-locked
This PR is a rebased, cleaned up version.

### DIFF
--- a/src/pfe/portal/modules/Templates.js
+++ b/src/pfe/portal/modules/Templates.js
@@ -45,144 +45,103 @@ const DEFAULT_REPOSITORY_LIST = [
 ];
 
 module.exports = class Templates {
-
   constructor(workspace) {
     // If this exists it overrides the contents of DEFAULT_REPOSITORY_LIST
     // One list for enabled templates, another that includes the disabled templates.
     this.enabledProjectTemplates = [];
     this.allProjectTemplates = [];
-    // If a repository is added or removed then update the template list on the next GetTemplates
-    this.projectTemplatesNeedsRefresh = true;
-    // If a repository is added or removed update the repository list
-    this.projectRepositoriesNeedsRefresh = true;
     this.repositoryFile = path.join(workspace, '.config/repository_list.json');
     this.repositoryList = DEFAULT_REPOSITORY_LIST;
     this.providers = {};
+    this.unlock();
   }
 
   async initializeRepositoryList() {
+    this.lock();
     try {
-      let repositories = [...this.repositoryList];
+      let repositories = cwUtils.deepClone(this.repositoryList);
       const repositoryFileExists = await cwUtils.fileExists(this.repositoryFile);
       if (repositoryFileExists) {
         repositories = await fs.readJson(this.repositoryFile);
-        this.projectTemplatesNeedsRefresh = true;
       }
       repositories = await updateRepoListWithReposFromProviders(this.providers, repositories, this.repositoryFile);
       repositories = await fetchAllRepositoryDetails(repositories);
       this.repositoryList = repositories;
+
+      // Fetch templates
+      const { enabledTemplates, allTemplates } = await fetchTemplates(this.repositoryList);
+      this.enabledProjectTemplates = enabledTemplates;
+      this.allProjectTemplates = allTemplates;
+
       await writeRepositoryList(this.repositoryFile, this.repositoryList);
     } catch (err) {
       log.error(`Error initializing repository list: ${err}`);
+    } finally {
+      this.unlock();
     }
+  }
+
+  lock() {
+    if (this._lock === true) throw new TemplateError('LOCKED');
+    this._lock = true;
+  }
+
+  unlock() {
+    this._lock = false;
   }
 
   // TEMPLATES
-
-  async getTemplates(enabledOnly) {
-    if (this.projectTemplatesNeedsRefresh) {
-      const enabledRepositories = await this.getEnabledRepositories();
-      const disabledRepositories = await this.getDisabledRepositories();
-      const newEnabledTemplates = await getTemplatesFromRepos(enabledRepositories);
-      const newDisabledTemplates = await getTemplatesFromRepos(disabledRepositories);
-      this.enabledProjectTemplates = newEnabledTemplates;
-      this.allProjectTemplates = newEnabledTemplates.concat(newDisabledTemplates);
-      this.projectTemplatesNeedsRefresh = false;
+  getTemplates(enabledOnly) {
+    this.lock();
+    try {
+      return enabledOnly ? this.enabledProjectTemplates : this.allProjectTemplates;
+    } finally {
+      this.unlock();
     }
-    return enabledOnly ? this.enabledProjectTemplates : this.allProjectTemplates;
   }
 
-  async getTemplatesByStyle(projectStyle, showEnabledOnly = false) {
-    const templates = await this.getTemplates(showEnabledOnly);
-    return filterTemplatesByStyle(templates, projectStyle);
+  getTemplatesByStyle(projectStyle, enabledOnly = false) {
+    this.lock();
+    try {
+      const templates = enabledOnly ? this.enabledProjectTemplates : this.allProjectTemplates;
+      return filterTemplatesByStyle(templates, projectStyle);
+    } finally {
+      this.unlock();
+    }
   }
 
   async getAllTemplateStyles() {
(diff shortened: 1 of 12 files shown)


src/pfe/portal/modules/Templates.js
```diff
@@ -546,3 +545,113 @@ function isTemplateSummary(obj) {
...
+  }
+}
+
+// Updates repositories in the repositoryList with the updated values
+// New repositories will not be added
+// Repositories with no updates will be kept with no changes
+function updateRepositoryList(currentRepositoryList, updatedRepositories) {
+  const newRepositoryList = currentRepositoryList.map(oldRepo => {
+    const modifiedRepo = updatedRepositories.find(modifiedRepo => modifiedRepo.url === oldRepo.url);
+    if (modifiedRepo) {
```
jcockbain: as above, think you could do `return modifiedRepo || oldRepo`

src/pfe/portal/routes/projectTypes.route.js
```diff
@@ -11,10 +11,13 @@
 const express = require('express');
 
 const Logger = require('../modules/utils/Logger');
+const TemplateError = require('../modules/utils/errors/TemplateError');
 
 const router = express.Router();
 const log = new Logger(__filename);
 
+const TEMPLATES_LOCKED_STATUS_CODE = 423;
```
jcockbain: could you write this constant once, and export it (probably utils)? It's duplicated across 2 routes

src/pfe/portal/routes/templates.route.js
```diff
@@ -36,6 +38,11 @@ router.get('/api/v1/templates', validateReq, async (req, res, _next) => {
       res.status(200).json(templates);
     }
   } catch (error) {
+    log.error(error);
```
jcockbain: double log 😮

src/pfe/portal/routes/templates.route.js
```diff
@@ -86,31 +105,51 @@ router.delete('/api/v1/templates/repositories', validateReq, async (req, res, _n
     if (error instanceof TemplateError && error.code === 'REPOSITORY_DOES_NOT_EXIST') {
       res.status(404).send(error.message);
       return;
+    } else if (error instanceof TemplateError && error.code === 'LOCKED') {
```
jcockbain: might be clearer to check instanceof once, and then then the error code inside that

test/modules/template.service.js
```diff
@@ -225,6 +234,24 @@ function saveReposBeforeEachTestAndRestoreAfterEach() {
...
+    before(async function() {
+        this.timeout(testTimeout.med);
+        // Save original repositories
+        const { body: getBody } = await getTemplateRepos();
+        originalTemplateRepos = getBody;
+        await setTemplateReposTo([sampleRepos.codewind]);
+    });
+    after(async function() {
+        this.timeout(testTimeout.med);
+
```
jcockbain: nit: blank line here?

test/src/API/templates/templates.test.js
```diff
@@ -37,24 +37,19 @@ chai.use(deepEqualInAnyOrder);
...
-        before(async function() {
-            this.timeout(testTimeout.short);
-            await setTemplateReposTo([
-                { ...sampleRepos.codewind },
-            ]);
-        });
+        setupReposAndTemplatesForTesting();
         describe('?projectStyle=', function() {
             it('should return at least the default Codewind templates as no projectStyle is given', async function() {
-                this.timeout(testTimeout.short);
```
jcockbain: are you sure this is always OK in Jenkins, and never times out when getting templates?

test/src/unit/template.test.js
```diff
@@ -73,164 +92,155 @@ describe('Templates.js', function() {
...
-                this.timeout(testTimeout.short);
-                const templateController = new Templates('');
+                const templateController = new Templates(workspace);
                 const { repositoryList } = templateController;
                 for (const repository of repositoryList) {
                     repository.enabled = false;
                     repository.enabled.should.be.false;
                 }
+                await templateController.initializeRepositoryList();
                 const enabledTmplateList = await templateController.getTemplates(true);
```
jcockbain: not you, but tmplate

test/src/API/templates/templates.test.js
jcockbain: Oops, see you're setting it for all tests. Nice 👍

test/src/unit/template.test.js
```diff
@@ -864,18 +762,41 @@ describe('Templates.js', function() {
...
         });
-        describe('updateRepoListWithReposFromProviders(providers, repositoryList, repositoryFile)', function() {
+        describe('constructRepositoryObject(url, description, name, isRepoProtected)', () => {
+            const constructRepositoryObject = Templates.__get__('constructRepositoryObject');
+            it('returns a repository object when all arguments are given', async() => {
+                const { url, description, name, protected } = sampleRepos.codewind;
+                const repository = await constructRepositoryObject(url, description, name, protected);
+                repository.should.have.keys('id', 'name', 'url', 'description', 'enabled', 'protected', 'projectStyles');
+            });
+            it('returns a repository object only a url is given - gets the name, description from the url and does not have a protected field', async() => {
```
jcockbain: reword message here

jcockbain: probably just add when

src/pfe/portal/modules/Templates.js
```diff
@@ -546,3 +545,113 @@ function isTemplateSummary(obj) {
...
+      if (provider.canHandle(copy) && typeof provider.removeRepository === 'function')
+        promises.push(provider.removeRepository(copy));
+    }
+  }
+  await Promise.all(promises);
+  return providers;
+}
+
+async function performOperationsOnRepositoryList(requestedOperations, repositoryList) {
+  // const newRepositoryList = cwUtils.deepClone(repositoryList);
```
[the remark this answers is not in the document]
james-wallis: 👀

src/pfe/portal/modules/Templates.js
james-wallis: 🔥

src/pfe/portal/routes/templates.route.js
james-wallis: 🙄

PR #2370, opened 2020-03-02
Cancel Java Profiling

ebuckle:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
Fixes some issues that are seen when trying to cancel load runs on Java projects.

Cancelling while the load test is preparing currently has no effect. 
Cancelling during a load run leaves the Java health center agent still gathering information and writing a .hcd file and this can sometimes be copied down to the users local workspace after cancelling.

### DIFF
--- a/src/pfe/portal/modules/User.js
+++ b/src/pfe/portal/modules/User.js
@@ -27,6 +27,7 @@ const FilewatcherError = require('./utils/errors/FilewatcherError');
 const log = new Logger('User.js');
 const util = require('util');
 const { installBuiltInExtensions } = require('./utils/installExtensions');
+const retry = require('async-retry')
 
 /**
  * The User class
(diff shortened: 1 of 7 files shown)


rwalle61:
@hhellyer @stalleyj could https://github.com/eclipse/codewind/pull/2352 be merged before this PR please, making it easier for @ebuckle to add tests for this?

src/pfe/portal/modules/User.js
```diff
@@ -183,21 +184,46 @@ module.exports = class User {
...
+      return promiseRetry({retries: 15, minTimeout: 2000, maxTimeout: 2000}, function (retry, number) {
+        log.info(`Attempting to cancel load for project ${project.projectID}. Attempt ${number}`);
+        return that.callCancelRunLoad(project)
+          .catch(function (err) {
+            retry(err);
+          });
+      })
+        .then(function (value) {
+          project.loadInProgress = false;
+          that.uiSocket.emit('runloadStatusChanged', { projectID: project.projectID, status: 'cancelled' });
```
james-wallis: Instead of doing `const that = this` I think you should be able use arrow functions to pass the `this` context down:
```
return cancelLoadResp;	      return promiseRetry({retries: 15, minTimeout: 2000, maxTimeout: 2000}, (retry, number) => {
    log.info(`Attempting to cancel load for project ${project.projectID}. Attempt ${number}`);
    return that.callCancelRunLoad(project)
      .catch(function (err) {
        retry(err);
    });
})
```

james-wallis: Also curious why you chose `promise-retry` over `async-retry` (which I notice is in the `package.json`) when `async-retry` means we can do `async/await` rather than conventional promises

tobespc:
I've run it, and whilst it does work I've hit some issues

  - If I cancel a load run, and then as soon as I can, request a new one.  I keep seeing the message about my project not running.  If I wait a few moments, I can then request a load run
  - Temp files are left in the codewind-workspace 
```
[root@283d6d3e27cf load-test]# find . -print
.
./20200310142039
./20200310142123
./20200310142123/tmp_100320_142146_
./20200310142123/tmp_100320_142146_/trace
./20200310142123/tmp_100320_142146_/methoddictionary
./20200310142123/tmp_100320_142146_/memorycounters
./20200310142123/tmp_100320_142146_/environment
./20200310142123/tmp_100320_142146_/locking
./20200310142123/healthcenter100320_142146_9323_1.hcd
./20200310142123/metrics.json
./config.json
./20200310142328
```

ebuckle:
Temp files are now being tidied up.

The ordering of socket emits have also been changed so the UI should remain on `Cancelling` until the project is completely cancelled.

PR #2396, opened 2020-03-04
Fix comparable baseline and filename for cross baseline comparisons

ssh24:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
We need to fix the comparable baseline directory and filename for across baseline comparisons. This includes not only comparison across different baselines but also setting the new baseline of the particular release.


ssh24:
@maysunfaisal Could you PTAL?

PR #2431, opened 2020-03-09
Fix creation of directory when setting new baseline

ssh24:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
- Fixes the directory for creating a new baseline
- Adds the ability to pass in specific cwctl version to install from


ssh24:
@maysunfaisal Could you PTAL?

Issue #2435, opened 2020-03-09
VS Code: Error initialing a build, Unexpected token T in JSON at position 0


makandre:
**Codewind version:** master
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**

When requesting a manual build, getting this popup in VS Code:

![image](https://user-images.githubusercontent.com/17596240/76239606-3b28a000-6208-11ea-8f43-b25e136baabd.png)

From what I can tell PFE always returned text (not JSON) https://github.com/eclipse/codewind/blob/master/src/pfe/portal/routes/projects/build.route.js#L46 but IDE now expects JSON?

**Steps to reproduce:**

**Workaround:**

tetchel:
it is the same as https://github.com/eclipse/codewind/issues/2434

jopit:
Rather than do this in a scattershot fashion, we should take a more consistent approach.  One possibility is to move all this function into cwctl.

This would involve producing a new version of the API, rather than replacing the existing one.

Julie will set up a discussion with TimE, Erin and JohnP to move this forward.

tetchel:
I'm fixing it on my end since I originally thought the problem was isolated to the build endpoint, but there are several more.

PR #2440, opened 2020-03-09
Make project build endpoint return JSON

tetchel:
Now ~all~ more endpoints used by the IDEs return JSON.

Fixes https://github.com/eclipse/codewind/issues/2435

Signed-off-by: Tim Etchells <timetchells@ibm.com>

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
The build endpoint seems to be the only one used by the IDEs which returns plain text instead of JSON. Also changed the error responses to provide a bit more info and use more descriptive status codes.

https://github.com/eclipse/codewind/issues/2435

I have updated the openapi.yml.

## Other notes for reviewers
I checked with Erin and John P and this change will not break Intellij or Eclipse

### DIFF
--- a/src/pfe/portal/routes/projects/build.route.js
+++ b/src/pfe/portal/routes/projects/build.route.js
@@ -31,20 +31,31 @@ async function buildProject(req, res) {
     const user = req.cw_user;
     const project = user.projectList.retrieveProject(id);
     const action = req.sanitizeBody('action');
+
     if (!project) {
-      res.status(400).send(`Unable to find project ${id}`);
+      res.status(404).json({
+        status: "failed",
+        error: `Action ${action} failed: Unable to find project ${id}`,
+        projectID: id,
+        action,
+      });
       return;
     }
 
     if (project.isClosed() || project.isClosing() || project.isDeleting()) {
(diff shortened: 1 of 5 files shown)


tetchel:
@tobespc can you or a team member review?

src/pfe/portal/routes/projects/build.route.js
```diff
@@ -35,17 +35,22 @@ async function buildProject(req, res) {
       res.status(400).send(`Unable to find project ${id}`);
       return;
     }
-    
+
     if (project.isClosed() || project.isClosing() || project.isDeleting()) {
       const msg = `Cannot perform build action ${action} on ${project.name} because it is in state ${project.state}.`;
       res.status(400).send(msg);
```
makandre: Error 400 will still send back plain text?

tetchel: thx

tetchel:
Hold off, there are a couple more endpoints i'd like to update.

tetchel:
There are too many to change; I'm going to work around it on my end, but i'd still like to put this in if we can.

tobespc:
If we want the api's to return JSON, then we should do this across the board for consistency.  That would also mean the api's will need a version change as well

Issue #2052, opened 2020-02-05
Default status for appsody projects shouldn't be [Starting][Build Succeeded]


johnmcollier:
When Appsody projects are created, either locally or on Kube, their default app status is always: `[Starting] [Build Succeded]`, even before any build has started, and long before the app is starting. (See https://github.com/eclipse/codewind-appsody-extension/blob/master/entrypoint.sh#L118).

Now, I know that this is at least partially a limitation on the Appsody CLI in that we can't easily get build and app status notifications from it, but I still feel that our default status shouldn't be `[Starting] [Build Succeded]`, as it's misleading and confusing to the user. Could we change the default state to something more meaningful?

makandre:
What would you suggest?  We in fact do not go through a codewind build phase for appsody projects and invoke “appsody run” immediately, hence it is “starting” in that sense.

johnmcollier:
> What would you suggest? We in fact do not go through a codewind build phase for appsody projects and invoke “appsody run” immediately, hence it is “starting” in that sense.

Do the states necessarily need to adhere to the Codewind-style app/build statuses that we have? Maybe we need some appsody-specific states that work better for Appsody projects? I believe we did that for Odo projects.

I'm not sure on _what_ exactly we should use, but I don't think starting as the default app state, or " Build succeeded" as the build state is right here.

makandre:
I see what ODO is doing, it defined the strings in Codewind itself:

https://github.com/eclipse/codewind/blob/master/src/pfe/file-watcher/server/src/utils/locales/en/translation.json#L30

I suppose we could do the same for Appsody but I'm hesitant to put Appsody specific strings into Codewind.  But it would be ok if the strings were generic, perhaps `[Staring][Build not required]`? (I still think `starting` is still correct in the context of Appsody but I'm opened to other suggestions)

tetchel:
I've asked this before; isn't the build status for Appsody always "succeeded", even though that isn't really true because there is no build?

We could always hide the build status for Appsody projects.

makandre:
Ok as long as it's not `if (it's appsody) then hide it`

tetchel:
how else would we do it :(

johnmcollier:
Can the appsody extension just not set a build status?

E.g. https://github.com/eclipse/codewind-appsody-extension/blob/master/entrypoint.sh#L122

tetchel:
That makes sense to me and the IDEs can hide the build status if it's unknown, the same way we do for app status.

[...]

PR #2446, opened 2020-03-10
Suppress build status for appsody projects

makandre:
## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?

For #2052 we want to stop displaying build status for appsody projects, since it is misleading as appsody projects do not go through a build phase.

Turbine by default emits several "in progress" build statuses when spinning up projects, including appsody ones.  This PR adds a property to the `ShellExtensionProject.ts` class which Turbine uses to suppress those build statuses where appropriate.

For #2052 

https://github.com/eclipse/codewind-docs/pull/473


makandre:
Seems to be some problems with the changes I'm closing this PR

PR #2463, opened 2020-03-11
Introduce buildByExtension so appsody project won't get stuck in build queue

makandre:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?

For #2052 we want to stop displaying build status for appsody projects, since it is misleading as appsody projects do not go through a build phase.  To do that, appsody extension will return the default `unknown` build status to Codewind.

But on Codewind side, there's now a problem that appsody projects are not removed from the build queue because the code always expects build success or failure.  This PR fixes that.

For #2052 

https://github.com/eclipse/codewind-docs/pull/473


GeekArthur:
All PR tests are passed, code coverage test fails since the code coverage decreases with the change, that should be fine.

Issue #2472, opened 2020-03-12
Templates API sometimes returning templates from disabled sources with showEnabledOnly=true


tetchel:
Master branch

When I GET `http://localhost:10000/api/v1/templates?showEnabledOnly=true` I get all the templates, even the ones from disabled sources. I see the same behaviour using the `cwctl` templates commands (the kabanero source is disabled, but its templates are still listed).

This means the user gets prompted with the entire list of templates every time they want to create a project, and disabling a source in the Source Manager has no effect.

```
[ /programs/codewind-vscode/dev ] 28 (testsInJenkins) $ cwctl templates repos ls
[
        {
                "description": "Codewind project templates help you create containerized projects for various runtimes.",
                "url": "https://raw.githubusercontent.com/codewind-resources/codewind-templates/master/devfiles/index.json",
                "name": "Default templates",
                "id": "",
                "enabled": true,
                "protected": true,
                "projectStyles": [
                        "Codewind"
                ]
        },
        {
                "description": "Kabanero, an open source project, brings together open source technologies into a microservices-based framework. Kabanero builds cloud native applications ready for deployment onto Kubernetes and Knative.",
                "url": "https://github.com/kabanero-io/collections/releases/latest/download/kabanero-index.json",
                "name": "Kabanero Collections",
                "id": "",
                "enabled": false,
                "protected": true,
                "projectStyles": [
                        "Appsody"
                ]
        },
        {
                "description": "Use Appsody in Codewind to develop applications with sharable technology stacks.",
                "url": "https://github.com/appsody/stacks/releases/latest/download/incubator-index.json",
                "name": "Appsody Stacks - incubator",
                "id": "incubator",
                "enabled": true,
                "protected": true,
                "projectStyles": [
                        "Appsody"
                ]
        }
]
[ /programs/codewind-vscode/dev ] 28 (testsInJenkins) $ cwctl templates ls --showEnabledOnly
[
        {
                "label": "Appsody Eclipse MicroProfile® template",
                "description": "Eclipse MicroProfile on Open Liberty \u0026 OpenJ9 using Maven",
                "language": "java",
                "url": "https://github.com/appsody/stacks/releases/download/java-microprofile-v0.2.24/incubator.java-microprofile.v0.2.24.templates.default.tar.gz",
                "projectType": "appsodyExtension",
                "projectStyle": "Appsody",
                "source": "Appsody Stacks - incubator",
                "sourceId": "incubator"
        },
        {
[...]

PR #2549, opened 2020-03-25
Fix problem with adding a new template repository

stalleyj:
Signed-off-by: Julie Stalley <julie_stalley@uk.ibm.com>

## What type of PR is this ? 

- [ X] Bug fix
- [ ] Enhancement

## What does this PR do ?
This PR fixes a problem with adding a new repository. We were incorrectly updating our cache of enabled templates with the cache of all templates so subsequent requests to get templates would give an incorrect list. I also discovered a discrepancy between our api docs and the route as we were not accepting a value for "repoEnabled" so it was not possible to add a repository in a disabled state. This may not be a user requirement (would be weird to add and not enable) but was necessary for the tests. 

https://github.com/eclipse/codewind/issues/2472

API docs have been updated, no external doc change required.

### DIFF
--- a/src/pfe/portal/modules/Templates.js
+++ b/src/pfe/portal/modules/Templates.js
@@ -418,14 +418,9 @@ function updateTemplates(repositories, allTemplates) {
 async function fetchTemplates(repositories) {
   const enabledRepositories = repositories.filter(repo => repo.enabled);
   const disabledRepositories = repositories.filter(repo => !repo.enabled);
-  log.info(`enabledRepos=${enabledRepositories}`);
-  log.info(`disabledRepos=${disabledRepositories}`);
   const enabledTemplates = await getTemplatesFromRepos(enabledRepositories);
-  log.info(`enabledTemplates=${enabledTemplates.length}`);
   const disabledTemplates = await getTemplatesFromRepos(disabledRepositories);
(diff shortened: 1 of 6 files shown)


test/modules/template.service.js
```diff
@@ -140,6 +156,19 @@ async function getNumberOfTemplates(queryParams) {
     throw new Error(`getNumberOfTemplates - Unknown status code received: ${statusCode}`);
 }
 
+async function getNumberOfEnabledTemplates(queryParams) {
+    const { statusCode, body: templates } = await getEnabledTemplates(queryParams);
```
james-wallis: Instead of adding the `getEnabledTemplates` function can't we just call `getTemplates` with the queryParams:
```
queryParams = { showEnabledOnly: true };
```

src/pfe/portal/modules/Templates.js
```diff
@@ -278,13 +279,17 @@ async function validateRepository(repoUrl, repositories) {
...
+  if (enabled == undefined) {
+    enabled = true;
+  }
   let repository = {
     id: uuidv5(url, uuidv5.URL),
     name,
     url,
     description,
-    enabled: true,
+    enabled: enabled,
```
james-wallis: can just be 

```
{
    description,
    enabled,
}
```

PR #2552, opened 2020-03-26
Enable two Codewind Docker projects to talk to each other (Networking drop 1)

james-wallis:
A few bugs in this code are fixed and functionality changed in network drop 2 https://github.com/eclipse/codewind/pull/2571 but I've kept this to keep the code changes as small as possible (but still big)
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Adds support for communication between Codewind applications on from Docker to Docker, and initial implementation of Local to a Remote Codewind. -> Through environment variables:
* For *Node, Liberty and Spring* the environment variables exist only in the application process in the Docker container to take advantage of file-watchers fast restarting and to not make adding a link slower than making a code change.
* For *Swift, generic Docker types (Go, Python) and Appsody (different PR) the environment variables will be added through the `docker run --env-file` so they will be visible in the `docker inspect`.

### Summary

**Portal Changes**
* Adds support for one project to be able to communicate via environment variables to another Codewind project.
* Adds an API (/api/v1/projects/:id/links) to use get, add, remove and update the links
* Adds a `Link` class in a new `project` directory to keep the size of `Project.js` the same (ish):
    * Had to add helper functions so that the Links can write to the `Project inf` when updated.
* Each Project will have a links object that can contain references to many projects, best to see design docs.

* Other changes:
    * Added a middleware function to check whether a project exists - this can be used in the future to remove code duplicated on all endpoints that use the get project. Also added a function to easily get the project from `req`. 
    * Added a docs refresh to the `pushLocal.sh` script for API changes.

**Turbine Changes**
* Changed the start scripts for Node, Spring and Liberty to `export` all the variables in a network env file (only when the file exists).
* Changed the `docker run` for Swift to add a `--env-file` (only when the file exists)
* Changed the `docker run` command in Docker utils to add an `--env-file` command (only when the file exists)

**Testing**
* Manually tested each project type in the default templates to ensure that the environment variables are picked up.
* Added API and Unit tests for the Portal functions.

Have updated the design docs: https://github.com/codewind-resources/design-documentation/blob/master/codewindServer/ProjectNetworking.md
Will require a docs change when the final parts of the networking are PR'd into master.

### DIFF
--- a/src/pfe/portal/modules/Project.js
+++ b/src/pfe/portal/modules/Project.js
@@ -23,6 +23,7 @@ const ProjectMetricsError = require('./utils/errors/ProjectMetricsError');
 const Logger = require('./utils/Logger');
 const LogStream = require('./LogStream');
 const metricsService = require('./metricsService');
+const Links = require('./project/Links');
 
 const log = new Logger(__filename);
 
@@ -129,6 +130,8 @@ module.exports = class Project {
       this.injectMetrics = args.injectMetrics;
     }
     this.capabilitiesReady = false;
+    
+    this.links = new Links(this.projectPath(), args.links);
   }
 
 
@@ -783,12 +786,28 @@ module.exports = class Project {
   /**
    * Resolves the given path against the project's monitor path.
    * The path is unchanged if it is already absolute.
-   * 
+   *
    * @param {String} path
    * @returns {String} an absolute path
    */
   resolveMonitorPath(path) {
     return isAbsolute(path) ? cwUtils.convertFromWindowsDriveLetter(path) : join(this.pathToMonitor, path);
+  }
+
+  // Project Link wrappers so they can access the writeInformationFile function
+  async createLink(newLink) {
+    await this.links.add(newLink);
+    return this.writeInformationFile();
+  }
+
+  async updateLink(envName, newEnvName, newProjectURL) {
+    await this.links.update(envName, newEnvName, newProjectURL);
+    return this.writeInformationFile();
+  }
+
+  async deleteLink(envName) {
+    await this.links.delete(envName);
+    return this.writeInformationFile();
   }
 }
 
--- a/src/pfe/portal/modules/project/Links.js
+++ b/src/pfe/portal/modules/project/Links.js
@@ -25,7 +25,7 @@ class Links {
     this._links = (args && args._links) ? args._links : [];
     this.filePath = path.join(directory, ENV_FILE_NAME);
   }
-  
+
   getAll() {
     return this._links;
   }
@@ -76,7 +76,7 @@ class Links {
   }
 }
 
-const validateLink = (newLink, links) => {
+function validateLink(newLink, links) {
   const { projectID, parentPFEURL, envName, projectURL, type } = newLink;
   // Only require a parentPFEURL and projectURL if the link is remote
   // If the link is local then the parentPFEURL is not required and the projectURL can be undefined
@@ -93,12 +93,12 @@ const validateLink = (newLink, links) => {
   return newLink;
 }
 
-const envNameExists = (links, envName) => {
+function envNameExists(links, envName) {
   const envList = links.map(link => link.envName);
   return envList.includes(envName);
 }
 
-const writeEnvironmentFile = async (filePath, envPairArray) => {
+async function writeEnvironmentFile(filePath, envPairArray) {
   const fileContents = envPairArray.join('\n');
   await fs.writeFile(filePath, fileContents);
 }
--- a/src/pfe/portal/modules/project/Links.js
+++ b/src/pfe/portal/modules/project/Links.js
@@ -1,0 +1,111 @@
+/*******************************************************************************
+ * Copyright (c) 2020 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v2.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v20.html
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+const fs = require('fs-extra');
+const path = require('path');
+
+const ProjectLinkError = require('../utils/errors/ProjectLinkError');
+const Logger = require('../utils/Logger');
+const log = new Logger(__filename);
+
+const ENV_FILE_NAME = '.codewind-project-links.env'; // This will be in the users container
+
+/**
+ * The Links class
+ */
+class Links {
+  constructor(directory, args) {
+    this._links = (args && args._links) ? args._links : [];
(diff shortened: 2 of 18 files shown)


src/pfe/portal/modules/Project.js
```diff
@@ -783,13 +786,29 @@ module.exports = class Project {
...
-   * 
+   *
    * @param {String} path
    * @returns {String} an absolute path
    */
   resolveMonitorPath(path) {
     return isAbsolute(path) ? cwUtils.convertFromWindowsDriveLetter(path) : join(this.pathToMonitor, path);
   }
+
+  // Project Link wrappers so they can access the writeInformationFile function
```
hhellyer, line 797: I'm confused by this, User.js and ProjectList.js also call writeInformationFile() already. (I'm not sure if that's a good thing and whether we should rely on callers to write the file to disk or not really though.)

src/pfe/portal/modules/project/Links.js
```diff
@@ -0,0 +1,111 @@
...
+const log = new Logger(__filename);
+
+const ENV_FILE_NAME = '.codewind-project-links.env'; // This will be in the users container
+
+/**
+ * The Links class
+ */
+class Links {
+  constructor(directory, args) {
+    this._links = (args && args._links) ? args._links : [];
```
hhellyer, line 25: We discussed whether it would be better index the `links` structure by env var name. This can be changed in a follow up PR if you think it might be easier.

src/pfe/portal/routes/projects/links.route.js
```diff
@@ -0,0 +1,221 @@
...
+  try {
+    const { cw_user: user } = req;
+    const project = getProjectFromReq(req);
+    await project.deleteLink(envNameOfLinkToDelete);
+
+    // Send status and then kick off the restart/rebuild
+    res.sendStatus(204);
+
+    if (project.isOpen()) {
+      await restartOrBuildProject(user, project);
```
hhellyer, line 133: Can this throw an exception? (I think it can further down in the restart code.) If so that would result in calling `res.sendStatus()` again and we'd get the message we usually get about not being able to set headers after they have been sent.

src/pfe/portal/routes/projects/links.route.js
```diff
@@ -0,0 +1,221 @@
...
+    // If newEnvName or newProjectURL are not given through the API, default them to their old values
+    const updatedLinkEnvName = (newEnvName) ? newEnvName : currentEnvName;
+    const updatedLinkProjectURL = (newProjectURL) ? newProjectURL : currentProjectURL;
+    await project.updateLink(currentEnvName, updatedLinkEnvName, updatedLinkProjectURL);
+
+    // Send status and then kick off the restart/rebuild
+    res.sendStatus(204);
+
+    if (project.isOpen()) {
+      await restartOrBuildProject(user, project);
```
hhellyer, line 110: See comment below about throwing an exception here.

src/pfe/portal/routes/projects/links.route.js
```diff
@@ -0,0 +1,221 @@
...
+  }
+  // Attempt to get container name fall back to docker network IP address (given to us by file-watcher)
+  const container = await dockerInspect(project);
+  const host = (container && container.Name) ? container.Name.substring(1) : dockerNetworkIP;
+  return `${host}:${internalPort}`;
+}
+
+const restartOrBuildProject = async(user, project) => {
+  const { projectType } = project;
+  const projectTypesThatPickUpEnvsThroughRestart = ['nodejs', 'liberty', 'spring'];
```
hhellyer, line 163: Is this going to work in Kubernetes? I think we have to rebuild node apps to restart them there.

james-wallis: No at the moment this is Docker only

src/pfe/portal/modules/project/Links.js
james-wallis: I'll change this in a follow up PR as it would clash with next pr.

src/pfe/portal/modules/Project.js
james-wallis: `Links.js` is contained within the Project object whereas `User.js` and `ProjectList.js` contain the project objects meaning `Links.js` cannot access the `writeInformationFile()` function. So i've added a wrapper in `Project.js` to ensure it gets written every time a link is changed.

src/pfe/portal/routes/projects/links.route.js
james-wallis: You're right, I'll change this

src/pfe/portal/routes/projects/links.route.js
hhellyer: That's fine - though should we do something like disable the links routes if PFE is running on Kubernetes for the moment?

src/pfe/portal/modules/Project.js
hhellyer: Ok, that makes sense. I missed that difference. We don't want to add back links from links to projects as that might trigger circular refs in the project.in file.
This is much better. ;-)

src/pfe/portal/routes/projects/links.route.js
james-wallis: I can do that

src/pfe/portal/routes/projects/links.route.js
```diff
@@ -0,0 +1,257 @@
...
+    const { Config: { Env: containerEnvs }} = container;
+    const linksExistInContainer = await checkIfEnvsExistInArray(project, containerEnvs);
+    // Only build and run the project if the links are not in the container
+    if (!linksExistInContainer) {
+      log.info(`Rebuilding ${name} to pick up network environment variables`);
+      await user.buildAndRunProject(project);
+    }
+  } else {
+    // if a build is in progress, wait 5 seconds and try again
+    await cwUtils.timeout(5000);
```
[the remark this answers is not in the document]
james-wallis: afaik I'm doing the same, waiting 5 seconds and then calling the `restartDocker` function again

src/pfe/portal/routes/projects/links.route.js
```diff
@@ -0,0 +1,257 @@
...
+      newProjectURL = localProjectURL;
+    }
+
+    // If newEnvName or newProjectURL are not given through the API, default them to their old values
+    const updatedLinkEnvName = (newEnvName) ? newEnvName : currentEnvName;
+    const updatedLinkProjectURL = (newProjectURL) ? newProjectURL : currentProjectURL;
+    await project.updateLink(currentEnvName, updatedLinkEnvName, updatedLinkProjectURL);
+
+    // Send status and then kick off the restart/rebuild
+    res.sendStatus(204);
```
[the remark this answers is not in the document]
james-wallis: If 202 is what we normally do then I will change it. I'll just double check though

PR #2571, opened 2020-03-30
[WIP] Enhance the ability for two Docker projects to communicate (Networking drop 2)

james-wallis:
Follows https://github.com/eclipse/codewind/pull/2552
View the diff of 2552 and this PR: https://github.com/james-wallis/codewind/compare/project-link...james-wallis:project-link-proxy

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Adds a proxy to PFE to allow Codewind links to direct their requests to PFE which PFE can then forward on to their intended destination.
  * This solves the issue with changing ports/addresses.
  * PFE will check its ProjectList each time a request comes in to ensure that the project exists (404 otherwise) and then gets the target application's current URL rather than saving it directly into the project's file system and having to do a manual update - causing the project to restart in the process.

### Summary
* Updates the `openapi.yml` to have the correct documentation for these changes.
* Fixes bug from Networking drop 1 where the env file would not be added into the project container for Node, Spring and Liberty before a restart.
* Updates the tests

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -1753,6 +1753,95 @@ paths:
                     example: "/Dockerfile-tools"
        500:
           description: Internal Server Error
+  /api/v1/projects/{id}/links:
+    get:
+      summary: Returns the links for a project
+      parameters:
+        - in: path
+          name: id
+          schema:
+            $ref: '#/components/schemas/ProjectID'
+          required: true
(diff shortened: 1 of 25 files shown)


james-wallis:
Closing as I've created another branch which does not have the proxy.

Issue #2284, opened 2020-02-20
SVT: [Remote] Performance Dashboard UI can present the load test stuck in "Requested"


sujeilyfonseca:
**Codewind version:** 0.9.0
**OS:** VM with Windows 10 (w. OKD/Openshift cluster)

**IDE extension version:** 0.9.0
**IDE version:** 1.42.0

**Description:** @jagraj and I noticed that the Performance Dashboard UI can present the load test stuck in "Requested". If users try to cancel the request, they will not be able to do so:

<img width="1671" alt="Screen Shot 2020-02-20 at 4 58 04 PM" src="https://user-images.githubusercontent.com/49531960/74986220-1e6a2b00-5407-11ea-92fc-05445cbd71b4.png">

However, the load test runs successfully. Users can see the test results, if they open a new window:

<img width="1655" alt="Screen Shot 2020-02-20 at 5 11 37 PM" src="https://user-images.githubusercontent.com/49531960/74986131-f7135e00-5406-11ea-90a1-b5a323b8cd90.png">

Also, if they attempt to run new load tests in the new window, they will not see the problem again:

<img width="1617" alt="Screen Shot 2020-02-20 at 5 32 51 PM" src="https://user-images.githubusercontent.com/49531960/74986197-1316ff80-5407-11ea-90f4-7c7e0b77c15a.png">

```
[root@sfonseca-okd-codewind ~]# oc  logs  codewind-performance-k6v9sv51-84f57c6cd7-hkgmw 

> performanceserver@1.0.0 start /usr/src/app
> PORT=9095 node server.js

[20/02/20 21:42:36 /usr/src/app/server.js] [INFO] Performance server listening on port 9095!
[20/02/20 21:58:56 /usr/src/app/server.js] [INFO] starting loadrun with options : {"path":"/","requestsPerSecond":"1","concurrency":"1","maxSeconds":"180","url":"http://172.30.209.241:9080/"}
[20/02/20 22:01:57 /usr/src/app/server.js] [INFO] completed - loadrun summary : { totalRequests: 180,
  totalErrors: 0,
  totalTimeSeconds: 180.005692361,
  rps: 1,
  meanLatencyMs: 6.1,
  maxLatencyMs: 67,
  minLatencyMs: 4,
  percentiles: { '50': 5, '90': 6, '95': 7, '99': 11 },
  errorCodes: {},
  instanceIndex: 0 }

[...]

PR #2574, opened 2020-03-31
create load-test dir if doesn't exist [0.10.0]

tobespc:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
When load test runs, the output produced (metrics.json) is sometimes failing to write to the container due the the output directory not having been created yet.  This change checks the directory exists and creates if needed.

It also contains a change to await for the creation of collection

#2284

no
no

### DIFF
--- a/src/pfe/portal/modules/LoadRunner.js
+++ b/src/pfe/portal/modules/LoadRunner.js
@@ -178,13 +178,8 @@ module.exports = class LoadRunner {
           metricsJson.desc = this.runDescription;
         }
         try {
-          // eslint-disable-next-line no-sync
-          if (!fs.existsSync(this.workingDir)) {
-            // eslint-disable-next-line no-sync
-            fs.mkdirSync(this.workingDir);
-          }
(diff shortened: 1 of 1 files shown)


src/pfe/portal/modules/LoadRunner.js
```diff
@@ -178,7 +178,13 @@ module.exports = class LoadRunner {
           metricsJson.desc = this.runDescription;
         }
         try {
-          await fs.writeJson(this.workingDir + '/metrics.json', metricsJson, { spaces: '  ' });
+          // eslint-disable-next-line no-sync
```
hhellyer: So long as we've got the awaits correct there's no need to move to the Sync versions of these methods. It looks like you fixed not awaiting for the whole function below. There's also no need to check if the directory exists, just do:
```
await fs.ensureDir(this.workingDir);
await fs.writeJson(this.workingDir + '/metrics.json', metricsJson, { spaces: '  ' });
```

PR #2621, opened 2020-04-07
[0.11.0] Determine name of profiling file before attempting copy

stalleyj:
## What type of PR is this ? 

- [X] Bug fix
- [ ] Enhancement

## What does this PR do ?
This PR changes the way we copy profiling information from the project pod back to the docker volume. We were copying the load-test/date directory across but this would overwrite the metrics.json file.  We are now determining the name of the hcd file and copying just that.  

This PR is part of the solution for https://github.com/eclipse/codewind/issues/2284

No


tobespc:
testing on kube.....

tobespc:
its all working on kube

PR #2671, opened 2020-04-14
Ensure async/await tests are returned, await fs.writeJson in LoadRunning

james-wallis:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
Fixes tests that were failing locally for me.
* Removed `Fails to get a profiling file using an invalid time stamp` from `getProfilingByTime` as I couldn't find which section would return a `NOT_FOUND` error code. (previously wasn't returned so mocha didn't await it).
* Added a cwUtils mock function for `findHCDfile` as we don't want to call `docker` in the unit test.

### DIFF
--- a/src/pfe/portal/modules/LoadRunner.js
+++ b/src/pfe/portal/modules/LoadRunner.js
@@ -433,7 +433,7 @@ module.exports = class LoadRunner {
     let hashObject = { gitHash: hashValue };
     const filePath = path.join(this.workingDir, 'runinfo.json');
     try {
-      fs.writeJson(filePath, hashObject, { spaces: '  ' });
+      await fs.writeJson(filePath, hashObject, { spaces: '  ' });
     } catch (err) {
       log.error(err);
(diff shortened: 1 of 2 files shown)


src/pfe/portal/modules/LoadRunner.js
```diff
@@ -433,7 +433,7 @@ module.exports = class LoadRunner {
     let hashObject = { gitHash: hashValue };
     const filePath = path.join(this.workingDir, 'runinfo.json');
     try {
-      fs.writeJson(filePath, hashObject, { spaces: '  ' });
+      await fs.writeJson(filePath, hashObject, { spaces: '  ' });
```
rwalle61, line 436: I wonder why our test for this function in LoadRunner.test.js was passing even though we weren't `await`ing this. Maybe just luck, because the test does an `await fs.readJson` after calling this function

Issue #2699, opened 2020-04-16
SVT : Mac  : Cannot create or import Appsody Kitura projects 


mmulholl:
**Codewind version:**0.11.0.v202004091756
**OS:**Mac

**Che version:**
**IDE extension version:**
**IDE version:**Eclipse jee 2019-09, VSCode
**Kubernetes cluster:**

**Description:**
Appsody kitura project will not create or import using eclipse or VSCode. Wizards do not show an error but the project fails to appear in the codewind project list. For example see this in the pfe logs:

[16/04/20 16:05:42 /portal/routes/projects/remoteBind.route.js] [ERROR] Project creation failed for project appsodyKituraTest011. Error: { [Error: EACCES: permission denied, open ‘/mounted-workspace/appsodyKituraTest011/.appsody/AppsodyKitura/Tests/LinuxMain.swift’]
 errno: -13,
 code: ‘EACCES’,
 syscall: ‘open’,
 path:
  ‘/mounted-workspace/appsodyKituraTest011/.appsody/AppsodyKitura/Tests/LinuxMain.swift’ }

**Steps to reproduce:**
Try creating a new appsodt kiutra project

**Workaround:**

sujeilyfonseca:
**Codewind version:** 0.11.0
**OS:** Mac OS Mojave 10.14.6, VM with Windows 10, Linux/Ubuntu

**IDE extension version:** 0.11.0
**IDE version:** Code 1.44.0

I reproduced the problem on Mac, using Appsody Kitura. I attempted to create/import the project, but nothing happened:

<img width="1672" alt="Screen Shot 2020-04-16 at 1 07 28 PM" src="https://user-images.githubusercontent.com/49531960/79490329-affeaf00-7fea-11ea-8977-a06329efc4b7.png">

I was not able to reproduce the problem on Linux and Windows:

![Screen Shot 2020-04-16 at 1 18 11 PM](https://user-images.githubusercontent.com/49531960/79490351-b725bd00-7fea-11ea-8351-a015b83d3c4b.png)

![Screen Shot 2020-04-16 at 1 19 20 PM](https://user-images.githubusercontent.com/49531960/79490356-b856ea00-7fea-11ea-8d8e-00b16b38d86d.png)

![Screen Shot 2020-04-16 at 1 19 36 PM](https://user-images.githubusercontent.com/49531960/79490359-b8ef8080-7fea-11ea-92ec-41437e035232.png)

[...]

PR #2701, opened 2020-04-16
0.11.0: preserve only in upload end

makandre:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?

Follow-up to PR https://github.com/eclipse/codewind/pull/2696, preserve timestamp only in `upload/end` and not in `bind/end`

#2699

### DIFF
--- a/src/pfe/portal/modules/utils/sharedFunctions.js
+++ b/src/pfe/portal/modules/utils/sharedFunctions.js
@@ -85,9 +85,9 @@ module.exports.updateObject = function updateObject(objectToUpdate, fieldsToAddT
   return objectToUpdate;
 }
 
-module.exports.copyProject = async function copyFile(fromProjectPath, toProjectPath, mode) {
+module.exports.copyProject = async function copyFile(fromProjectPath, toProjectPath, mode, preserve = false) {
   log.debug(`copyProject fromPath: ${fromProjectPath}, toPath: ${toProjectPath}`);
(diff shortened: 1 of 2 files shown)


makandre:
@elsony 

This PR revert the behaviour for `bind/end`, so essentially the effective change in both PRs is now in `upload/end`, we preserve the timestamp.

So I tested with Codewind spring project as an example, and in fact it appears the behaviour now is improved than before.  Without preserving timestamp, regardless of what file I changed (even if it's just a file in `src/test/java`, I saw this in the maven build log for the `default-compile` phase:

```
[INFO] --- maven-compiler-plugin:3.1:compile (default-compile) @ cwspring02 ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 4 source files to /root/app/./target/classes
```

After the change to preserve timestamp, if I changed a file in `src/test/java`, I see this now:

```
[INFO] --- maven-compiler-plugin:3.1:compile (default-compile) @ cwspring02 ---
[INFO] Nothing to compile - all classes are up to date
```

So I think we were actually compiling more than necessary before because when the project is copied from `cw-temp`, all the timestamps are touched and appears updated.

hhellyer:
@elsony - I did a quick check locally instrumenting the upload REST function. Only the changed files are be being uploaded. I'm going to approve and merge this fix so it's ready for 0.11.0.

PR #2702, opened 2020-04-16
master: preserve only in upload end

makandre:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?

Follow-up to PR #2695, preserve timestamp only in upload/end and not in bind/end

#2699


makandre:
Closing since @hhellyer found it was an `fs-extra` bug and the proper fix is to upgrade that package. See https://github.com/eclipse/codewind/issues/2699#issuecomment-615175252 and PR #2705

PR #2703, opened 2020-04-17
Add support for Appsody networking in Kubernetes

james-wallis:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
* Adds support for Appsody in Kubernetes. To pass envs through to Appsody we need to send them as `docker-options` (its an Appsody thing) and rebuild the project (what we do for local). 
* The new function will check if the Appsody deployment already has the networking envs, if it doesn't it will rebuild the project to pick up links.

Note: This needs https://github.com/eclipse/codewind-appsody-extension/pull/97 to land before it can be tested easily (you can copy the change into your running PFE container to test).

### DIFF
--- a/src/pfe/portal/routes/projects/links.route.js
+++ b/src/pfe/portal/routes/projects/links.route.js
@@ -165,9 +165,14 @@ async function getProjectURL(project) {
 
 async function restartProjectToPickupLinks(user, project) {
   if (global.codewind.RUNNING_IN_K8S) {
-    // In K8s we use configmaps and restart Pods
-    await updateNetworkConfigMap(project);
-    await restartDeployment(project);
+    const { extension } = project;
+    if (extension) {
+      await restartAppsodyKubernetes(user, project);
+    } else {
+      // In K8s we use configmaps and restart Pods for our templates
+      await updateNetworkConfigMap(project);
+      await restartDeployment(project);
+    }
   } else {
     // In Docker we use an env file or export into a container process
     const { projectType } = project;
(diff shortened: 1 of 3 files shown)


src/pfe/portal/routes/projects/links.route.js
```diff
@@ -165,9 +165,14 @@ async function getProjectURL(project) {
 
 async function restartProjectToPickupLinks(user, project) {
   if (global.codewind.RUNNING_IN_K8S) {
-    // In K8s we use configmaps and restart Pods
-    await updateNetworkConfigMap(project);
-    await restartDeployment(project);
+    const { extension } = project;
+    if (extension) {
```
hhellyer, line 169: How will this work with other extension types? (Like ODO!)

src/pfe/portal/routes/projects/links.route.js
```diff
@@ -270,4 +275,50 @@ function checkIfEnvsExistInArray(project, array) {
   return envPairs.every(env => array.includes(env));
 }
 
+
+async function restartAppsodyKubernetes(user, project) {
```
hhellyer, line 279: Should we have extension specific code here? Does this belong in the appsody extension?

james-wallis: Already added the code to the Appsody extension, unfortunately Appsody just needs to be updated differently in Kuberentes. Think the function should be `restartExtension` as each extension will need to be restarted using its extension framework rather than using Codewind specific stuff.

src/pfe/portal/routes/projects/links.route.js
james-wallis: Right... Odo, I'll need to check this

Issue #2692, opened 2020-04-15
SVT: PFE container died after creating several projects


sujeilyfonseca:
**Codewind version:** 0.9.0
**OS:** MacOS

**IDE extension version:** 0.9.0
**IDE version:** 1.44.0

**Description:** I installed Codewind 0.9.0 on VS Code because I wanted to perform a migration scenario. I created several projects, and suddenly the connection was lost: 

<img width="1669" alt="Screen Shot 2020-04-13 at 2 23 53 PM" src="https://user-images.githubusercontent.com/49531960/79148860-39687400-7d94-11ea-9377-e278f58e80c8.png">

I tried a refresh, but it didn't work:

<img width="1669" alt="Screen Shot 2020-04-13 at 2 25 42 PM" src="https://user-images.githubusercontent.com/49531960/79148867-3c636480-7d94-11ea-8578-8f5618724c08.png">

I verified the running pods, and the `codewind-pfe` was no longer running:

```
(base) Sujeilys-MBP:IBM.codewind sfonseca$ docker ps
CONTAINER ID        IMAGE                                                     COMMAND             CREATED             STATUS              PORTS                       NAMES
e9895c9fe0f4        cw-cwdefaultpython-c1732160-7db2-11ea-b3c6-e746faddb2be   "python app.py"     13 minutes ago      Up 13 minutes       127.0.0.1:32776->5000/tcp   cw-cwdefaultpython-c1732160-7db2-11ea-b3c6-e746faddb2be
f23fc0102afc        eclipse/codewind-performance-amd64:0.9.0                  "npm start"         21 minutes ago      Up 21 minutes       127.0.0.1:9095->9095/tcp    codewind-performance
```

Following @jgwest comment on issue https://github.com/eclipse/codewind/issues/2640#issuecomment-613595594, it looks like PFE container died with the following message:

```
[13/04/20 18:23:22 /portal/modules/LoadRunner.js] [INFO] Loadrunner has disconnected
/portal/modules/LoadRunner.js:428
      this.project.loadInProgress = false; 
                                  ^

TypeError: Cannot set property 'loadInProgress' of null
    at Socket.socket.on (/portal/modules/LoadRunner.js:428:35)
    at Socket.Emitter.emit (/portal/node_modules/component-emitter/index.js:133:20)
    at Socket.emit (/portal/node_modules/socket.io-client/lib/socket.js:138:10)
[...]

PR #2715, opened 2020-04-20
Guard against this.project being null inside LoadRunner.js

hhellyer:
## What type of PR is this ? 

- [X] Bug fix
- [ ] Enhancement

## What does this PR do ?
Adds checks that `this.project` isn't null before we call functions that use in the socket message handlers.
The first time I ran this against a node project I also saw this execption:
```
[20/04/20 10:48:34 /portal/modules/LoadRunner.js] [INFO] profiling.json saved for project nodetoupload
(node:66) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'projectID' of null
    at LoadRunner.emitCompleted (/portal/modules/LoadRunner.js:709:81)
    at Socket.socket.on (/portal/modules/LoadRunner.js:573:14)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```
which is why one call to `emitCompleted()` was moved and the other deleted so that it was always called once and once only.

https://github.com/eclipse/codewind/issues/2692

https://github.com/eclipse/codewind/issues/2692

No

The diff is simpler with ignore whitespace set on.

### DIFF
--- a/src/pfe/portal/modules/LoadRunner.js
+++ b/src/pfe/portal/modules/LoadRunner.js
@@ -524,7 +524,10 @@ module.exports = class LoadRunner {
     */
     this.socket.on('error', (err) => {
       log.error('LoadRunner socket error');
-      this.project.loadInProgress = false;
+      // A socket error could occur while a load run is not in progress.
+      if (this.project) {
+        this.project.loadInProgress = false;
+      }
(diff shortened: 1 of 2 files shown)


ebuckle:
This looks good to me!

PR #2748, opened 2020-04-23
Enhance the method we use to determine whether a project has a metrics dashboard

james-wallis:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Changes the implementation of the `metricsAvailable` and `metricsDashboard` to check via a HTTP request without changing the output.

### Summary
The change should clear up whether we can `a)` show an *metrics-dash and `b)` read and parse a `/metrics` endpoint for the performance dashboard.
Examples:
* For a Java project, if both `/metrics` and `/javametrics-dash` are available, `/metrics` takes precedence.
* For Node and all other languages they will only report a dashboard as available if they have either a `/javametrics-dash`, `/appmetrics-dash` or `/swiftmetrics-dash` endpoint available.
* Although many apps such as Appsody may have a `/metrics` endpoint, testing has shown that we likely will not be able to read the data from it at the moment, so we will ignore it.

Also added a `metricsCapabilities` object to Project.js that is used by the `/projects/:id/metrics/status` API which is called by the performance dash (details in the linked issue).
The `metricsCapabilities` object contains the following details:
```javascript
{
    liveMetricsAvailable: false, // Whether any type of metrics is available at an endpoint
    metricsEndpointReachable: false, // Whether the `/metrics` endpoint is available (will be the endpoint if one is e.g. `/metrics`)
    appmetricsEndpointReachable: false, // Whether a *metrics-dash endpoint is available (will be the endpoint if one is e.g. `/appmetrics-dash`)
    microprofilePackageFoundInBuildFile: false, // Whether a Java project has the microprofile package in its pom.xml
    appmetricsPackageFoundInBuildFile: false, // Whether the project has an appmetrics package in its build file
}
```

The `/projects/:id/metrics/status` API returns the following (Same as above but with two more fields):
```javascript
{
  "liveMetricsAvailable": false,
  "metricsEndpoint": false,
  "appmetricsEndpoint": false,
  "microprofilePackageFoundInBuildFile": true,
  "appmetricsPackageFoundInBuildFile": false,
  "canMetricsBeInjected": false,
  "projectRunning": true
}
```

### Testing
* Removed test for deleted function from `Project.js`
* Added tests for `metricsStatusChecker.js`
* Added tests to Project.js for the additional `metricsCapabilities` object and also `metricsDashboard` and `metricsAvailable`.
    * Also added tests for `getPerfDashPath`.
* Changed the API tests for metrics status to reflect the new API.
* Changed the API tests for inject metrics to not check `metricsAvailable` or `metricsDashboard` as they're now set when we get a project status update from the file-watcher.

* Ran the test scripts, verified that VSCode IDE shows the expected response when a dashboard is available / isn't available.


tobespc:
project inf is now updating correctly but I'm not getting detection on remote connection for /metrics

tobespc:
tested on local and remote, all good and detection works fine

Issue #1888, opened 2020-01-27
Notify 'culprits' when master or release branch build fails


tetchel:
After a merge to master or a release branch, if the build fails, an email should be sent to:
- any person who committed changes to the merged PR
- the person who merged the PR
- the releng head (joseph) and perhaps the backup (me)

this way we can notice and fix build failures asap.

https://stackoverflow.com/questions/30398465/how-to-get-culprits-or-committers-inside-a-jenkins-workflow-with-one-or-more-scm
https://stackoverflow.com/questions/43982256/how-to-invoke-email-ext-plugin-from-jenkins-declarative-script

josephca:
A sample email notification 

![image](https://user-images.githubusercontent.com/25185698/80835145-29d49200-8bc0-11ea-94b1-80b6320d2c70.png)

josephca:
For reference as a part of this issue, https://bugs.eclipse.org/bugs/show_bug.cgi?id=562629 was opened to be able to send email to unregistered users.

PR #2838, opened 2020-05-01
Email notification on build failure for Codewind 

josephca:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Email notification on build failure for Codewind 

https://github.com/eclipse/codewind/issues/1888

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,6 +65,9 @@ def sendEmailNotification() {
     }
 }
 
+def IS_MASTER_BRANCH = env.BRANCH_NAME == "master"
+def IS_RELEASE_BRANCH = (env.BRANCH_NAME ==~ /\d+\.\d+\.\d+/)
+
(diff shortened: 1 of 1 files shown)


josephca:
> Would it be possible to just do this on main branch builds (master or release branches like 0.12.0)? We already get enough email from github about PRs without getting more from Jenkins too. Generally we follow the progress of our PR builds and reviewers will check for a clean build before merging. If we get emails on every build failure then I suspect we'll just end up tuning them out.
> 
> On the other hand if it can be set so we only get them for master or branch builds then they are likely to catch our attention as something serious that could block others will be broken.

Thanks for the feedback. It makes sense. I'll update the code to send a notification for master and release branches.

hhellyer:
Thanks for changing that, it makes these something people should take notice of!

PR #2875, opened 2020-05-07
Add target projectName to project link object

james-wallis:

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
Adds the project name to the Link object so that the IDE's can show a nice error even if the projectID no longer exists.

@tetchel fyi

### DIFF
--- a/src/pfe/portal/routes/projects/links.route.js
+++ b/src/pfe/portal/routes/projects/links.route.js
@@ -43,6 +43,7 @@ router.post('/api/v1/projects/:id/links', validateReq, checkProjectExists, async
 
     await project.createLink({
       projectID: targetProjectID,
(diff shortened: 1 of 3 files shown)


src/pfe/portal/routes/projects/links.route.js
```diff
@@ -43,6 +43,7 @@ router.post('/api/v1/projects/:id/links', validateReq, checkProjectExists, async
 
     await project.createLink({
       projectID: targetProjectID,
+      projectName: targetProject.name,
```
tetchel, line 46: does this still work if the `targetProject` has been deleted (given that we verify the target project exists a few lines above) ?

Issue #2506, opened 2020-03-18
Provide a new API on the performance container to support visualisations of profiling data.


hhellyer:
At the moment we have only one limited way of displaying profiling data within Codewind via the LSPs. These are tightly bound to the editors, the language they support and their data models.

To support providing different visualisations of our profiling data easily we should add API endpoints to the performance container that multiple different visualisations can use to obtain profiling information suitable for their needs.

The two main use cases will be:
- Visualisations that want to present call tree view of the profiling run. Where each code location (function, file, line number, class) may exist in multiple trees.
    - This would include uses such as flame graphs and call trees.
    - Each record in the tree would include location information, sample count, child sample count and parent/child information. 
- Visualisations that want to present total calls to particular code locations.
    - This would include applications such as the LSPs (or other code highlighters) that want to provide total hits for different code locations.
    - Each record would include location information, total sample counts, total child counts.

To provide data for these use cases we will generate a standardised version of the tree data for each language we support profiling in JSON, this will be similar to the standard NodeJS profiling data but augmented with post processed data like child counts and multiple samples combined into one tree.

From the standardised tree for each language we will provide a flattened version without the tree information but with total counts and other summary data calculated.

For NodeJS the process would look like:

`profiling.json -> codewind_profiling.json`

For Java:

`healthcenter.hcd -> codewind_profiling.json`

For both to create the summary:
`codewind_profiling.json -> codwind_profiling_summary.json`

A suggested format for both of these files is included at the end of this description.

This data would be made available to visualisations via a REST API on the performance container. This REST API would allow the data model to be queried using JSONPath. This will allow visualisations to query against the standard data model and get back just the information they are interested in rather than sending back the whole, potentially large, profiling data. (It would be trivial to construct a query to return the whole tree if they wanted it.)

The post processing steps would need to be invoked when a profiling run was complete. Only one per-language script would be required to produce the normalised profiling tree, the same script should then be able to be run against those to produce the summary file. (There should be minimal work required to support a new language.)

This data would then be made available on two new REST APIs that would take a JSONPath query as a parameter and would run that query on the server side to reduce the amount of data sent back. This is important as the visulisations would be running in a browser window and may not be able to store and process the whole profiling tree.

REST API's
```
/api/v1/profiling/tree
/api/v1/profiling/summary
```
Both of these would take projectId and JSONPath query as parameters.

This should support many different kinds of visualisation without writing a specialised back end for each.

---

Sample data format:

[...]

PR #2948, opened 2020-05-14
Enhancement 2506: Add new profiling API

hhellyer:
## What type of PR is this ? 

- [ ] Bug fix
- [X] Enhancement

## What does this PR do ?

 - Adds profiling rest APIs to allow the UI to run JSONPath queries against our profiling data.
 - Adds post processing scripts for node profiling data.
 - Invokes post processing after profiling is complete.
 - Adds tests and test data.

The REST API's use POST rather than GET as GET requests are not expected to have a body. Proxy's between PFE and the user may drop the body from a GET request. (And the bug would be ours to fix.) An alternative would be to pass the JSONPath query in the query field of the URL but that may be ugly as it will contain many special characters.

This is the first part of implementing:
https://github.com/eclipse/codewind/issues/2506

The next part will be adding the ability to post process Java profiling data (in .hcd) files to the standard format we will be using so the UI can work with 

https://github.com/eclipse/codewind/issues/2506

No - the openapi.yml for our REST API changes but we do not expect users to interact directly with this feature.

### DIFF
--- a/src/pfe/portal/modules/LoadRunner.js
+++ b/src/pfe/portal/modules/LoadRunner.js
@@ -651,17 +651,34 @@ module.exports = class LoadRunner {
     const normalisedProfilingPath = this.project.getPathToProfilingTreeFile(this.metricsFolder);
     const mergeCommand = `${process.execPath} /portal/scripts/mergeNodeProfiles.js ${profilingPath} ${normalisedProfilingPath}`;
     log.debug(`Running: ${mergeCommand}`);
-    const mergeProcess = await exec(mergeCommand);
-    log.trace(mergeProcess.stdout);
-    log.trace(mergeProcess.stderr);
-    log.debug('Merging done');
+    try {
+      const mergeProcess = await exec(mergeCommand);
+      // Log these but they should be empty in the success case.
+      log.trace(mergeProcess.stdout);
+      log.trace(mergeProcess.stderr);
+      log.debug('Merging done');
+    } catch (err) {
+      // stdout/stderr may contain useful feedback if there was an error.
+      log.trace(err.stdout);
+      log.trace(err.stderr);
+      throw new Error(`Error running merge command. Exit code ${err.code}`);
+    }
 
     const summarisedProfilingPath = this.project.getPathToProfilingSummaryFile(this.metricsFolder);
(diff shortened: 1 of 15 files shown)


src/pfe/portal/modules/LoadRunner.js
```diff
@@ -636,6 +645,26 @@ module.exports = class LoadRunner {
...
+  async postProcessProfilingData() {
+    /* Run the merge out of process as the profiling data could potentially be quite large. */
+    const profilingPath = path.join(this.workingDir + '/profiling.json');
+    const normalisedProfilingPath = this.project.getPathToProfilingTreeFile(this.metricsFolder);
+    const mergeCommand = `${process.execPath} /portal/scripts/mergeNodeProfiles.js ${profilingPath} ${normalisedProfilingPath}`;
+    log.debug(`Running: ${mergeCommand}`);
+    const mergeProcess = await exec(mergeCommand);
+    log.trace(mergeProcess.stdout);
+    log.trace(mergeProcess.stderr);
+    log.debug('Merging done');
```
markcor11, line 657: Nice debug

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -636,6 +645,43 @@ module.exports = class LoadRunner {
...
+
+    const summarisedProfilingPath = this.project.getPathToProfilingSummaryFile(this.metricsFolder);
+    const summariseCommand = `${process.execPath} /portal/scripts/summariseProfile.js ${normalisedProfilingPath} ${summarisedProfilingPath}`;
+    log.debug(`Running: ${summariseCommand}`);
+    try {
+      const summariseProcess = await exec(mergeCommand);
+      // Log these but they should be empty in the success case.
+      log.trace(summariseProcess.stdout);
+      log.trace(summariseProcess.stderr);
+      log.debug('Merging done');
```
tobespc, line 675: if there is an error in merging, should we still say 'merging done' ?

src/pfe/portal/modules/LoadRunner.js
hhellyer: That should result in an exception and triggering of the `catch` block.
I've realised the message from the summarising block below says `merging` so I'll push a fix for that.

hhellyer: That should result in an exception and triggering of the `catch` block.
I've realised the message from the summarising block below says `merging` so I'll push a fix for that.

Issue #2825, opened 2020-05-01
Provide ability to run load concurrently against multiple projects


tobespc:
Current design of load run only allows one project to be tested at a time.  We want to enable multiple projects to be tested concurrently (especially when using a remote connection and testing on a cloud)

ebuckle:
As it currently stands, the design of LoadRunner is as follows:
- User.js creates a single LoadRunner object that it calls whenever a request for load from a project is made
- This LoadRunner can only hold a single Project object and considers itself to be 'in use' if it's project object is not null
- Once load is finished, this object is nulled and a new project can then request a load run

In the new design:
- Instead of a single LoadRunner object being held by User.js, each project object that gets created will make it's own LoadRunner object
- Instead of telling LoadRunner to perform a load, User.js now tells the project to start load on its own LoadRunner object
- This LoadRunner now has a status object and will now only reject load requests if there is already one in progress for it's parent project
- This allows multiple load requests to be made at once, from separate projects

The main issues that will need to be tested are:
- Whether we can maintain the reliability of LoadRunner when running load runs in series
- The limits of running load tests in series. How many tests we can expect a user to run at once before we hit a reasonable limit

micgibso:
We don't mention any restrictions about running load tests concurrently so no doc updates required.

PR #2977, opened 2020-05-20
Multiple Load Tests

ebuckle:

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

## What does this PR do ?
- Moves the LoadRunner object from being a single object held by User.js to being one instance of the object per Project object
- Enables multiple load runs to be run at once by reworking how the 'Load Run in progress' error is handled throughout Codewind. LoadRunner.js now has a method to state if it is idle, which is used to determine if a project can run a load test
- All LoadRunner socket messages now contain the ProjectID they are associated with. Meaning each LoadRunner object can filter out incoming messages from multiple in progress load runs.

TODO:
- Rework existing testing scripts to work with new layout of Loadrunner.js
- Extensive testing of LoadRunner scenarios to make sure that no functionality has been lost with the changes

https://github.com/eclipse/codewind/issues/2825
Yes. Users now have the ability to run multiple load runs at the same time, which changes the functionality of the performance dashboard.

None.

### DIFF
--- a/src/performance/server.js
+++ b/src/performance/server.js
@@ -31,8 +31,7 @@ const serverPort = 9095;
 const server = app.listen(serverPort, () => log.info(`Performance server listening on port ${serverPort}!`))
 const io = Io.listen(server);
 
-var loadProcess;
-let projectURL;
+var loadProcess = {};
 
 const codewindVersion = process.env.CODEWIND_VERSION;
 const imageBuildTime = process.env.IMAGE_BUILD_TIME
@@ -61,19 +60,19 @@ app.get('/performance/api/v1/environment', (req, res) => {
 */
 app.post('/api/v1/runload', async function (req, res) {
     try {
-        if (loadProcess) {
+        if (!loadProcess[req.body.projectID]) {
+            if (!req.body.url) {
+                log.error('URL is required');
+                res.sendStatus(500);
+                return;
+            }
+            runLoad(req.body);
+            res.sendStatus(202);
+        } else {
             log.error('A load run is already in progress');
             res.status(409).send("A load run is already in progress");
             return;
         }
-        if (!req.body.url) {
-            log.error('URL is required');
-            res.status(500);
-            return;
-        }
-        runLoad(req.body);
-        projectURL = req.body.url;
-        res.sendStatus(202);
     } catch (err) {
         res.status(500).send(err);
         log.error(err);
@@ -88,20 +87,13 @@ app.post('/api/v1/runload', async function (req, res) {
 */
 app.post('/api/v1/cancelLoad', async function (req, res) {
     try {
-        if (req.body.url == projectURL) {
-            if (loadProcess) {
-                loadProcess.kill('SIGKILL');
-                res.sendStatus(200);
-                return;
-            } else {
-                log.error('No run in progress');
-                res.status(409).send("No run in progress");
-            }
-        } else {
-            const msg = 'No run in progress for this project'
-            log.error(msg)
-            res.status(409).send(msg);
+        if (loadProcess[req.body.projectID]) {
+            loadProcess[req.body.projectID].kill('SIGKILL');
+            res.sendStatus(200);
             return;
+        } else {
+            log.error('No run in progress');
+            res.status(409).send("No run in progress");
         }
     } catch (err) {
         res.status(500).send(err);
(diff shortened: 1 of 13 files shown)


src/pfe/portal/modules/LoadRunner.js
```diff
@@ -487,91 +505,99 @@ module.exports = class LoadRunner {
...
-      log.error(err);
-      // If this.up is false we're already trying to reconnect
-      if (this.up) {
-        // socket.io-client will automatically reconnect and trigger the connect event.
-        this.up = false;
+    this.socket.on('error', (err, data) => {
+      if (data.projectID === this.project.projectID) {
+        log.error('LoadRunner socket error');
+        // A socket error could occur while a load run is not in progress.
+        if (this.project) {
```
hhellyer, line 553: I think in the new model this might be wrong. I can't see anywhere that checks `loadInProgress` on the project object anymore (it's not used by loadtest.route.js in particular( so I think this needs to be changed to mark the load run (if running) as finished the right way.

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -487,91 +505,99 @@ module.exports = class LoadRunner {
...
+          this.project.loadInProgress = false;
+        }
+        log.error(err);
+        // If this.up is false we're already trying to reconnect
+        if (this.up) {
+          // socket.io-client will automatically reconnect and trigger the connect event.
+          this.up = false;
```
hhellyer, line 560: `up` is only set to false if the error is for this project.  However I don't think loadrunner.up is ever read anywhere. (I think it was a hangover from when the LoadRunner.js was created from FileWatcher.js which at the time also used a socket.)
It might be less confusing to remove this. (It doesn't affect reconnecting.)

src/pfe/portal/modules/LoadRunner.js
ebuckle: Yep, that's a mistake. Will update to use the ifIdle() method instead, as everything else is using it now

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -487,91 +505,99 @@ module.exports = class LoadRunner {
...
-      // A socket error could occur while a load run is not in progress.
-      if (this.project) {
-        this.project.loadInProgress = false;
-      }
-      log.error(err);
-      // If this.up is false we're already trying to reconnect
-      if (this.up) {
-        // socket.io-client will automatically reconnect and trigger the connect event.
-        this.up = false;
+    this.socket.on('error', (err, data) => {
```
hhellyer, line 549: On there's no call to `this.setStatus` in the error handler here to reset state. The old loadInProgress flag is set to false, do you need to set the run status here too?

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -487,91 +505,99 @@ module.exports = class LoadRunner {
...
     * Socket event for disconnect
     * Logs the error and starts reconnect procedure
     */
-    this.socket.on('disconnect', () => {
-      log.info('Loadrunner has disconnected')
-      // If this.up is false we're already trying to reconnect
-      if (this.up) {
-        // socket.io-client will automatically reconnect and trigger the connect event.
-        this.up = false;
+    this.socket.on('disconnect', data => {
```
hhellyer, line 519: `data` is unused which is a linter error.

src/performance/server.js
```diff
@@ -111,11 +106,11 @@ app.post('/api/v1/cancelLoad', async function (req, res) {
 
 function runLoad(options) {
     log.info(`starting loadrun with options : ${JSON.stringify(options)}`);
-    io.emit('starting');
+    io.emit('starting', {projectID: options.projectID});
     var output = "";
     var errOutput = "";
     loadProcess = childProcess.spawn('node', ['runload.js', JSON.stringify(options)], { stdio: 'pipe' });
```
hhellyer, line 112: `loadProcess` (and `projectURL`) are both global to the whole of the load runner.
If you start two load runs against different urls and cancel the first then the cancel fails because the url doesn't match project URL.
The processes running the load runs should probably be stored in an object indexed by project id. The project id can be passed over on a cancel, which will also remove the need to match the url in the processes command line to find the one to cancel.

src/performance/server.js
```diff
@@ -61,18 +60,12 @@ app.get('/performance/api/v1/environment', (req, res) => {
...
 app.post('/api/v1/runload', async function (req, res) {
     try {
-        if (loadProcess) {
-            log.error('A load run is already in progress');
-            res.status(409).send("A load run is already in progress");
-            return;
-        }
         if (!req.body.url) {
             log.error('URL is required');
             res.status(500);
```
hhellyer, line 65: This line is wrong, it should be sendStatus. (The request never terminates if you call it with curl.)

src/performance/server.js
```diff
@@ -61,18 +60,12 @@ app.get('/performance/api/v1/environment', (req, res) => {
 */
 app.post('/api/v1/runload', async function (req, res) {
     try {
-        if (loadProcess) {
```
hhellyer, line 64: We still need this check to prevent multiple load processes running against one project.
It should check `loadProcess[req.body.projectID]` (after making sure projectID is set!)

src/performance/server.js
```diff
@@ -88,20 +81,13 @@ app.post('/api/v1/runload', async function (req, res) {
 */
 app.post('/api/v1/cancelLoad', async function (req, res) {
     try {
-        if (req.body.url == projectURL) {
-            if (loadProcess) {
```
hhellyer, line 92: Again we should probably have a check for the load process. It'll help us diagnose issues with multiple cancels or bad parameters from portal.

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -462,12 +479,14 @@ module.exports = class LoadRunner {
           'Content-Type': 'application/json'
         }
       }
+      loadConfig.projectID = this.project.projectID;
```
hhellyer, line 482: I think projectID is now the only parameter we need to send. See: https://github.com/eclipse/codewind/blob/7064533114f43f4621563f178c19817ba2e8a430/src/performance/server.js#L88
which is a big improvement as we only check the project id. (We may have been vulnerable to not being able to cancel a run if the config had been modified after it started before.)
You can just pass an object with:
`{projectID: this.project.projectID}`
to the http request and remove the need to pass `loadConfig` from this function and it's callers.

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -705,10 +724,36 @@ module.exports = class LoadRunner {
...
+
+  isIdle() {
+    if (this.status.CURRENT_STATE === states.IDLE) {
+      return true;
+    }
+    return false;
+  }
+
+  setStatus(newStatus) {
+    if (this.status[newStatus] !== null) {
```
hhellyer, line 754: Can we add a `log.trace()` here to say what we're setting the status to for which project? It would be very useful if we are debugging loadrunner in the future.

src/pfe/portal/modules/Project.js
```diff
@@ -129,6 +130,8 @@ module.exports = class Project {
     this.links = new Links(this.projectPath(), args.links);
 
     this.perfDashboardPath = `/performance/charts?project=${this.projectID}`;
+    log.debug(`Creating LoadRunner for user ${this.user_id}`);
```
hhellyer, line 133: `Project.js` has no field called user.id. (I think you probably copied this message from `User.js`.) Should you output the projectID/name here?

src/pfe/portal/modules/User.js
```diff
@@ -257,7 +247,7 @@ module.exports = class User {
           let settingsFilePath = path.join(projFile.workspace, projName, '.cw-settings');
           const settFileExists = await fs.pathExists(settingsFilePath);
           const settFile = settFileExists ? await fs.readJson(settingsFilePath) : {};
-          const project = new Project({ ...projFile, ...settFile });
+          const project = new Project({ ...projFile, ...settFile }, this);
```
hhellyer, line 250: Is passing `this` to a new Project object out of data here and on line 363?
It looks like you changed to pass the user object to `loadrunner.runLoad` on line 166.

tobespc:
looking better but cancelling is still not quite right

PR #2980, opened 2020-05-20
Enhance /metrics/auth API to return better errors + add microprofilePackageAuthenticationDisabled to /metrics/status

james-wallis:
Note: the button on the dashboard won't show yet, but the changes are coming in @markcor11's next PR. Edit: This line: https://github.com/eclipse/codewind/pull/2994/files#diff-fafbd9cba5399ce927af26e85cb12eedR36

What type of PR is this ? 
- [x] Bug fix
- [x] Enhancement

### Summary
* Adds the field `microprofilePackageAuthenticationDisabled` into the metrics status so that the performance dashboard knows whether it should show the `disable metrics authentication or not`
* Enhances the error handling for the metrics auth disable API to send back `400` errors when the request is not wrong. Spoken to Mark and the performance dashboard shouldn't hit these conditions based on the dashboard logic, but they should be included in the API anyway.
* Improve the way the metrics state is checked by having the `setMetricsCapabilities` function run when the project is first bound to Codewind and also when the FileWatcher sends a `projectCreated` Socket message.

### Testing
* Added tests for new functionality.
* Corrected tests with incorrect comments.

### DIFF
--- a/src/pfe/portal/controllers/metrics.controller.js
+++ b/src/pfe/portal/controllers/metrics.controller.js
@@ -12,6 +12,7 @@ const metricsService = require('../modules/metricsService');
 const Logger = require('../modules/utils/Logger');
 const cwUtils = require('../modules/utils/sharedFunctions');
 const { getProjectFromReq } = require('../middleware/checkProjectExists');
+const ProjectMetricsError = require('../modules/utils/errors/ProjectMetricsError');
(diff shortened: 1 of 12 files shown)


tobespc:
Successfully removed authentication and then re enabled it for appsody openliberty

Issue #3029, opened 2020-05-27
Insane number of socket messages when cancelling a load run


tobespc:
**Codewind version:** latest
**OS:** N/A

**Che version:**N/A
**IDE extension version:**N/A
**IDE version:**N/A
**Kubernetes cluster:**N/A

**Description:**
Cancelling load causes an insane amount of socket messages to be emitted which makes all the screen status values flash
![Screenshot 2020-05-27 at 11 37 38](https://user-images.githubusercontent.com/13151286/83009461-b4b57a80-a00e-11ea-90df-9f5e7366969d.png)

**Steps to reproduce:**

**Workaround:**

james-wallis:
Am unable to recreate this:
```
SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "started"}
main.js:1 SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "running", timestamp: "20200602151937"}
main.js:1 SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "running", timestamp: "20200602151937"}
main.js:1 Cancel accepted
main.js:1 SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "cancelled"}
main.js:1 SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "app-is-using-old-metrics"}
main.js:1 Ignoring UISocket RX:  {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "app-is-using-old-metrics"}
main.js:1 SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "started"}
main.js:1 SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "running", timestamp: "20200602152026"}
main.js:1 SocketIO RX: runloadStatusChanged {projectID: "b28deb30-a1b1-11ea-a3c6-719f9fd37463", status: "running", timestamp: "20200602152026"}
main.js:1 Cancel accepted
[...]

PR #3038, opened 2020-05-27
Reduce Cancel Socket Messages

ebuckle:
## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
Reduces the amount of 'Cancelling' socket messages coming out of User.js when a cancel request is made on a load run.

Adds a cancelRequests field to User.js that tracks whether a project load run is currently being cancelled. Prevents additional requests to cancel being made. Previously the user could click the cancel button multiple times on the performance dashboard, which would stack up these requests. 

https://github.com/eclipse/codewind/issues/3029

No

No


ebuckle:
Merged into #2977

PR #3058, opened 2020-05-28
add port discovery using Openshift Routes

james-wallis:

## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

### Summary
Enhance the method we use to determine the application port to use for load testing and linking - when a project has multiple ports we can use the ingress/route port to work out which one is supposed to be the application port.
Support for using the Ingress port was added in [PR 2748](https://github.com/eclipse/codewind/pull/2748/files#diff-77019e368b4024f7f954f2008537bb8aR138). This PR enhances this for the `odo` extension which doesn't create an ingress for its projects by using the `Openshift Route` when an ingress is not available for a project.

### Testing
Manually tested this works.
Added unit tests for new functions.

### DIFF
--- a/src/pfe/portal/modules/utils/kubernetesFunctions.js
+++ b/src/pfe/portal/modules/utils/kubernetesFunctions.js
@@ -126,17 +126,66 @@ async function getProjectIngress(projectID) {
 }
 
 async function getServicePortFromProjectIngress(projectID) {
-  const ingress = await getProjectIngress(projectID);
-  // sanitise the response ingress
-  if (!ingress || !ingress.spec || !ingress.spec.rules
-    || !ingress.spec.rules[0] || !ingress.spec.rules[0].http
-    || !ingress.spec.rules[0].http.paths || !ingress.spec.rules[0].http.paths.length > 0
-    || !ingress.spec.rules[0].http.paths[0].backend) {
+  try {
+    const ingress = await getProjectIngress(projectID);
+    // sanitise the response ingress
+    if (!ingress || !ingress.spec || !ingress.spec.rules
+      || !ingress.spec.rules[0] || !ingress.spec.rules[0].http
+      || !ingress.spec.rules[0].http.paths || !ingress.spec.rules[0].http.paths.length > 0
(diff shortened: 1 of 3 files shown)


src/pfe/portal/modules/utils/kubernetesFunctions.js
```diff
@@ -126,17 +126,66 @@ async function getProjectIngress(projectID) {
...
+    const { targetPort } = route.spec.port;
+    return targetPort;
+  } catch(err) {
+    // If getProjectRoute throws an error return null
+    log.debug('Unable to get the service port from route');
+    return null;
+  }
+}
+
+async function getServicePortFromProjectIngressOrRoute(projectID) {
```
hhellyer, line 181: TBH - I would just call this function `getServicePort()` the rest of it doesn't add much. (It would be a horrible name if there were three, four or five ways to get ports.)

src/pfe/portal/modules/utils/kubernetesFunctions.js
```diff
@@ -151,4 +200,7 @@ module.exports = {
   patchProjectDeployments,
   getProjectIngress,
   getServicePortFromProjectIngress,
+  getProjectRoute,
+  getServicePortFromProjectRoute,
+  getServicePortFromProjectIngressOrRoute,
```
hhellyer, line 205: Are the last two new functions are only exported so the tests can access them?

src/pfe/portal/modules/utils/kubernetesFunctions.js
james-wallis: I've named it this way so its clear that it doesn't use the port from the Kubernetes service. I get your point but I don't think there are anymore ways (at least for us) to get the service port.

src/pfe/portal/modules/utils/kubernetesFunctions.js
james-wallis: Not on purpose but yes the tests are the only thing that calls them. I'll change this.

Issue #3036, opened 2020-05-27
Incorrect capabilities status being returned when testing various projects


tobespc:
**Codewind version:** latest
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**
The capabilities panel in the perf dashboard that shows the features available for the project show different values before and after a load run

![Screenshot 2020-05-27 at 14 01 23](https://user-images.githubusercontent.com/13151286/83021939-aa9d7700-a022-11ea-8fe7-df49f2b05b45.png)
![Screenshot 2020-05-27 at 14 01 35](https://user-images.githubusercontent.com/13151286/83021949-af622b00-a022-11ea-80a1-adfe654f3e3e.png)

**Steps to reproduce:**
create a websphere project, open perf dashboard, see the status, run a load, when the load completes the status changes

stalleyj:
I see some really weird results here too, 'Your project is not running" - yes it is and the load is in progresss.
![Screenshot 2020-05-28 at 10 21 49](https://user-images.githubusercontent.com/13417476/83123975-7b900f80-a0cd-11ea-88a0-c25f8c44e9bd.png)

tobespc:
start a load running against a websphere project and the hit cancel when its in the preparing state to recreate

tobespc:
```{
  "liveMetricsAvailable": true,
  "metricsEndpoint": "/metrics",
  "appmetricsEndpoint": false,
  "microprofilePackageFoundInBuildFile": true,
  "appmetricsPackageFoundInBuildFile": true,
  "hasTimedMetrics": false,
  "microprofilePackageAuthenticationDisabled": false,
  "canMetricsBeInjected": true,
  "projectRunning": true
}```

```
{
  "liveMetricsAvailable": true,
  "metricsEndpoint": "/metrics",
  "appmetricsEndpoint": "/javametrics-dash/",
  "microprofilePackageFoundInBuildFile": true,
  "appmetricsPackageFoundInBuildFile": true,
  "hasTimedMetrics": true,
  "microprofilePackageAuthenticationDisabled": false,
  "canMetricsBeInjected": true,
  "projectRunning": true
}
```

stalleyj:
Also seeing incorrect information for a running Node Express app, {
  "liveMetricsAvailable": false,
  "metricsEndpoint": false,
  "appmetricsEndpoint": false,
  "microprofilePackageFoundInBuildFile": false,
[...]

PR #3059, opened 2020-05-28
Fix incorrect status of metrics in the performance dashboard

tobespc:
## What type of PR is this ? 

- [ x ] Bug fix
- [ ] Enhancement

## What does this PR do ?
This PR resolves the issues seen with the metrics capabilities panel and incorrectly giving status.  The change does the following
* Only refreshes the panel when load collection completes, not on a project restart
* Changes the metrics/status api to return a live value rather than a cached value
* When cancelling load, we no longer stop and restart the project (This was to stop profiling collection but actually causes more harm than good)

#3036

no

### DIFF
--- a/src/pfe/portal/routes/projects/metrics.route.js
+++ b/src/pfe/portal/routes/projects/metrics.route.js
@@ -67,12 +67,7 @@ router.get('/api/v1/projects/:id/metrics/status', checkProjectExists, async func
   try {
     const project = getProjectFromReq(req);
     const { canMetricsBeInjected, appStatus } = project;
-    let capabilities = project.getMetricsCapabilities();
-    //if (!capabilities || Object.keys(capabilities).length === 0) {
-    // If projectMetrics is an empty object, fetch them first
-    const { capabilities: updatedCapabilities } = await project.setMetricsState();
-    capabilities = updatedCapabilities;
-    //}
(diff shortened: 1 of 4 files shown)


src/pfe/portal/routes/projects/metrics.route.js
```diff
@@ -68,11 +68,11 @@ router.get('/api/v1/projects/:id/metrics/status', checkProjectExists, async func
     const project = getProjectFromReq(req);
     const { canMetricsBeInjected, appStatus } = project;
     let capabilities = project.getMetricsCapabilities();
-    if (!capabilities || Object.keys(capabilities).length === 0) {
-      // If projectMetrics is an empty object, fetch them first
-      const { capabilities: updatedCapabilities } = await project.setMetricsState();
-      capabilities = updatedCapabilities;
-    }
+    //if (!capabilities || Object.keys(capabilities).length === 0) {
```
hhellyer, line 71: These lines should probably be removed not commented out.
Also this means that capabilities is assigned on line 70, then reassigned on line 74 without being read. Line 70 could also be deleted. (Assuming `project.getMetricsCapabilities()` has no side effects.)

tobespc: ok, modified those lines

PR #3076, opened 2020-05-29
Only show disable metrics auth after project has started, don't scan metrics when project is stopped

james-wallis:
## What type of PR is this ? 

- [ ] Bug fix
- [x] Enhancement

### Summary
1. Makes the `setMetricsState` only run when the project is added to Codewind and when it's state changes to `started` (application is running) because we do not benefit from show the user that `javametrics-dash` is unavailable when their project is not running. This also reduces flicker on the performance dashboard as the state changes less.

2. Changes the performance dashboard to only show the "allow anonymous-access" button when the project has started once i.e. we've actually been able to scan the metrics. This ensure the user that we don't ask them if they want to disable metrics authentication before we check whether they need to. -> Thanks @markcor11 for this enhancement.

Before the first build the project will no longer show:
image

and will show:
image

Once the project has started once (has a `project.host` property, if the live monitoring is false it will show the "allow anonymous-access" button as before.

### Testing
Ran unit and API tests.

### DIFF
--- a/src/performance/dashboard/src/components/status/CapabilitiesPanel.jsx
+++ b/src/performance/dashboard/src/components/status/CapabilitiesPanel.jsx
@@ -108,6 +108,12 @@ class CapabilitiesPanel extends React.Component {
 
     getCapabilityLiveMetrics(capabilityData, feature) {
 
+        // When the project is not running we can not determine if it supports live metrics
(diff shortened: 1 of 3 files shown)


tobespc:
need to think this one through a bit

Issue #3077, opened 2020-05-29
Restart appsody on k8s, it's possible for the ingress to be created to the wrong port


makandre:
**Codewind version:** 0.13.0/master
**OS:**

**Che version:**
**IDE extension version:**
**IDE version:**
**Kubernetes cluster:**

**Description:**

For restarting appsody app on k8s, a fix (#2805) was put in to re-expose the application over ingress.

I just discovered that sometimes the ingress can be created to the wrong port if the stack has multiple ports

![image](https://user-images.githubusercontent.com/17596240/83280878-3dfeb580-a1a5-11ea-96e9-134e511b05b5.png)

**Steps to reproduce:**

**Workaround:**

malincoln:
@makandre can we get the PR merged no later than tomorrow morning?  Thanks

PR #3082, opened 2020-05-29
0.13.0: Pass in port when re-exposing over ingress

makandre:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?

Pass in port when re-exposing an appsody app over ingress

Port of #3078 to 0.13.0

#3077

### DIFF
--- a/src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
+++ b/src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
@@ -204,8 +204,20 @@ export class ShellExtensionProject implements IExtensionProject {
      * @param projectInfo <Required | ProjectInfo> - The metadata information for the project.
      */
     start = async (projectInfo: ProjectInfo): Promise<void> => {
+
         await projectUtil.runScript(projectInfo, path.join(this.fullPath, "entrypoint.sh"), "start");
-        await projectUtil.exposeOverIngress(projectInfo);
+
+        if (process.env.IN_K8) {
+            let port = undefined;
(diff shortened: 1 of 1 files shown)


src/pfe/file-watcher/server/src/projects/ShellExtensionProject.ts
```diff
@@ -204,8 +204,20 @@ export class ShellExtensionProject implements IExtensionProject {
...
     start = async (projectInfo: ProjectInfo): Promise<void> => {
+
         await projectUtil.runScript(projectInfo, path.join(this.fullPath, "entrypoint.sh"), "start");
-        await projectUtil.exposeOverIngress(projectInfo);
+
+        if (process.env.IN_K8) {
+            let port = undefined;
+            const ports = this.getDefaultAppPort();
+            if (Array.isArray(ports)) {
+                port = parseInt(ports[0]);
```
elsony, line 214: If we just pick the first port, does it mean that we may sometimes get the wrong port?

makandre: For some legacy reasons we allowed https://github.com/eclipse/codewind-appsody-extension/blob/master/.sh-extension#L12 to define multiple app ports before.  But now it only defines 1 app port so it would actually not land here.

johnmcollier: Would this change impact non-Appsdoy projects that do expose multiple ports?

Issue #3032, opened 2020-05-27
vscode reports bad request when attempting to build when build in progress


tobespc:
**Codewind version:** latest
**OS:**

**Description:**
Following error message displayed when manually triggering a build 
![Screenshot 2020-05-27 at 12 09 42](https://user-images.githubusercontent.com/13151286/83012490-622a8d00-a013-11ea-8338-e4aac5ba3ee4.png)

**Steps to reproduce:**
Create a project and start it building, then from the overview page click the build button and then the message will appear

tetchel:
That's the response from the portal server, this is a portal bug

tobespc:
retesting now @tetchel , no one else could get this yesterday so it may be just me

tobespc:
recreated it again.  Steps are create a websphere project, start a load run, whilst its in the load running state, click the build button on the project overview page

stalleyj:
So as per Toby's comment above, it's hitting the build api when a load run is in progress that is the issue.  The build API has been modified to reject a build request and return 400 with message when there is a load run in progress (it already did this for a project that is disabled or being deleted).  Note that the API spec is not accurate and we also return a 400 for project not found.  I propose that we update the API to return 404 for project not found and then return 400 "Unable to build project as  a load run is in progress' separately from 'Cannot perform build action ....' msg.  @tetchel Will VSCode (and other IDEs) also need changes ?

tetchel:
No, VS Code does not look for any specific responses or errors from the build API

tobespc:
why can't you display the error message ?

tetchel:
oh i see, the message being sent is the "cannot build because project is in state" one... I should be showing it, I will see why that isn't working. https://github.com/eclipse/codewind/issues/3057

that said, I can't reproduce this for some reason, I always get the 202 response

stalleyj:
@tetchel the change to the API went into master yesterday some point before DCUT

tetchel:
here is the what the current error looks like with the fixed HTTP error reporting

![image](https://user-images.githubusercontent.com/8888438/83276085-7a7ee100-a1a6-11ea-9b28-aa1fb17b48ad.png)

PR #3086, opened 2020-06-02
Build API Error Status Changes

ebuckle:

## What type of PR is this ? 

- [x] Bug fix
- [ ] Enhancement

## What does this PR do ?
Updates project not found error to use status 404.

Separate error message returned when a project cannot be built due to a load run being in progress. 

#3032 

No

No

### DIFF
--- a/docs/openapi.yml
+++ b/docs/openapi.yml
@@ -119,6 +119,12 @@ paths:
             text/html:
               schema:
                 type: string
+        400:
+          description: Cannot perform action
+          content:
+            text/html:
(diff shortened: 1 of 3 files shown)


docs/openapi.yml
```diff
@@ -119,7 +119,7 @@ paths:
             text/html:
               schema:
                 type: string
-        400:
+        404:
```
hhellyer, line 122: We still return 400 when a build can't be run. We need to add that status code here.

PR #3093, opened 2020-06-03
Fix delete log file/folder error

GeekArthur:

## What type of PR is this ? 

- [x] Bug fix

## What does this PR do ?
This PR fixes the issue of deleting log file/folder if user uses NFS as persistent volume type.

N/A

N/A

### DIFF
--- a/src/pfe/file-watcher/server/src/controllers/projectsController.ts
+++ b/src/pfe/file-watcher/server/src/controllers/projectsController.ts
@@ -994,17 +994,17 @@ export function deleteFolder(dir: string): Promise<void> {
                 if (err) {
                     return reject(err);
                 }
-                // Delete all files in the directory
-                Promise.all(files.map((file) => {
-                    return deleteFile(dir, file);
-                })).then(() => {
-                    rimraf(dir, (err) => {
-                        if (err) {
-                            return reject(err);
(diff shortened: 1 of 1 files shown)


GeekArthur:
@elsony Please review when you get a chance.

src/pfe/file-watcher/server/src/controllers/projectsController.ts
```diff
@@ -994,17 +994,17 @@ export function deleteFolder(dir: string): Promise<void> {
...
-                    return deleteFile(dir, file);
-                })).then(() => {
-                    rimraf(dir, (err) => {
-                        if (err) {
-                            return reject(err);
-                        }
-                        resolve();
-                    });
-                }).catch(reject);
+                // Delete log folder and log files inside log folder
```
elsony, line 997: Why do you change the comment to log folder and log files only given that the dir has not been changed? Are we still deleting all files instead of just the logs?

GeekArthur: Yes, same behavior as before (delete both folder and files inside folder, not specific to log folder/file). I agree this comment is confusing as the the function is util function and should be shared by all callers, I will update the comment.

Issue #3095, opened 2020-06-03
Loadrunner/Performance server - Socket.io Event "Error" is a reserved event 


james-wallis:
**Codewind version:** Master
**OS:**

**Description:**
Loadrunner.js has a Socket event listening to the "Error" event that is sent from the Performance server.js to notify PFE that an error has been hit. This event is actually reserved by Socket.io https://socket.io/docs/emit-cheatsheet/ so we shouldn't be using it.

Additionally, if Socket.io does encounter an error and emits the "Error" event it could potentially crash PFE as `data` will be null and we could get a `data.projectID` property `projectID` cannot be found in null.

[socket.on('error in Loadrunner.js](https://github.com/eclipse/codewind/blob/master/src/pfe/portal/modules/LoadRunner.js#L557)
```
    /**
    * Socket event for error
    * Logs the error and starts reconnect procedure
    */
    this.socket.on('error', (err, data) => {
      if (data.projectID === this.project.projectID) {
        log.error('LoadRunner socket error');
        log.error(err);
        // If this.up is false we're already trying to reconnect
        if (this.up) {
          // socket.io-client will automatically reconnect and trigger the connect event.
          this.up = false;
        }
        this.resetStatus();
      }
    });
```

io.emit('error in Performance:
https://github.com/eclipse/codewind/blob/master/src/performance/server.js#L127-L134

**Steps to reproduce:**

**Workaround:** 

Think we should handle both the socket.io error and also have an error event of our own.

PR #3147, opened 2020-06-16
Stop emitting our own 'error' event via socket.io to avoid clash with socket.io's own 'error' event.

hhellyer:
## What type of PR is this ? 

- [X] Bug fix
- [ ] Enhancement

## What does this PR do ?

- Rename 'error' to `loadrunError` to avoid a clash between our event and a reserved `error` event in socket.io. https://socket.io/docs/emit-cheatsheet/
- Handle both `error` and `loadrunError` in PFE.
Handle `loadrunError` in the UI.
The 'error' event is a socket error and won't relate to specific project so it is not handled by the UI.

We were emitting an `error` event when a load run failed. This clashed with the `error` event that socket.io will emit when an internal error occurs. Our version of the event also specified different parameters and the listening code assumed there would be a projectID sent in the `data` object for the error. This would not be true if an error occurred in socket.io itself and it emitted it's own `error` event.

https://github.com/eclipse/codewind/issues/3095

### DIFF
--- a/src/performance/dashboard/src/components/actions/ActionRunLoad.jsx
+++ b/src/performance/dashboard/src/components/actions/ActionRunLoad.jsx
@@ -216,7 +216,7 @@ class ActionRunLoad extends React.Component {
         const { loadRunStatus, inlineTextLabel, timeRemaining } = this.state;
 
         let loadRunCompleted = loadRunStatus.status === 'completed';
-        const options = ['preparing', 'connecting', 'starting', 'started', 'running',  'collecting', 'completed', 'requesting', 'requested', 'cancelling', 'cancelled']
+        const options = ['preparing', 'connecting', 'starting', 'started', 'running',  'collecting', 'completed', 'requesting', 'requested', 'cancelling', 'cancelled', 'failed']
         const showBusy = options.includes(loadRunStatus.status)
 
         let inlineTextLabelFormatted = (timeRemaining !== 0) ? `${inlineTextLabel}${timeRemaining}` : `${inlineTextLabel}`
--- a/src/performance/dashboard/src/components/actions/ActionRunLoad.jsx
+++ b/src/performance/dashboard/src/components/actions/ActionRunLoad.jsx
@@ -102,6 +102,14 @@ class ActionRunLoad extends React.Component {
                         }, 2000);
                         break;
                     }
+                    case 'failed': {
+                        this.setState({ timeRemaining:0, showModalRunTest: false, loadRunStatus: data, inlineTextLabel: 'Failed...' });
+                        setTimeout(() => {
+                            this.setState({ loadRunStatus: { status: 'idle' } })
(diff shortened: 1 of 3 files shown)


hhellyer:
@ebuckle @markcor11 - Could you both take a look at what I've done here?

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -552,15 +552,33 @@ module.exports = class LoadRunner {
...
+      // If this.up is false we're already trying to reconnect
+      if (this.up) {
+        // socket.io-client will automatically reconnect and trigger the connect event.
+        this.up = false;
+      }
+    });
+
+    /**
+     * Socket event for loadrunError
+     * Logs the error and starts reconnect procedure
```
james-wallis, line 567: Think this comment is wrong, `Logs the error and cancels the loadrun`?

src/performance/dashboard/src/components/actions/ActionRunLoad.jsx
```diff
@@ -102,6 +102,14 @@ class ActionRunLoad extends React.Component {
                         }, 2000);
                         break;
                     }
+                    case 'failed': {
+                        this.setState({ timeRemaining:0, showModalRunTest: false, loadRunStatus: data, inlineTextLabel: 'Failed...' });
+                        setTimeout(() => {
+                            this.setState({ loadRunStatus: { status: 'idle' } })
+                            this.props.dispatch(addNotification({kind: KIND_ERROR, title:'Load run failed', subtitle:'An unexpected error occurred' }));
+                        }, 2000);
```
markcor11, line 110: This timeout is supposed to allow the action button time to display the inlineTextLabel text for two seconds (giving the user chance to read it) before reverting the button back to idle.

However as is,  the button will immediately revert to idle inviting the user to start a new run,  then two seconds later the animated notification will slide in which might be a bit disjointed. I think you may want to add "failed" here:

https://github.com/eclipse/codewind/blob/60a80476d918d8ea7c15098551130f38affac89d/src/performance/dashboard/src/components/actions/ActionRunLoad.jsx#L219

so that the button will render the temporary  "Failed..." text, then show the notification making better use of that timeout.

src/pfe/portal/modules/LoadRunner.js
```diff
@@ -582,7 +600,7 @@ module.exports = class LoadRunner {
       if (data.projectID === this.project.projectID) {
         log.info(`Load run on project ${this.project.projectID} cancelled`);
         this.heartbeat('cancelling');
-        this.cancelProfiling();
+        await this.cancelProfiling();
```
hhellyer, line 603: @tobespc - I asked @ebuckle about the await here and he advised testing this as we changed codewind doesn't start/stop liberty during cancelling phase. Cancelling a liberty run seemed fine anbd awaiting is probably safer to make sure we've finished cancelling before we report that we've cancelled.

